PR #100, opened 2026-07-27
Add arxiv artifact kind and a candidate proposer

ccerv1:
Follows #98. Unblocks citation-based capability scoring for the benchmark/eval datasets category, which today has no capability basis at all — 23 of its 27 products are scored `n/a`.

## Why an identifier, not a name

I tested matching benchmark papers to OpenAlex by product name across 10 products. **Two were correct.** The failures were not near-misses:

| Product | Top match by name | Cites |
|---|---|---|
| **APPS** | "Rayyan—a web and mobile app for systematic reviews" | 25,313 |
| **MATH** | "Stereotype Threat and Women's Math Performance" | 3,405 |
| HumanEval | "CodeGeeX: A Pre-Trained Model for Code Generation" | 236 |
| GSM8K | a paper *about* GSM8K, not GSM8K | 11 |

Short generic names are hopeless by title, and the wrong answers carry large citation counts that look entirely plausible downstream. So the fix is an exact identifier in the registry.

## What's here

**`arxiv` as an artifact kind.** Accepts `https://arxiv.org/abs/2110.14168`, `https://arxiv.org/pdf/2009.03300v3`, `arxiv:1803.05457` or a bare `2311.12022`, and normalises to the bare id. That makes the DOI derivable as `10.48550/arXiv.<id>`, which OpenAlex resolves exactly.

**`build/propose_arxiv.py`**, which does the legwork without making the call. It pulls candidates from two sources we already hold:

1. arXiv links already in a product's yaml or its score sources
2. `arxiv:` tags on the product's Hugging Face artifact — the Hub carries these as first-class tags

Run against `benchmark_eval_data`:

```
single candidate (safe to accept): 16
multiple candidates (pick one):     4
no candidate found:                 7
```

Spot-checking the singles: HumanEval → 2107.03374, GSM8K → 2110.14168, SWE-bench → 2310.06770, MMLU-Pro → 2406.01574, TruthfulQA → 2109.07958. All correct.

## It proposes, it does not decide

Hub tags list *cited* papers alongside the describing paper. `cais/mmlu` tags four ids, only one of which is the MMLU paper. So multi-candidate products are flagged `REVIEW` rather than guessed at, and the script writes nothing to `sources/`.

The four needing a human: APPS, CodeContests, MMLU, MultiPL-E.

Of the seven with no candidate, three look **legitimately paperless** — Anthropic Internal Coding Evals, OpenAI Internal Evals, Scale SEAL Leaderboard. That is a fact about them rather than missing data, and arguably a signal in itself: an eval nobody can cite is a different artifact from one with thousands of citations. The other four (MATH, HLE, EvalPlus, Compar:IA Datasets) likely do have papers and need a manual lookup.

Hub lookups fail soft — a Hub outage degrades a hint rather than failing the run.

## Test plan

- `uv run pytest` — 60 passed (3 new, covering id normalisation, rejection of non-ids, and end-to-end serialization of an arxiv artifact)
- `uv run python -m build.serialize_registry --check` — passes, same 2 pre-existing warnings, 0 errors
- `uv run python -m build.propose_arxiv --category benchmark_eval_data` — output above

## Next

Once ids are added to `sources/`, the OpenAlex UDM is a lookup by DOI with no fuzzy matching, and the benchmark/eval category gets its first capability basis.

### DIFF
--- a/build/propose_arxiv.py
+++ b/build/propose_arxiv.py
@@ -1,0 +1,171 @@
+"""Propose arXiv identifiers for products, from evidence we already hold.
+
+Citation counts need an exact identifier. Matching papers by product name was
+measured at 2 correct out of 10: APPS resolved to a medical software paper with
+25k citations, MATH to a psychology paper with 3.4k. Both would have looked
+entirely credible in a table, which is the dangerous kind of wrong.
+
+So this proposes rather than decides. It gathers candidate arXiv ids from two
+places we already have and prints them with their evidence for review. Nothing is
+written to sources/ — a human picks the canonical paper and adds it.
+
+Two evidence routes:
+  1. arXiv links already present in a product's own yaml or its score sources.
+  2. `arxiv:` tags on the product's Hugging Face artifact. The Hub carries these
+     as first-class tags.
+
+Neither is authoritative. Route 2 in particular lists *cited* papers as well as
+the paper describing the artifact — cais/mmlu tags four, only one of which is the
+MMLU paper. Anything with more than one candidate is flagged REVIEW and needs a
+person to choose.
+
+Usage:
+    uv run python -m build.propose_arxiv                 # every product
+    uv run python -m build.propose_arxiv --category benchmark_eval_data
+    uv run python -m build.propose_arxiv --format yaml   # paste-ready snippets
+"""
+
+from __future__ import annotations
+
+import argparse
+import json
+import os
+import re
+import urllib.error
+import urllib.request
+from pathlib import Path
+
+from build.validate import load_sources
+
+ROOT = Path(__file__).resolve().parents[1]
+HF_API = "https://huggingface.co/api"
+USER_AGENT = "os-ai-map-arxiv-proposer/1.0"
+
+ARXIV_RE = re.compile(r"(?:arxiv\.org/(?:abs|pdf)/|arxiv:)(\d{4}\.\d{4,5})", re.I)
+
+
+def _hf_ids(product: dict) -> list[tuple[str, str]]:
+    """Return (kind, hf_id) for the product's Hugging Face artifacts."""
+    out: list[tuple[str, str]] = []
+    for key, segment in (("huggingface_model", "models"), ("huggingface_dataset", "datasets")):
+        value = product.get(key)
+        entries = [value] if isinstance(value, str) else (value or [])
+        for entry in entries:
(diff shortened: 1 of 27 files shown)


ccerv1:
## Researched and added: 23 of 27 now carry a verified arXiv id

Every id confirmed against the paper's **actual title** via the arXiv API, not accepted from a name match.

### The four REVIEW cases, resolved by fetching candidate titles

| Product | Chosen | Rejected candidate |
|---|---|---|
| APPS | `2105.09938` Measuring Coding Challenge Competence With APPS | `2203.07814` AlphaCode |
| CodeContests | `2203.07814` Competition-Level Code Generation with AlphaCode | `2105.12655` CodeNet |
| MMLU | `2009.03300` Measuring Massive Multitask Language Understanding | UnifiedQA, GPT-3, Shared Human Values |
| MultiPL-E | `2208.08227` MultiPL-E: A Scalable and Extensible Approach… | SantaCoder, StarCoder |

**MultiPL-E vindicates flagging rather than auto-picking.** Both its candidates were wrong — the Hub tagged SantaCoder and StarCoder, models that *use* the benchmark. The real paper wasn't in the candidate list at all. An auto-pick would have attached a code model's citation count to a benchmark, silently.

### Four with no candidate in our data, found by title search

- `math` → `2103.03874` Measuring Mathematical Problem Solving With the MATH Dataset
- `hle` → `2501.14249` Humanity's Last Exam
- `evalplus` → `2305.01210` Is Your Code Generated by ChatGPT Really Correct?

### One extra change the additions forced

`build.validate` failed with **23 errors** — `Additional properties are not allowed ('arxiv' was unexpected)`. The product JSON schema needed `arxiv` too; the serializer change alone wasn't enough. CI would have caught it, but only after review.

### Remaining four are legitimately paperless

Anthropic Internal Coding Evals, OpenAI Internal Evals, Scale SEAL Leaderboard, Compar:IA Datasets. Not missing data — a fact about them, and arguably a signal: an eval nobody can cite is a different artifact from one with thousands of citations.

### Checks, exit codes verified

```
validate:            exit=0   (was 23 errors before the schema fix)
serialize_registry:  exit=0
pytest:              exit=0   60 passed
serialize dry-run:   exit=0
```

Registry now serializes **23 arxiv artifacts** alongside 253 github, 83 hf_model, 66 pypi, 64 hf_dataset, 11 npm, 1 crates.

OpenAlex is now a DOI lookup with no matching step.

PR #105, opened 2026-07-28
Verify the data-openness claims behind base-model scores (RWKV correction)

ccerv1:
First real verification pass. Run **locally** — this session's inference plus web search — rather than waiting on a UDM and an LLM secret. The evidence store is the interface between layer-2's stages, so the research executor is swappable; proving the checklist by using it comes before freezing it into a model.

**Stacked on #102**, which declares `last_verified`. This is its first genuine use.

## Target selection

Not alphabetical. The seven score-5 `open_source` ratings in `base_pretrained` **all rest on a `data:open` claim**. If one is wrong, the map overstates openness in its flagship category — the single most damaging error available. Nine products have a load-bearing `data:` claim; those were the target.

## RWKV was wrong: 5 → 4

The score rested on the RWKV-7 paper claiming a 3.1T-token open corpus. The published datasets:

| repo | contents |
|---|---|
| `Goose-World/RWKV-World-v3` | 10 files, 8.17GB — `100k/subsample_100k.jsonl`, `1m/subsample_1m.jsonl`, READMEs |
| `RWKV/RWKV-World-Listing` | identical |

No corpus shards. The corpus is **described, not released** → `documented-not-released`, which by the rubric gives `(4, open_weights)`.

Weights, training code and license are untouched and remain fully open under Apache-2.0. This is a narrower claim, not a demotion of the project.

Worth naming the pattern: the old evidence was a **paper's claim** rather than a verifiable artifact — precisely what the map's anchor principle ("third-party sources, not vendor release notes") exists to prevent.

## Nemotron 3 overstated, score unchanged

"~10T-token pretraining" implied a full corpus release. Reality: `Nemotron-Pretraining-Specialized-v1.2` is 53.6GB of real synthetic pretraining data, but `Nemotron-Pretraining-Code-v3` is 1.2GB of `Nemotron-Code-Metadata/*.parquet` — metadata, not corpus.

Left at `data:open` because a substantial dataset genuinely is downloadable, and its non-OSI license caps it at 4 regardless. **Rubric gap recorded:** there is no value for a substantial *partial* release.

## Confirmed

Stamped with the verifying API call as a source, `last_verified: 2026-07-28`:

- `lucie-7b` — Lucie-Training-Dataset, public, ungated
- `olmo-3` — Dolma, public, ungated
- `pythia` — the Pile, public, ungated
- `apertus` — `swiss-ai/pretrain-data` resolves 200; reconstruction scripts satisfy `data:open`

## Deliberately not stamped

- **`smollm3`** — cites "public data mixtures, ~11T tokens". `HuggingFaceTB/smollm-corpus` exists, but that is the earlier SmolLM corpus and its correspondence to SmolLM3's ~11T is unverified.
- **`marin`** — cites "documented mixtures". DCLM-based configs appear released, which would satisfy reconstruction, but I confirmed that from secondary sources, not the configs.

Both keep their scores and stay unverified. Stamping them on weaker evidence than the four above would defeat the purpose of the field.

## Also

`nemotron-3`'s `note` contained a copy of its components string, now fixed. `qwen-3-6` has the identical defect — worth a sweep, not smuggled in here.

## Test plan

- RWKV's new values match the rubric rule `{data: documented-not-released, code: open} -> (4, open_weights)`. `check_rubric` lives on the #101 branch so it cannot run here; it should be re-run after both merge.

### DIFF
--- a/sources/scores/apertus.yaml
+++ b/sources/scores/apertus.yaml
@@ -11,7 +11,11 @@ openness:
     code/recipe, intermediate checkpoints, a technical report (arXiv 2509.14233), and EU AI Act + Swiss
     data-protection/copyright compliance docs. All Apache-2.0. Sits in the OLMo/Pythia full-openness
     tier.'
+  last_verified: '2026-07-28'
   sources:
+  - url: https://api.github.com/repos/swiss-ai/pretrain-data
+    shows: Repository resolves 200. Pretraining-data reconstruction scripts are published, satisfying data:open by reconstruction.
+    accessed: '2026-07-28'
   - url: https://huggingface.co/swiss-ai/Apertus-70B-2509
     shows: Apache-2.0; BF16 safetensors weights; training checkpoints on branches; data reconstruction
       scripts; EU AI Act docs; 15T tokens; 4,096 GH200 GPUs; Megatron-LM
--- a/sources/scores/lucie-7b.yaml
+++ b/sources/scores/lucie-7b.yaml
@@ -8,7 +8,11 @@ openness:
   note: 'Full open-source tier alongside OLMo / Apertus: open weights, open training data, open training
     code, intermediate checkpoints. Authors position Lucie as OSI-definition-compliant for open-source AI.
     Training-dataset product deferred — artifacts cited here for openness only.'
+  last_verified: '2026-07-28'
   sources:
+  - url: https://huggingface.co/api/datasets/OpenLLM-France/Lucie-Training-Dataset
+    shows: Public, ungated dataset repo present on the Hub; 3,101 downloads. Corpus is published, not merely described.
+    accessed: '2026-07-28'
   - url: https://huggingface.co/OpenLLM-France/Lucie-7B
     shows: Apache-2.0 weights; links to Lucie-Training-Dataset; intermediate checkpoints; Llama-3.1 arch
     accessed: '2026-07-19'
--- a/sources/scores/nemotron-3.yaml
+++ b/sources/scores/nemotron-3.yaml
@@ -6,7 +6,7 @@ openness:
     is 53.6GB of synthetic pretraining data, but Nemotron-Pretraining-Code-v3 ships metadata parquet rather
     than the code corpus);code:open(recipes+21 RL env configs);license:NVIDIA-Open-Model-License(permissive,commercial,non-OSI)'
   confidence: high
-  note: 'Open weights and open training recipes under a permissive but non-OSI licence, which caps the
+  note: 'Open weights and open training recipes under a permissive but non-OSI license, which caps the
     score at 4 regardless of how open the data is. Data claim narrowed on 2026-07-28: the previous ''~10T-token
     pretraining'' wording implied the full corpus was released. NVIDIA publishes substantial portions
     — 53.6GB of synthetic pretraining data across fact-seeking and related sets — while the code portion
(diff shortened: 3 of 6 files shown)


ccerv1:
## Re-checked after challenge: correction stands, reasoning was too thin

Carl asked me to double-check this, and he was right to. The 5 → 4 outcome holds, but my first note was dismissive about what RWKV actually publishes, and one look isn't enough for a score change on a real project. Latest commit rewrites the evidence.

### What RWKV does publish, which I undersold

An **itemized, machine-readable component index** — `index/train.jsonl` and `train.csv` — naming every source dataset with a resolvable URL, category, citation and version. Nearly all components are public. Plus 100k and 1M preview subsamples.

That is genuine openness work and considerably more than a prose description. My original "component listing + subsamples only" was unfair.

### What still blocks `data:open` — now checked, not assumed

I went looking for the reconstruction path and it isn't there:

- **No corpus, no reconstruction scripts.**
- Exact subsampling amounts are deferred to *"supplementary material (see `wiki.txt`, `oscar.txt`)"*. That is **not** in the dataset repo, **not** an arXiv ancillary file (the abs page lists none), and **not** in `BlinkDL/RWKV-LM`, which carries model and training code by version but no data-prep directory.
- No mixing proportions or per-component token counts.
- One component (OSCAR23.01) is itself **gated**.
- The maintainers' own notes record `peS2o — "Which version out of the 3?"`, so **the mixture is not reproducible in detail even in principle**.

The rubric's `data:open` is "a released dataset, **or** scripts that reconstruct the corpus end to end." Neither holds. 4/`open_weights` is correct, and weights, code and license remain fully open under Apache-2.0 — this narrows one claim rather than downgrading the project.

### The more important finding

**The `data` enum is too coarse.** "Every component named and publicly resolvable, but the mixture not reproducible" sits between `open` and `documented-not-released`, and there is no value for it. Nemotron 3's partial release lands in the same gap.

**Two of the nine products I checked fall between the available values.** That is a rubric problem, not a scoring problem, and it belongs in the #101 discussion about per-category rubrics rather than being patched here.

Proposed fourth value, for that discussion:

```
open                      released corpus, or scripts that rebuild it end to end
components-listed         every component named with resolvable public sources,
                          but mixture/sampling not published, so not reproducible
documented-not-released   composition described in prose only
closed
```

RWKV and Nemotron 3 would both take `components-listed`. Worth noting it likely scores the same as the current outcome for both, so this improves how accurately the map *describes* openness without moving numbers.

ccerv1:
### Merge order and status

- Base auto-retargeted to `main` when #102 merged. **MERGEABLE / CLEAN**, no action needed.
- Best merged **after #101**, so `check_rubric` exists on `main` and can verify this change independently. RWKV's new values match the rule `{data: documented-not-released, code: open} → (4, open_weights)`, but that's currently my hand-verification rather than the checker's.
- The rubric-vocabulary gap this surfaced is now tracked in #106 rather than patched here. Both RWKV and Nemotron 3 keep the scores in this PR under either vocabulary, so #106 doesn't block this.

ccerv1:
### Now verified by the checker, not by hand

#101 is merged, so `build/check_rubric.py` is on `main`. Rebased this branch onto it and ran it:

```
base_pretrained: 47/47 reproduced  [OK]   exit 0
```

Worth being precise about what that does and does not prove. The rubric is a function from evidence to score. I changed RWKV's **evidence** (`data: open` → `documented-not-released`), and the formula maps `{data: documented-not-released, code: open}` to `(4, open_weights)`. So 47/47 holds both before and after the change — the checker confirms the new score follows from the new evidence under the same rule.

That is internal consistency. The **factual** claim — that the 3.1T corpus is not published — rests on the web verification recorded in the score file's sources, not on this.

`build/validate` also clean.

Issue #106, opened 2026-07-28
Scoring: the data-openness vocabulary is too coarse


ccerv1:
## The problem

Openness scoring asks four questions of every base model — weights, data, code, license. The `data` question currently has three answers:

```
open                       a released dataset, or scripts that reconstruct the corpus end to end
documented-not-released    composition is fully described but no data or scripts ship
closed
```

While verifying the nine base models whose openness rests on a `data:` claim, **two of the nine landed between those values**. Not edge cases — two of the most-discussed open model families.

## The two that don't fit

**RWKV** publishes an itemized, machine-readable component index (`index/train.jsonl`, `train.csv`) naming every source dataset with a resolvable URL, category, citation and version, nearly all of them public, plus 100k and 1M preview subsamples.

But the corpus is not released and there are no reconstruction scripts. Exact subsampling amounts are deferred to *"supplementary material (see `wiki.txt`, `oscar.txt`)"*, which is not in the dataset repo, not an arXiv ancillary file, and not in `RWKV-LM`. No mixing proportions. One component is itself gated. And the maintainers' own notes record `peS2o — "Which version out of the 3?"`, so the mixture is not reproducible in detail even in principle.

That is much more open than a prose description and clearly less than a released corpus.

**Nemotron 3** releases substantial portions — `Nemotron-Pretraining-Specialized-v1.2` is 53.6GB of real synthetic pretraining data — while `Nemotron-Pretraining-Code-v3` ships 1.2GB of `Nemotron-Code-Metadata/*.parquet`, metadata rather than the code corpus. A partial release, with no value to express it.

## Proposal

Add a fourth value between the existing two:

```
open                       released corpus, or scripts that rebuild it end to end
components-listed          every component named with resolvable public sources, but the
                           mixture or sampling is not published, so it is not reproducible
documented-not-released    composition described in prose only
closed
```

RWKV and Nemotron 3 would both take `components-listed`.

**This would not change either score.** Both currently land at 4 and would stay there. The gain is that the map would describe their openness accurately instead of flattening a real, distinct posture into the nearest available word.

## Questions for discussion

1. **Is `components-listed` the right cut**, or should partial-release (Nemotron) and full-listing-no-pipeline (RWKV) be separate values? They differ: one ships some real data, the other ships none but names everything.
[...]

PR #116, opened 2026-07-29
Port the scoring rubric to a second category (finetuned_chat)

ccerv1:
The base-model rubric was reverse-engineered from one category's scores. This tests whether it was a general mechanism or a bespoke one.

## The result

**32 of 46 products reproduced against the pilot rubric with nothing changed** — no aliases, no `reads` declarations, no corrections. That is the number worth quoting, and it is recorded in the recipe, because the headline 43/43 comes after correcting three scores and would be circular.

The tier ladder and the formula port intact. What needed adding was vocabulary, not structure. `base_pretrained` holds at 47/47 and 195 evidence rows throughout, which was the constraint the change had to respect.

## Scores this changes

Three published scores move. Each was surfaced by the port, none by re-research.

| product | was | now | why |
|---|---|---|---|
| `codegemma` | 3 open_weights | **2 restricted** | Identical facts and the same license family as `gemma-3`, which scores 2. Its own note already read "use-restricted". |
| `tulu-3` | 3 restricted | **2 restricted** | `3/restricted` is a pair the ladder cannot emit. The extra point credited AI2's open post-training layer on the checkpoint. |
| `tinyllama-1-1b-chat-v1-0` | 4 open_weights | **5 open_source** | `zephyr` scores 5 on the same two post-training datasets while sitting on Mistral-7B, whose corpus is closed outright. |

On `tulu-3`: a fine-tuning approach generalizes across bases — the same AI2 recipe produces `olmo-3-instruct` at 5/open_source. Scoring the checkpoint on the method either buries the method or overstates the artifact. The map already resolves this by scoring the method's artifacts as their own products: `tulu-3-sft-mixture` carries 5/open in `training_synthetic_datasets`. What a user downloading the Llama-based checkpoint is bound by is Meta's license, which is 2. That pair of scores is the signal.

## A bug this fixes

The serializer emitted evidence under the raw `components` key, so `post-training-data` would never have matched the formula's `data` condition. Six products would have fallen through to `otherwise` and scored 3 on an absence — no error, no warning. Evidence is now emitted under the dimension name the warehouse joins on, raw keys are kept as traceability rows, and a test asserts one value per grain.

## Machinery added

- **Recorded-name license aliases**, in `signal_routing.yaml` beside the Hub table that solves the same problem for published slugs. Aliases resolve names only — tier assignment stays a per-category judgment, as that file already argued. Left unmapped, a present use-restricting license reads as *absent*, which is the direction that overstates openness.
- **`reads` on a dimension**, naming which recorded keys answer it, preferring the first key whose value is in the enum over the first key present.
- **`deferred`**, for products the rubric does not decide, each with a reason, printed every run and excluded from the reproduction count rather than counted as passing.

## Deferred, not scored

Three products share one cause: a license restricting **conduct** rather than commerce, which the ladder has no tier for. This is not local to this category — the same licenses are scored three ways across the map, including a contradiction where `DeepSeek-Model-License` is `permissive_non_osi` in `base_pretrained` (`deepseek-v3-base`, 3/open_weights) and use-restricting here (2/restricted). Adding a tier on one category's evidence is what #106 exists to prevent, so it goes to an issue.

## Left alone deliberately

- `nemotron-3` and `nemotron-3-nvidia` are the same product entered twice, once per category, both `display_name: 'Nemotron 3'`.
- `granite-code-instruct` records `data` about its base corpus, which this category's vocabulary spends on post-training data. Now warns rather than dropping the value silently.
- A typed `base:` link from a tune to its base model. `lineage` exists but is dataset-only and free-text.

## Test plan

- `uv run python -m build.check_rubric` — `base_pretrained` 47/47, `finetuned_chat` 43/43 with 3 deferred
- `uv run python -m build.validate` — 0 errors
- `uv run python -m build.serialize_rubric --check` — no errors; one new warning, for the `granite` relabel
- `uv run python -m pytest` — 115 passed, up from 92

### DIFF
--- a/build/check_rubric.py
+++ b/build/check_rubric.py
@@ -75,6 +75,31 @@ def head(value: str) -> str:
     return re.split(r"[(,]", value)[0].strip()
 
 
+_RECORDED_ALIASES: dict[str, str] | None = None
+
+
+def recorded_license_aliases() -> dict[str, str]:
+    """Canonical license name for a name as hand-written in a `components` string.
+
+    Read from `sources/signal_routing.yaml`, which already owns the equivalent map
+    for Hub-published slugs. One declaration for both, because they are the same
+    kind of fact — what a license is CALLED — and two copies would drift.
+
+    Deliberately does not carry a tier. Which tier a license belongs to stays a
+    per-category judgment, and a canonical name that appears in no category example
+    still abstains. That abstention is the signal to extend the rubric.
+    """
+    global _RECORDED_ALIASES
+    if _RECORDED_ALIASES is None:
+        routing = yaml.safe_load((ROOT / "sources" / "signal_routing.yaml").read_text()) or {}
+        aliases = ((routing.get("dimensions") or {}).get("license") or {}).get("aliases") or {}
+        _RECORDED_ALIASES = {
+            str(slug).strip().lower(): str(name).strip()
+            for slug, name in (aliases.get("recorded") or {}).items()
+        }
+    return _RECORDED_ALIASES
+
+
 def normalize_license(raw: str) -> str:
     """Reduce a recorded license string to the license that governs the weights.
 
@@ -84,14 +109,19 @@ def normalize_license(raw: str) -> str:
         because it is the one attached to the artifact being scored.
       * `assumed-Modified-MIT` — the `assumed-` prefix marks confidence, not a
         different license.
+
+    Then the recorded-name alias, which resolves spellings of one license to one
+    name. Applied last, so it sees the value after the mechanical steps rather
+    than having to enumerate every prefixed variant.
+
     Purely mechanical. Anything needing judgment is left alone to be flagged.
     """
     value = head(raw)
     if "+" in value and "model" in value.lower():
         value = value.split("+")[-1]
         value = re.sub(r"(?i)^\s*model\s+", "", value)
-    value = re.sub(r"(?i)^assumed-", "", value.strip())
-    return value.strip()
+    value = re.sub(r"(?i)^assumed-", "", value.strip()).strip()
+    return recorded_license_aliases().get(value.lower(), value)
 
 
 def license_tier(raw: str, recipe: dict) -> str | None:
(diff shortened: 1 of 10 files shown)


ccerv1:
Deployed and verified end to end. Both categories reproduce **through the platform**, checked per product against `sources/scores` rather than by row count:

```
base_pretrained   47/47 reproduce
finetuned_chat    43/43 reproduce, 3 deferred
```

Runs: registry static models `54fb49dd` SUCCESS → `evidence.product_evidence` `6c4b1e77` SUCCESS → `scores.openness_computed` `14ed2d0e` SUCCESS.

### The warehouse needed a matching fix, which is not in this diff

`scores_openness_computed` mirrors `check_rubric.normalize_license` as two `REGEXP_REPLACE` calls, and it never consulted the alias table. So the recorded-name aliases this PR adds resolved locally but not in the warehouse: `check_rubric` reported `base_pretrained` 47/47 while the warehouse abstained on `glm-4-9b` and `nemotron-3-nvidia`, scoring both null on "license maps to no declared tier".

Fixed in **`scores.openness_computed` revision #6**, which adds a `doc_license_aliased` CTE joining `currentai.registry.license_aliases` on `source = 'recorded'` — mirroring what the hub path already did. Noting it here because `udms/` lives outside this repo and is not version controlled, so this comment and the platform's revision history are the only record.

The general point is worth keeping: **any logic hand-mirrored between `build/` and the warehouse SQL can drift, and only an end-to-end check catches it.** A green `check_rubric` is not evidence the platform agrees.

### One hole left open

The three deferred products do come back null from the warehouse — but only because their licenses are undeclared, so the pre-existing unmapped-license guard fires. The recipe's `deferred` block never crosses the bridge. A deferral for any other reason (an ambiguous `data` value, say) would be scored by the warehouse regardless. That wants a `category_deferrals` table and a filter; it is not in this PR.

Also worth flagging while these files are open: `pyoso` served a stale result set immediately after the fix materialized, reporting the two products as still unscored when `execute_sql` showed them correct. Verification through pyoso needs the query text varied.

PR #118, opened 2026-07-29
Merge the three display_name collisions, and warn on future ones

ccerv1:
Splitting the Nemotron finding out of #116 as its own change. This PR adds the guard and fixes one stale fact; it deliberately does **not** resolve the duplicate, because that needs a decision.

## The guard

Nothing compared `display_name`s — slugs are unique by construction, since they are filenames — so four collisions had accumulated:

```
'GPT-4.1'           gpt-4-1 (base_pretrained)          + gpt-4-1-openai (finetuned_chat)
'Gemini 3.5 Flash'  gemini-3-5-flash (base_pretrained) + gemini-3-5-flash-google (finetuned_chat)
'GitHub Copilot'    github-copilot (orchestration_agents) + github-copilot-github-microsoft (ui_api)
'Nemotron 3'        nemotron-3 (base_pretrained)       + nemotron-3-nvidia (finetuned_chat)
```

Each renders as two identically labeled entries, so a reader cannot tell them apart or tell whether the map is double counting.

`build_registry` now warns, naming both slugs **and both categories**. The categories are the useful signal: a `base_pretrained` / `finetuned_chat` pair is usually one product entered twice, while two categories on different layers is usually two real surfaces needing distinct labels.

A warning rather than an error, because these four are not one defect and do not share one fix. Copilot genuinely is an agent mode plus an IDE assistant. Nemotron 3 is one family entered twice, and resolving it means deleting a record. Making the build fail would force whoever hits it next to decide under time pressure.

## The stale fact

Both Nemotron records describe Ultra as *"announced for H1 2026"*. It shipped:

```
nvidia/NVIDIA-Nemotron-3-Ultra-550B-A55B-BF16   created 2026-06-03   339,360 downloads
```

Corrected to *"550B total / 55B active, shipped 3 June 2026"* — the announced ~500B / 50B was also wrong.

## What this leaves open, and why

**The Nemotron duplicate is real.** I checked the Hub rather than trusting the records' own comments, which claim a base anchor exists:

```
nvidia/NVIDIA-Nemotron-3-Nano-30B-A3B-BF16      994,374 dl
nvidia/NVIDIA-Nemotron-3-Super-120B-A12B-BF16   855,265 dl
nvidia/NVIDIA-Nemotron-3-Ultra-550B-A55B-BF16   339,360 dl
nvidia/nemotron-3-8b-base-4k                          1 dl   created 2023-11-14
```

The 2026 family ships **instruction-tuned weights only**. The sole `-base` repo is the 2023 8B generation with one download — a different model. So `nemotron-3` in `base_pretrained` has no base checkpoint to point at, both records describe the same family, and both score 4/open_weights on identical facts.

Resolving it means deleting a record, which moves a roster count and removes a published score. `base_pretrained` 47 → 46 (and that category's `scores_reproduced`) or `finetuned_chat` 46 → 45. That is a curation decision, so it is not in this PR.

**Neither record declares any Hub artifacts**, which is why neither has dataset-grade evidence for weights or license. The three repos above are verified and ready to declare — but putting them on both products would assert that two products own one set of weights, so that waits on the dedup too. Worth knowing: the Hub reports `license: other` on all three, which is in the abstain list, so declaring artifacts would earn dataset-grade **weights** but not license.

`GPT-4.1` and `Gemini 3.5 Flash` look like the same shape — a closed API model entered once as a base and once as a chat product, both scoring 1/closed. Same decision, and consistency across the three matters more than which way it goes.

## Test plan

- `uv run python -m build.serialize_registry` — reports all four collisions with their categories
- `uv run python -m build.validate` — 0 errors
- `uv run python -m pytest` — 95 passed, up from 93 (two new tests: the collision warning names both slugs and both categories, and distinct names stay quiet)


ccerv1:
Updated: this now **resolves** the three duplicates as well as adding the guard. Carl confirmed the prior resolution was to keep the `finetuned_chat` record, which #114 had already settled:

> `finetuned_chat` owns closed frontier models … It is also the honest artifact — there is no observable base model for a closed system.

and which #114 listed as a follow-up ("the four tiers spanning base_pretrained and finetuned_chat"). So the survivor is the chat record in each case, taking the bare slug:

```
nemotron-3        <- nemotron-3 (base) + nemotron-3-nvidia
gpt-4-1           <- gpt-4-1 (base) + gpt-4-1-openai
gemini-3-5-flash  <- gemini-3-5-flash (base) + gemini-3-5-flash-google
```

### A first reading I had to reverse

I initially judged the retired records better sourced, because `nemotron-3` carried 5 openness sources against the survivor's 2. Running them through `admit()` inverted it: 3 of those 5 say only `flagship phase-C verification source`, and *both* its adoption and *both* its capability sources are the same filler. `gemini-3-5-flash`'s adoption and capability sources are all rejected too. Counting rows rather than admissible rows is exactly the error the evidence policy exists to catch, and on admissible evidence the surviving records win in all three cases — an independent check on #114's convention.

### Counts

`base_pretrained` 47 → 44 and its recipe moves with it; `check_rubric` stays green. **`finetuned_chat` is unchanged at 46, so #116's counts are untouched.** Products 498 → 495, frozen long-tail re-synced with `overlap` held at 245 — none of the three declared a github artifact, the same reasoning #114 used.

### Merge conflict with #116

Both branches edit the same assertion block in `tests/test_serialize_rubric.py`. #116 rewrites it to per-category counts; this one changes `195` → `183`. Whichever lands second needs that hunk resolved by hand — the correct combined value is `{"base_pretrained": 183, "finetuned_chat": 205}`.

### Left deliberately

GitHub Copilot still collides. Its two records are an agent mode and an IDE assistant — two real surfaces needing distinct labels, not a merge — and the warning will keep saying so.

PR #143, opened 2026-08-06
feat: the payload carries slugs, organizations, aliases and freshness

ccerv1:
The gap map front end cannot build a stable per-product URL today. `build/registry/` is gitignored, so `products.csv` never reaches a consumer, and `notebook_data.json` carries display names only — which is why the app currently keys every product on `${org}__${product}`.

`sources/slug_aliases.yaml` puts the deadline plainly: *"the front end keys links on display names today, so nothing pointed at a product slug yet. After links exist, a rename costs an alias forever."* This adds the identity layer before those links are minted.

## What the payload gains

- `slug` and `org_slug` on every product record. Both were already resolved inside `_row`'s caller and thrown away in favour of display names.
- A top-level `organizations` block: 257 entries from `sources/organizations/*.yaml`, with `github` flattened from `[{url}]` to plain strings. Not from `build/registry/organizations.csv`, which is gitignored and whose `github` column is a stringified Python repr.
- A top-level `aliases` block: the 63 product aliases and 2 organization aliases, and deliberately nothing else. `renamed_before_links` is excluded because `grok` is a live product today, so emitting that block as redirects would 308 a working page onto a different one.
- `freshness` per product: `{date, basis}`, where `basis` distinguishes a real `last_verified` (6 of 470) from the commit-date fallback, so a consumer can label the weaker claim honestly instead of passing it off as a verification.

## Why `fetch-depth: 0` moved with it

`regenerate.yml` and `validate.yml` both used a bare `actions/checkout@v4`, which is depth 1. In a shallow clone every score file reports as added at the grafted root commit, so all 470 products come back stamped with the tip commit date. It does not error; it produces a uniform, plausible, wrong date and publishes it as a verification record. `build/freshness_payload.py` now refuses to emit anything when history is shallow — or when it cannot determine whether history is shallow, which is the same hazard through a different door.

## Gates

`build/check_payload.py` covers the serialized artifact rather than the sources: slug present and unique, `org_slug` resolves, alias keys disjoint from live slugs, every alias target live, and more than one distinct freshness date across the corpus — the shallow-clone symptom seen from the inside.

`build/check_retirement.py` fails when a slug leaves the payload without an alias. 69 product files have been deleted since June and 14 have no alias, so this is a live gap rather than a theoretical one. It distinguishes "no previous payload" (a legitimate first-run skip) from "git failed" (an error).

Every gate was injured and observed red for the failure it names before being wired in.

## Also

60 occurrences of "licence" in score-file notes corrected to American English, across 6 files. These are user-visible prose. No key was touched; every changed line is inside a `note:` string.

## Verification

311 tests, up from 266. `validate`, `check_payload` and `check_retirement` all green on the real payload. Two consecutive serializer runs are byte-identical, which matters because this file drives a daily automated PR downstream.

### DIFF
--- a/.github/workflows/regenerate.yml
+++ b/.github/workflows/regenerate.yml
@@ -18,6 +18,10 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v4
+        with:
+          # Full history: build.freshness_payload dates each score file from its last
+          # commit, and a shallow clone collapses all 470 onto the tip commit.
+          fetch-depth: 0
       - uses: astral-sh/setup-uv@v5
         with:
           enable-cache: true
--- a/.github/workflows/regenerate.yml
+++ b/.github/workflows/regenerate.yml
@@ -31,6 +31,14 @@ jobs:
         run: uv run python -m build.validate
       - name: Serialize (generated = run date)
         run: uv run python -m build.serialize --date "$(date -u +%F)"
+
+      # Same identity gate validate.yml runs on every PR, plus the retirement check that
+      # only a two-payload diff can do: fail the bot commit rather than let a slug 404
+      # silently. See build/check_payload.py and build/check_retirement.py.
+      - name: Gate the serialized payload
+        run: uv run python -m build.check_payload
+      - name: Fail on an unrouted retirement
+        run: uv run python -m build.check_retirement
       - name: Render notebook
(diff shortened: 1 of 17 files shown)


ccerv1:
Reopening to re-trigger CI: the force-push that removed the bot-owned generated file did not fire a workflow run.

PR #150, opened 2026-08-08
Name the gates instead of numbering them

ccerv1:
Stacked on #149 — review that one first; this retargets to `main` automatically when it lands.

`G1`–`G6` work fine sitting next to their own definitions. The problem is where the number travelled away from the table and a reader has nothing to resolve it against:

- `sources/categories/benchmark_eval_data.yaml` — "G3 caught it, which is what G3 is for", in a category note an external contributor reads when tracing why a product was deferred.
- `docs/guides/product-info.md` — "gated (G1/G2)", in a prose style guide that never mentions gates otherwise.
- `--gate {g1,g2,g3}` — a flag whose values you have to go look up before you can run one check.

A numbered handle works when the number is always resolvable, which is true inside one spec document and false the moment it leaves. So: names that carry their own meaning.

| was | prose | code / CLI |
|---|---|---|
| G1 | the invariant | `invariant` |
| G2 | the digest requirement | `digests` |
| G3 | the producible-pair check | `producible_pairs`, `--gate producible-pairs` |
| G4 | the sampled re-fetch | `refetch` |
| G5 | the parity gate | `parity` |
| G6 | the release assertion | `release` |

Applied everywhere: docs, module docstrings, function names, `GATES` keys and CLI values, CI step names, test names, and that category note. The gate table in `verification.md` keeps one line recording the old numbering, because older PRs and commit messages still use it — #148's squash commit says G4 permanently and history stays as it is.

## Two things found while renaming

**A name collision that broke CI.** `check_verification` had an unrelated helper already called `producible_pairs(recipe)`. The gate took that name and the helper became `rule_outcomes` — but `build/check_recipe.py` imports it, so the surviving one-arg call bound the four-arg gate and died with a `TypeError`. Caught by the verification pass, not by review; `validate.yml` runs that command, so the branch was red until it was fixed.

**The parity row said "network, daily".** The cron is `0 6 * * 1` — Monday only — and `parity.yml`'s own comment says "this schedule is why the gate is weekly rather than daily". The wrong cell sat twelve lines above a paragraph arguing that a daily gate over a weekly chain is exactly how a gate earns its way into being ignored. `verification-pass.md` repeated it. Both corrected.

## Prose

The passages the rename touched got a pass for readability at the same time, since a lot of people read this repo: `scepticism` → `skepticism`, `catalogues` → `catalogs`, semicolons split into periods where the house style prefers them, and one plural that had lost its subject ("The gate therefore covers…" after the subject became two gates). No rule, threshold, count or date was changed — with one exception, called out below.

One meaning change was caught and reverted before commit: "the sampled re-fetch checks a sample against the recorded digests" had quietly dropped the network fetch, which is the entire point of that gate and sits in the normative paragraph that says not to relax any of the three. It now reads "goes back to the network for a sample and compares".

## Test plan

- `build.validate` → `0 error(s)`
- `build.check_verification --verbose` → all three `[OK]`, columns aligned on the longest name
- `--gate invariant` / `--gate digests` / `--gate producible-pairs` → each runs alone; `--gate g3` is now rejected
- `build.check_recipe` → `0 failure(s)`
- `pytest tests/ -q` → 337 passed
- `.github/workflows/*.yml` all parse
- Repo-wide sweep for `\bG[1-6]\b` → only the retirement note. The remaining hits are Glean's Gartner/G2 rating, Mali-G610 part numbers, and generated files.

### DIFF
--- a/.github/workflows/parity.yml
+++ b/.github/workflows/parity.yml
@@ -1,6 +1,6 @@
 name: parity
 
-# Gate G5: the repo and the warehouse must compute the same score for every product.
+# The parity gate: the repo and the warehouse must compute the same score for every product.
 # `build/check_rubric.py` walks each ladder in Python and
 # `currentai.scores.openness_computed` walks it in Trino, and only this makes the two agree.
 # Every drift this project has shipped was invisible to a count — the reasoning and the four
--- a/.github/workflows/refetch.yml
+++ b/.github/workflows/refetch.yml
@@ -1,12 +1,12 @@
 name: refetch
 
-# Gate G4: independent evidence that a recorded fetch actually happened.
+# The sampled re-fetch: independent evidence that a recorded fetch actually happened.
 #
-# G1 and G2 both read what the writer wrote. An agent that never opened a page can emit a
-# plausible `accessed`, a plausible 200 and sixty-four plausible hex characters, and both
-# gates pass — #108's failure mode with better spelling. G4 is the only check that goes and
-# looks, and it is the third leg the runbook names. The reasoning is in
-# build/check_refetch.py.
+# The invariant and the digest requirement both read what the writer wrote. An agent that
+# never opened a page can emit a plausible `accessed`, a plausible 200 and sixty-four
+# plausible hex characters, and both gates pass — #108's failure mode with better spelling.
+# The sampled re-fetch is the only check that goes and looks, and it is the third leg the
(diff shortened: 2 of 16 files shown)


ccerv1:
Superseded by #151 — GitHub closed this one when #149's branch was deleted on merge. Same commit, replayed onto `main`.

PR #161, opened 2026-08-09
inference_code: the category re-read

ccerv1:
Second category of the verification sweep. All 20 products re-read against primary sources, prose and scores together.

**56 of 60 axes now carry a `last_verified`**, up from 0. Every source behind them records `http_status` and `content_sha256` — 111 of 111 sources in the category carry a digest. `check_refetch` reproduced 12 digests exactly across a five-product sample, with 0 dead and 0 unreachable; the rest drifted on live counters (GitHub star counts, a rotating provider leaderboard) with every cited span still present.

## Held, per axis rather than per product

| product | axis | why |
|---|---|---|
| `fireworks-inference` | capability | No longer among the twelve providers the Artificial Analysis gpt-oss-120b board tracks, so the recorded 658.8 t/s third place cannot be re-derived. |
| `onnx-runtime` | capability | The MLPerf citation behind `basis: benchmark` is an MLCommons CI test-submission row — open division, submitted by MLCommons itself, configs naming GitHub-hosted runners. Not a benchmark result for this product. |
| `aws-neuron` | adoption | No developer, customer or download figure published anywhere this pass reached, and the cited source establishes nothing about adoption. |
| `qualcomm-ai-engine-direct` | openness | `source:closed` rests on the qai-appbuilder wrapper's README alone; the Software Center and docs pages that would show the EULA-gated distribution render as JavaScript shells. |

Each is in `sources/verification_queue.yaml` naming what would settle it.

## Scores that moved

- **`lmdeploy` adoption 3 → 2.** PyPI reports 50,402 downloads in the trailing 30 days against the 165,001 the prior read cited. Both the API and the human page agree, so it is not a fetch artifact. Same methodology, lower band — worth a sanity check on whether the June figure was the anomaly.
- **`aws-neuron` openness components.** The driver license was recorded Apache-2.0 and is GPL-2.0, confirmed against the driver repo and its LICENSE body. Score stays 1/closed: the driver is open glue, not the headline component.

## Four stale deferrals came off

`check_recipe` failed until they did. `vllm`, `apple-core-ml-runtime`, `google-cloud-tpu-inference` and `qualcomm-ai-engine-direct` now record the `source` and `core-gated` values their deferrals were waiting on. **The category goes from 10 products reproducing to 15**, and parity moves 387/85 to 391/81.

`qualcomm-ai-engine-direct` keeps its queue entry while losing its deferral, which is the right split: the ladder can now read the values, and what remains in doubt is the evidence behind them.

## What the audits caught

Two `failed` verdicts became the holds above. Twenty-one `suspect` verdicts were applied as corrections, including:

- **`vllm` openness rested on a `core_gated` nothing recorded.** The score-5 rung requires `core_gated: ungated` and the components string had no such key — the exact defect its deferral named. Now recorded and attributed.
- **A Groq source claimed to have read the org's complete repo listing.** The body was page one of two, and the count was 29 against an actual 41. Conclusion survived; the citation did not.
- **`vllm` capability `basis_detail: MLPerf` stood for an April 2024 result.** Read with a 2026-08-09 accessed date it looks like current standing. Now dated in the field itself, which matters because three products in this category peg their band to this anchor.
- Several asserted negatives had no stated method, or a method that could not have found what it claimed. Corrected to say how the page was walked.

## Test fixtures

Two pinned counts updated with the reason, in the house style: `test_check_parity` 387/85 → 391/81, and `test_serialize_rubric` `inference_code` 47 → 64 evidence rows.

## Gates

```
validate              0 error(s)
check_verification    invariant / digests / producible-pairs   [OK]   134 axes dated
check_capability      [OK]  27 of 472 products record a comparison
check_recipe          0 failure(s)
check_freshness       inference_code: 60 axes, median 0d
pytest                394 passed
```

Next category by worst artifact coverage is `ui_api`, 43 products at 53%.

### DIFF
--- a/sources/categories/inference_code.yaml
+++ b/sources/categories/inference_code.yaml
@@ -56,23 +56,7 @@ scoring_recipe:
       because: >-
         source or core-gated is not recorded in a form the ladder can read; needs a read of
         the repo and pricing page
-    google-cloud-tpu-inference:
-      because: >-
-        source or core-gated is not recorded in a form the ladder can read; needs a read of
-        the repo and pricing page
-    apple-core-ml-runtime:
-      because: >-
-        source or core-gated is not recorded in a form the ladder can read; needs a read of
-        the repo and pricing page
     sambanova-cloud:
-      because: >-
-        source or core-gated is not recorded in a form the ladder can read; needs a read of
-        the repo and pricing page
-    qualcomm-ai-engine-direct:
-      because: >-
-        source or core-gated is not recorded in a form the ladder can read; needs a read of
-        the repo and pricing page
-    vllm:
       because: >-
         source or core-gated is not recorded in a form the ladder can read; needs a read of
         the repo and pricing page
--- a/sources/products/apple-core-ml-runtime.yaml
+++ b/sources/products/apple-core-ml-runtime.yaml
@@ -6,4 +6,5 @@ description: Apple Core ML is an on-device inference runtime that executes model
   model graphs across the available hardware and runs .mlpackage and .mlmodel files, which the separately distributed
   coremltools package produces. Apple's own on-device features, including Apple Intelligence, are built on
   it.
-comments: Verified 2026-08-08 via Apple Developer docs.
+comments: Verified 2026-08-09 via Apple's Core ML documentation, the MLModel availability list, and the coremltools
+  optimization guide.
--- a/sources/products/aws-neuron.yaml
+++ b/sources/products/aws-neuron.yaml
@@ -5,4 +5,5 @@ description: AWS Neuron is the SDK and runtime for running machine learning on A
   Trainium accelerators (Inf1/Inf2 and Trn1/Trn2/Trn3 instances). Its compiler converts models from PyTorch
   and JAX into executables for the custom silicon, with serving through a vLLM plugin and a kernel-authoring
   interface (NKI). AWS distributes it as a separate SDK, publishing samples and some tooling on GitHub.
-comments: Verified 2026-08-08 via the AWS Neuron documentation and its release notes.
+comments: Verified 2026-08-09 via the AWS Neuron documentation, the aws-neuron-sdk and aws-neuron-driver GitHub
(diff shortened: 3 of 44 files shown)


ccerv1:
Review follow-ups pushed (`4f84d66`). Three of the four flagged items are resolved; the queue is down from 4 axes to 3.

**`vllm` capability re-sourced to a 2026 benchmark.** It now cites [AMD's MLPerf Inference v6.0 submission on Instinct MI355X](https://rocm.blogs.amd.com/artificial-intelligence/mlperf-inference-v6.0/README.html) (April 2026), where vLLM is the serving stack across gpt-oss-120b, Llama 2 70B and a multi-node Llama 3.1 405B run reported at 97-98% scaling efficiency. This replaces the April 2024 Red Hat result, which left the anchor's confirmation two years fresher than the measurement under it — the thing `check_capability` cannot see, and three products here peg their band to this anchor. The new digest reproduced on re-fetch.

**`lmdeploy` stays at 2, now banded on history rather than two point reads.** The monthly series is 171,750 (May), 57,776 (June), 54,993 (July), 50,402 trailing — a step change between May and June that has held for three months. The prior 165,001 was a trailing window ending 2026-06-04, so it was measuring May; the two reads never disagreed.

On whether this is two products: no. Releases from v0.13.0 onward attach no wheel assets, so PyPI is the only route the package ships through and there is no higher-adoption channel to prefer. Both endpoints are now cited as sources rather than merely quoted in the note.

**`onnx-runtime` capability comes off the queue**, re-banded to `feature_matrix` against the published execution-provider summary. The score stays 4 — only the basis was wrong. The withdrawn MLPerf citation is recorded in the note so it does not get quietly re-added.

Remaining queue: `fireworks-inference` capability, `aws-neuron` adoption, `qualcomm-ai-engine-direct` openness.

The same download-history method would settle `llama-factory`, which is queued in `finetuning_code` for exactly this reason. That is a different category, so it wants its own PR rather than being folded in here.

PR #172, opened 2026-08-09
Mirror signal_pypi's SQL into warehouse/models, reading the declared bands

ccerv1:
Second half of #171. The bands are declared and `registry.adoption_bands` is live with 15 rows; this is the SQL that consumes them instead of hardcoding.

**Not deployed — see the blocker below.** Review the SQL here and I'll apply it, or apply it yourself in the UI.

## What changes

The band CASE goes away. The model joins `currentai.registry.adoption_bands` on `product_type` and takes the highest level whose exclusive lower bound the figure clears:

```sql
JOIN currentai.registry.adoption_bands b
  ON b.product_type = m.product_type
 AND m.downloads_30d > b.above
GROUP BY ...  -- MAX(b.level)
```

A product type with no bands declared contributes no row, so the LEFT JOIN leaves `adoption_level` NULL. `hardware` abstains by construction rather than borrowing another type's scale — the same "abstain rather than substitute" rule `signal_routing.yaml` states for sources.

## Why the file exists at all

`docs/runbooks/deploy-udms.md` says UDM source of truth lives in `warehouse/models/`, and **no `signal_*` model was ever put there** — not `signal_github`, `signal_huggingface`, `signal_pypi`, or any of the other four. That gap is exactly how the adoption bands came to live in a CASE expression nothing in the repo could see, applying the software scale to datasets where level 5 is unreachable.

This is the first one mirrored back. The README now says so rather than implying the directory is complete.

## The blocker

`createDataModelRevision` requires `cron`, `kind` and `schema` alongside the code, and none of the three is readable through the GraphQL API that exposes revisions — `GetDataModel` and `GetAssetRevision` both return the code and neither returns those. Inferring them is not safe on a production model: a wrong cron silently changes the refresh cadence, and a wrong column type silently changes what everything downstream reads.

I stopped rather than guess.

## Worth knowing before choosing the cron

**`signal_pypi` has never run on a schedule.** All four of its runs are `MANUAL`:

| when | who |
|---|---|
| 2026-07-27 | manual (one failed, one succeeded) |
| 2026-08-08 | manual |
| 2026-08-09 | manual — mine, to pick up the 34 new declarations |

Its description says "Weekly". So the freshness of every adoption figure it feeds has depended on somebody remembering, and the four-day-old window I reported earlier was four days old because you refreshed it by hand on the 8th.

That's worth fixing in the same edit as the band change, since setting the cron is part of creating the revision either way.

### DIFF
--- a/warehouse/models/README.md
+++ b/warehouse/models/README.md
@@ -10,7 +10,23 @@ entities (USER_MODEL)      — resolved identities: repos, projects, packages, m
 events (USER_MODEL)        — pre-filtered event logs (GitHub Archive)
 metrics (USER_MODEL)       — normalized daily activity per repo
 scores (USER_MODEL)        — interpretive: taxonomy, dependencies, fragility, rankings
+registry (STATIC_MODEL)    — pushed by CI from sources/; see build/publish_registry.py
+signal_* (USER_MODEL)      — one dataset per external source; see below
 ```
+
+### The signal_* models are deployed and mostly NOT mirrored here
+
+`signal_github`, `signal_huggingface`, `signal_pypi`, `signal_lmarena`,
+`signal_artificialanalysis`, `signal_semanticscholar` and `signal_goodailist` all run in
+the warehouse, and until 2026-08-09 none of their SQL was in this directory — while the
+deploy runbook says this is where the source of truth lives. That gap is how
+`signal_pypi` came to carry the adoption bands as a hardcoded CASE that nothing in the
+repo could see.
+
+`signal_pypi_package_downloads.sql` is the first one mirrored back. The rest still need
+recovering from the platform, and a revision's `cron`, `kind` and `schema` are not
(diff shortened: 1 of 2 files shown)


ccerv1:
Closing unmerged — the premise is wrong.

The SQL is deployed (revision 3, released, run, verified: every band boundary now lands exactly on the declared threshold), so the change itself is live. But copying it into `warehouse/models/` is the part we don't want: OSO UDMs shouldn't be mirrored into the repo, and `warehouse/` is heading for deprecation rather than expansion. Adding a seventh model there would have made the confusion worse, not better.

The cron is set to weekly (Monday 04:00 UTC), per the default we're now assuming for any UDM. That also fixes the finding this PR surfaced: `signal_pypi` had never run on a schedule, all four prior runs were manual.

Filing the deprecation separately.

Issue #223, opened 2026-08-12
[methodology] 166 adoption bands are off the scale their own rubric declares


ccerv1:
`build/check_adoption.py` (new, report-only) compares each product's recorded `(level, reach)` against the `adoption_bands` its product type declares. **166 of 432 banded products are off-scale.**

The bands were already declared in `sources/rubrics/*.yaml`, serialized into `currentai.registry.adoption_bands`, and read by the scoring SQL. Tests covered the declaration's shape and the serializer's output — but nothing ever compared a *product* against them. So the bands were authoritative for the warehouse and advisory for the corpus, which is the hand-mirrored-logic drift `check_parity` exists to catch one axis over.

`dataset.yaml` predicted it in a comment: *"14 dataset products record level 4 against a reach of 1M-10M, which is level 5 on this scale ... They are a work list, not a migration."* A work list in a comment is not a gate, so it grew.

## Already fixed (#223)

- 38 case-only labels (`10k-100k` → `10K-100K`). No level moved.
- 13 benchmark corpora re-verified against live HF figures and rebanded onto the dataset scale. Three moved level: `mmlu` 5→4 (it claimed `>10M` against 477,890 monthly downloads), `mmmu` 4→3, `evalplus` 4→3.

## What remains, and why it needs a ruling rather than a fix

| count | shape |
|---|---|
| 91 | reach label not on the applicable scale |
| 60 | a level recorded with **no reach label at all** |
| 15 | level disagrees with its own reach label |

**The big one is `stars_fallback`.** 58 products carry `signal_type: stars_fallback` but record a label from the *download* vocabulary (`100K-1M`, `<10K`) rather than the declared stars vocabulary (`>10K stars`, `1K-10K stars`, `<1K stars`). The corpus never adopted that scale.

That would be mechanical, except **the levels don't follow the stars scale either**:

| product | stars (per its own note) | records | stars scale says |
|---|---|---|---|
| `bigcodebench` | 503 | level 2 | 1 |
| `bigcode-dataset` | ~497 | level 2 | 1 |
| `agent-infra-sandbox` | ~4.9k | level 1 | 2 |
| `autogpt` | 185k | level 3 | 3 (label wrong only) |
| `coder-oss` | ~13.4k | level 3 | 3 (label wrong only) |

[...]

PR #224, opened 2026-08-12
Gate adoption bands against the declared scale, and reband 13 stale benchmarks

ccerv1:
## The gap this closes

`adoption_bands` are declared per product type in `sources/rubrics/*.yaml`, serialized into `currentai.registry.adoption_bands`, and read by the scoring SQL. Tests covered the declaration's shape and the serializer's output.

**Nothing ever compared a product's recorded `(level, reach)` against them.** The bands were authoritative for the warehouse and advisory for the corpus — the hand-mirrored-logic drift `check_parity` exists to catch, one axis over. 217 of 432 banded products were off-scale when first measured.

`dataset.yaml` predicted it in a comment: *"14 dataset products record level 4 against a reach of 1M-10M, which is level 5 on this scale ... They are a work list, not a migration."* A work list in a comment is not a gate, so it grew instead of shrinking.

## `build/check_adoption.py`

Checks two things:

1. the reach label is one the applicable scale declares
2. the recorded level is the level that label denotes

Scale resolution is keyed on `(product_type, signal_type)`, not type alone, because `stars_fallback` has its own labels and its own ceiling of 3 — a star is not a use.

Report-only by default, like `check_freshness`. Gating today would fail on a backlog that predates the bands existing, and a gate that fails on day one teaches people to skip it.

## Fixed here: 217 → 166

**38 case-only labels** normalized (`10k-100k` → `10K-100K`). No level moved.

**13 benchmark corpora** re-verified against live Hugging Face figures and rebanded onto the dataset scale — whose top band is `>1M`, one order of magnitude below software and model, because no dataset artifact in the corpus has ever exceeded 10M.

Three moved **level**, not just label:

```
mmlu       5/'>10M'    -> 4/'100K-1M'   on   477,890 monthly downloads
mmmu       4/'1M-10M'  -> 3/'10K-100K'  on    51,058
evalplus   4/'1M-10M'  -> 3/'10K-100K'  on    51,081  (HumanEval+ and MBPP+ summed)
```

`mmlu` was the worst of them: it claimed the top band and a reach of `>10M` against a real figure of 477,890, and its own note argued *"Level 5 on download volume"*.

All 13 now carry `last_verified` plus `http_status` and `content_sha256` from a live read through `build.fetch_source`, so `check_refetch` can later prove the read was real. Those bodies carry a download counter and drift between any two requests — `check_refetch` documents that case and never fails on drift.

**Real `last_verified` coverage: 180 → 193 axes.**

## Deliberately not fixed — see #223

The remaining 166 need a ruling, not a correction. 58 `stars_fallback` products record download-vocabulary labels, and their levels don't follow the stars scale either (`bigcodebench` records level 2 on 503 stars; the scale says 1). Rebanding them would move published adoption levels on dozens of products, which feeds the gap map's maturity signal. That's a methodology decision.

## Test plan

- `pytest` — 461 passed, plus `tests/test_check_adoption.py` (5 new)
- `validate`, `check_recipe`, `check_rubric`, `check_components`, `check_routing`, `check_capability`, `check_verification`, `check_retirement`, `check_adoption` — all green
- The new tests pin the 13 refreshed corpora and their `last_verified`, so they cannot silently drift back
- Generated files not committed (bot-owned)

One thing caught in review of my own work: the first attempt wrote `establishes: [adoption]` onto the new sources, which `validate` rejected — `establishes` names a scoring dimension, not an axis. Stripped rather than re-fetched, so the figures on record are the ones reported above.

### DIFF
--- a/build/check_adoption.py
+++ b/build/check_adoption.py
@@ -81,10 +81,40 @@ def declared_scales(sources: dict) -> dict[tuple[str, str], dict[str, int]]:
     return scales
 
 
+# Instruments whose bands are read in downloads, and which may therefore fall back to the
+# product type's declared download scale. Everything else measures something else entirely.
+#
+# `unknown` is absent deliberately: it means nobody has decided which instrument applies, so
+# there is no scale to check against and no basis for guessing one.
+_DOWNLOAD_INSTRUMENTS = {"usage_volume", ""}
+
+
 def scale_for(scales: dict, product_type: str, signal_type: str) -> dict[str, int] | None:
-    """The most specific scale that applies, signal first."""
+    """The scale that applies, or None when the instrument declares none.
+
+    ABSTAIN RATHER THAN SUBSTITUTE. `sources/signal_routing.yaml` states the rule directly:
+    "When the authoritative signal for a dimension is missing or unusable, the rule is to
+    produce NO evidence. Falling through to a less authoritative signal is how the failure
+    above happens." `docs/guides/adoption.md` says the same thing about this exact check —
+    "a `reported_traction` or `active_users` band is measuring something else entirely.
+    Comparing it against a download count is a category error, and a check must SKIP it
+    rather than flag or waive it."
+
+    The first version of this function did exactly that, and it is worth recording why the
+    bug was invisible: it read `if the signal declares its own scale use it, otherwise use
+    the product type's`, which is correct for `stars_fallback` and for `usage_volume` and
+    silently wrong for the two instruments that have no scale at all. It manufactured a
+    download-scale finding for all 93 `reported_traction` products and every `active_users`
+    one, and those findings looked exactly like the real ones.
+
+    So the fallback is now allowed only for instruments actually denominated in downloads.
+    `reported_traction` and `active_users` return None and are skipped, which is not a
+    waiver — it is the checker declining to compare a user count against a download band.
+    """
     if signal_type and ("*", signal_type) in scales:
         return scales[("*", signal_type)]
+    if signal_type not in _DOWNLOAD_INSTRUMENTS:
+        return None
     return scales.get((product_type, "*"))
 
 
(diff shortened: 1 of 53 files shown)


ccerv1:
## Correction pushed: the checker had the same bug it was written to catch

Review of this branch found that `scale_for` **substitutes where the design says abstain** — and 26 of the findings it reported were manufactured by the checker itself.

`registry.adoption_bands` is populated from exactly two places: the four type rubrics (emitted as `usage_volume`) and the stars route. So **`reported_traction` (108 products) and `active_users` (23) have no declared scale anywhere in the repo** — two of the five values in the `signal_type` enum, with no band table. The old fallback sent them to the product type's *download* scale.

Both normative sources forbid precisely that:

> `docs/guides/adoption.md`: "a `reported_traction` or `active_users` band is measuring **something else entirely**. Comparing it against a download count is a category error, and a check must **skip** it rather than flag or waive it."

> `sources/signal_routing.yaml`: "**Abstain rather than substitute.** When the authoritative signal for a dimension is missing or unusable, the rule is to produce NO evidence… Falling through to a less authoritative signal is how the failure above happens."

```
examined  432 -> 321   (111 correctly skipped)
findings  166 -> 140   (26 were the checker's own category error)
```

The bug was invisible because its output looked exactly like a real finding — same shape, same message. That is the fourth instance in this repo of an instrument reporting confidently while pointed at the wrong thing, and the first one I shipped myself.

The fallback is now allowed only for instruments denominated in downloads. `unknown` is deliberately excluded: it means nobody has decided which instrument applies, so there is no scale and no basis for guessing one.

`tests/test_check_adoption.py` pins the abstention for both instruments across all three product types **and** asserts `usage_volume` / `stars_fallback` still resolve — otherwise the abstention could be bought by disabling the check. The corpus-walk floor moves 400 → 300 for the same reason and stays a real floor.

467 tests pass; all gates green.

**This does not waive the 15 products carrying a level with no reach.** It establishes that their instrument has no scale for them to be off — a different statement, and the one #223 actually asked for.

PR #227, opened 2026-08-12
Re-verify 31 adoption bands: label the unlabelled, and move stars onto real download signals

ccerv1:
Applies the findings of four parallel research agents. Every figure was independently re-fetched and verified before writing.

## Part 1 — the 18 unlabelled `usage_volume` bands

Each claimed `signal_type: usage_volume` and a level while recording **no reach label**, so the unit the band was read in was never written down. Writing the labels forced the figures to be re-read, and **11 of 18 were banded above what their own downloads support**.

```
gaia                 4 -> 3       17,975 HF   (-57%, real decline)
gemini-cli           5 -> 4    1,850,744 npm  (-35%, and 5 was never reachable)
haystack             4 -> 3      933,408 PyPI (7% under the 1M line)
humanitys-last-exam  4 -> 3       34,715 HF   (flat — a banding error)
langflow             4 -> 2       85,138 PyPI
lightrag             4 -> 3      313,096 PyPI
llama-index          5 -> 4    7,203,303 PyPI (-29%, briefly cleared 10M)
mcp-inspector        4 -> 3      905,396 npm  (9% under the 1M line)
mimo-pro             4 -> 2       59,525 HF
n8n                  5 -> 3      393,738 npm
pathway              3 -> 2       14,405 PyPI (-52%)
trulens              4 -> 3      111,932 PyPI (nearly tripled, still band 3)
```

**The root cause is not decay.** Only four moved because the world moved. In the rest the note's *own cited figure* already contradicted the recorded level — and three notes say why, citing GitHub **stars** beside the download figure:

- `langflow` records 4 on *"~150k stars and ~71k PyPI"* — 71k is level 2
- `n8n` records 5 on *"~193k stars … ~316k npm"* — 316k is level 3

Same defect as the stars cap being read as a default. `adoption.md` already names it: *"20,000 stars and 20,000 downloads are not the same reach."*

Every borderline case sat **below** its threshold, never above — the original pass rounded up at boundaries.

## Part 2 — 13 `stars_fallback` products that had a real signal all along

Several notes said the download volume was *"not retrievable"* or *"not surfaced"*. It was retrievable, and three of the figures were already in our own warehouse.

```
mcp-python-sdk    3 -> 5   333,111,291  PyPI `mcp`
markitdown        3 -> 5    12,755,994  PyPI
docling           3 -> 5    12,493,111  PyPI
semantic-kernel   3 -> 4     2,638,796  PyPI (Python only; NuGet larger, uncounted)
filesystem-mcp    3 -> 4     1,945,615  npm, per-server package
slack-mcp         2 -> 3       111,307  npm
spec-kit          3 -> 2        76,423  PyPI `specify-cli`
beeai             3 -> 2        63,527  PyPI + npm
gpt4all           3 -> 2        60,336  PyPI + npm
mt-bench          3 -> 2         2,361  HF dataset
nano-graphrag     2 -> 1         1,334  PyPI
rwkv              3 -> 1         4,652  HF, declared repos
```

**Seven go down.** `spec-kit` is the sharpest: 126,418 stars, 76,423 downloads a month.

`filesystem-mcp` is the best find — its note objected that the monorepo's 89k stars are shared across every server in it. The npm package is per-server, so it measures this product exactly.

## Identity was verified per package

The obvious name is a different project more often than not. Found and **avoided**:

| trap | what it actually is |
|---|---|
| PyPI `github-mcp-server` | a third party's, not GitHub's |
| PyPI `slack-mcp-server` | unedited template, homepage `github.com/yourusername` |
| npm `markitdown` | converts GFL to HTML |
| PyPI `openrag` | a different vendor's product |

`mcp` was checked hardest, since a wrong package there would have fabricated a 333M figure. Its Repository URL is `github.com/modelcontextprotocol/python-sdk` and its author is the LF project itself.

## Deferred deliberately

**`graphrag`.** PyPI carries `Source = github.com/microsoft/graphrag`, which is right, but the author field reads *"Mónica Carvajal"* rather than Microsoft. Downgrading Microsoft's GraphRAG 3 → 2 on ambiguous provenance is the wrong trade; it keeps its stars band until first-party ownership is confirmed.

**Genuinely undistributable, left on stars:** `github-mcp` (Go binary + ghcr.io, no counter), `localai` and `ragflow` (Docker Hub reports *cumulative* pulls, not monthly, so it doesn't fit the scale), `surfsense`, `openrag`, `blaxel-sandbox` (only platform-wide SDKs exist, not sandbox-specific), `openmanus` (PyPI package is vestigial at 144/month, description literally "Add your description here").

## Blast radius: zero

Simulated across all 16 categories against `main`: **no category stage and no gap list changes.** Structural reason — stars cap at 3 and "mature" needs 4.5, so none of these products was contributing to `mature_open`.

## Test plan

- `pytest` — 461 passed
- `validate`, `check_recipe`, `check_rubric`, `check_components`, `check_verification`, `check_capability` — green
- All 31 carry `last_verified` plus `http_status` and `content_sha256` from a live read

One mechanic worth recording: the first attempt called `fetch()` for the digest and `requests.get()` for the body. The second request was rate-limited into an HTML error page while the digest read 200 — a digest attached to a body nobody parsed. `fetch()`'s own docstring warns about exactly this: *"Fetching a second time to save the body would break that."* Fixed by using its `body_dir`.

### DIFF
--- a/docs/guides/adoption.md
+++ b/docs/guides/adoption.md
@@ -223,6 +223,47 @@ it does not make the claim automatic, it makes it falsifiable.
 - **MLPerf and Artificial Analysis** — capability instruments, not adoption. Recorded here only
   so nobody re-proposes them; see `signal_routing.yaml` for why both are unbridged anyway.
 
+## The mirror trap: a channel that is not counted at all
+
+Attribution has two failure modes and only one of them was written down. The `cohere-rerank-api`
+case above is **over**-attribution — one SDK's downloads credited wholly to one of the N products
+it serves. The opposite is **under**-coverage, and it is the more common one:
+
+> **If the declared artifact is not the product's primary distribution channel, banding on it is
+> a substitution, not a measurement.**
+
+`n8n` is the case that established the rule. It records `usage_volume`, and its declared artifact
+is the npm package at 393,738 downloads a month, which bands at level 3. But n8n is deployed
+overwhelmingly as a self-hosted Docker container, and Docker Hub reports **246 million cumulative
+pulls** — averaging about 2.9 million a month over the image's lifetime. Banding on npm alone
+published a precise number for the wrong channel, and the note said so in its own second sentence
+before recording the band anyway.
+
+That is exactly what `sources/signal_routing.yaml` forbids: "Abstain rather than substitute. When
+the authoritative signal for a dimension is missing or unusable, the rule is to produce NO
+evidence." A partial channel is an unusable signal wearing a usable one's clothes, and it is worse
+than an absent one, because it carries a `last_verified` date asserting that somebody confirmed it.
+
+**What to do instead**, in order of preference:
+
+1. **Count every channel the product actually ships through** and sum them, which is what the unit
+   already says — "summed across declared artifacts". Declare the missing artifact so the sum is
+   reproducible rather than hand-assembled.
+2. Where a channel reports only a **cumulative** total — Docker Hub's `pull_count` is the case —
+   a lifetime average is admissible as a floor, provided the note says it is a lifetime average
+   and therefore understates a growing product. It is not a trailing-30-day figure and must not be
+   presented as one.
+3. Where the primary channel publishes nothing at all, use `reported_traction` and abstain on
+   `reach`, rather than banding on the minority channel that happens to be countable.
+
+`langflow` is the same shape and moved with it: PyPI alone gave level 2, PyPI plus the Docker
+average gives about 166,000 a month and level 3. `semantic-kernel` is a third — its Python package
+reaches level 4 on its own, and the .NET/NuGet channel is larger and uncounted, so its band is a
+floor and its note says so.
+
+The tell to look for when reviewing: **a note that describes the signal as understating the
+product, followed by a band recorded on that signal anyway.**
+
 ## Checklist
 
 - [ ] `level` is 1-5 and follows the band table for the product's **type**.
--- a/sources/products/codegemma.yaml
+++ b/sources/products/codegemma.yaml
@@ -6,5 +6,9 @@ description: 'Google''s code-specialized Gemma family: a 2B pretrained model for
   on 500B-1T tokens of public code, math, and synthetic data.'
 huggingface_model:
 - url: https://huggingface.co/google/codegemma-7b-it
+- url: https://huggingface.co/google/codegemma-7b
+- url: https://huggingface.co/google/codegemma-2b
+- url: https://huggingface.co/google/codegemma-1.1-7b-it
+- url: https://huggingface.co/google/codegemma-1.1-2b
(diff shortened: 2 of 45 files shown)


ccerv1:
## Correction: n8n and langflow were banded on the wrong channel

Review flagged n8n at level 3 as too low. It was — and the cause is a methodological error, not a judgment call.

n8n's declared artifact is npm at 393,738/month → level 3. But **n8n is deployed overwhelmingly as a self-hosted Docker container**, and Docker Hub reports **246,014,721 cumulative pulls** since 2019-06-22 — averaging ~2,871,000/month over the image's lifetime.

| | Docker avg/mo | + registry | combined | band |
|---|---|---|---|---|
| **n8n** | ~2,871,000 | 393,738 npm | **~3,265,000** | 3 → **4** |
| **langflow** | ~81,000 | 85,138 PyPI | **~166,000** | 2 → **3** |

**The damning part: I wrote the caveat and published the band anyway.** n8n's note read *"predominantly self-hosted via Docker, so npm understates real deployment, but it is the countable artifact"* — and then recorded the npm band.

That is exactly what `signal_routing.yaml` forbids: *"Abstain rather than substitute."* A partial channel is an unusable signal wearing a usable one's clothes, and it's **worse than an absent one**, because it carries a `last_verified` date asserting somebody confirmed it.

## The rule that was missing

`adoption.md` named the opposite trap — `cohere-rerank-api` banded on an SDK serving Cohere's entire API surface, *"an SDK covering N products may not be attributed wholly to one"* — and never named this one. It does now, in preference order:

1. **Count every channel and sum them**, which is what the unit already says ("summed across declared artifacts").
2. Where a channel reports only a **cumulative** total (Docker Hub's `pull_count`), a lifetime average is admissible **as a floor**, provided the note says it is one.
3. Where the primary channel publishes nothing, use `reported_traction` and abstain on `reach` rather than banding on whichever minority channel happens to be countable.

**The tell for reviewers**, written into the guide: *a note that describes its own signal as understating the product, followed by a band recorded on that signal anyway.*

## Swept the corpus for that tell

Five products' notes say "understates" or "is a floor". After this change all five are honest — `semantic-kernel` and `nemotron` declare their bands as floors explicitly, and `llama-factory` is the best of them: it traces PyPI through a spike to 779,461 in June and back to 21,867, bands on the 20–30K steady state, and records the git-clone undercount as *unquantifiable* rather than guessing at it.

## Caveat carried in both notes

Both Docker figures are **lifetime averages**, not trailing-30-day counts. For a growing product that understates the current rate, so these levels are floors.

`check_verification` caught a second error of mine here: I carried the registry figures over from the earlier pass without their digests, claiming a same-day read with no proof of fetch. Re-fetched both.

461 tests, all gates green, zero category-stage impact re-simulated against `main`.

Issue #184, opened 2026-08-10
Invert the artifact-kind routing test so it catches new kinds by default


ccerv1:
`tests/test_signal_routing.py` asserts that every artifact kind a product declares has a declared source in `sources/signal_routing.yaml`. It builds the "in use" set by intersecting each product record's keys against a hardcoded `ARTIFACT_KINDS`:

```python
ARTIFACT_KINDS = {"github", "npm", "pypi", "crates", "go", "huggingface_model", "huggingface_dataset"}
in_use = {k for k in ARTIFACT_KINDS if record.get(k)}
assert in_use <= declared
```

So it only catches an undeclared kind whose name is *already in that set*. A genuinely new artifact kind added to a product record later never enters `in_use` at all, and the assertion passes. That is the regression the test is nominally there to prevent.

This is not hypothetical: products already declare an `arxiv` key, which is in neither `ARTIFACT_KINDS` nor `sources:`.

The test's docstring now states this limit honestly, so nothing is currently misleading. But the guard is weaker than it looks.

### The inverted form

Enumerate every top-level product key, subtract an allowlist of known non-artifact metadata keys, and assert the remainder is a subset of the declared sources. That catches any new kind by default, which is the right polarity for this check.

### What blocks it

It needs a ruling on `arxiv`: is it a countable distribution artifact that should be declared as a source (`bridged: false`, like npm and crates), or metadata that belongs on the allowlist? It routes indirectly via `semanticscholar` today, which argues for metadata, but that is a judgment about the taxonomy rather than an implementation detail.

Same question applies to any other product key that turns out to sit in the gap. Worth enumerating them all as the first step.

### Suggested order

1. List every distinct top-level key across `sources/products/*.yaml` and sort into artifact-kind vs metadata.
2. Settle `arxiv` and anything else ambiguous.
3. Invert the test, with the metadata allowlist as the explicit thing a contributor updates.

Context: the current test landed in #182, which declared `npm` and `crates` as unbridged sources after 13 products were found declaring those artifacts with no route.

Issue #173, opened 2026-08-09
Deprecate warehouse/ — it is a stale mirror that actively misleads


ccerv1:
`warehouse/` presents itself as the source of truth for warehouse assets and is not one. It cost real time this week, so this is a concrete case rather than a tidiness complaint.

## The evidence

`docs/runbooks/deploy-udms.md` opens with:

> UDM SQL source-of-truth lives in `warehouse/models/`.

That is false for every `signal_*` model — `signal_github`, `signal_huggingface`, `signal_pypi`, `signal_lmarena`, `signal_artificialanalysis`, `signal_semanticscholar`, `signal_goodailist`. None of their SQL was ever in the directory, and `warehouse/models/README.md` documents five datasets that predate all of them.

**That gap is how the adoption bands went unnoticed.** They lived in `signal_pypi` as a hardcoded CASE applying the software scale to every product type — under which level 5 is unreachable for datasets entirely — and nothing in the repo could see it. Looking for the SQL in the place the runbook names turns up nothing, which reads as "no such model" rather than "the directory is incomplete."

## What is actually in there

| path | contents | status |
|---|---|---|
| `warehouse/models/` | 12 `.sql` for entities / events / metrics / scores, plus a README | deployed models exist; whether the files still match them is unverified |
| `warehouse/catalog/` | 5 dirs of CSV reference data | partly superseded — `sources.yaml` already records the GoodAI roster being retired because "a mirror could only be staler than the live table, and it was" |
| `warehouse/ingest/` | 7 fetcher scripts | some replaced by UDMs that read sources directly |
| `warehouse/sources.yaml` | source → fetcher registry | describes the fetcher model we have moved away from |

`sources.yaml` already contains the argument for this issue, written about one source: a committed mirror can only be staler than the live table, and it was.

## Not a delete — it is referenced

Seven files point at it, including a live workflow and two skills:

```
.github/workflows/refresh-data.yml
docs/runbooks/deploy-udms.md
docs/runbooks/refresh-data.yml
docs/guides/queries.md
docs/guides/notebooks.md
skills/pyoso-analyst/SKILL.md
skills/add-data-source/SKILL.md
```

So the work is: establish which models and CSVs are still live, retire what the UDMs have replaced, repoint the runbooks and skills at the platform as the source of truth, and keep whatever genuinely has no home elsewhere.

## Suggested first step

[...]

Issue #169, opened 2026-08-09
Reconcile recorded adoption bands against the computed signal — and declare the bands first


ccerv1:
With PyPI coverage at 100 products (#165), 41 recorded adoption bands disagree with the computed one — 39 after #167 dropped two bad declarations. **They are not one problem, and the direction of the disagreement says which side is wrong.**

There is also a prerequisite that has to land first, described at the bottom: the computed band's thresholds are hardcoded in the UDM.

## The asymmetry

**Recorded *below* computed → the recorded band is wrong.** 17 products. There is no honest reason to band `mcp-python-sdk` at level 3 when the signal reports **328M** monthly downloads. Most are `stars_fallback`, which the rubric caps at 3 — so the human had no number, and one existed:

| product | recorded | computed | downloads/30d | signal_type |
|---|---|---|---|---|
| `mcp-python-sdk` | 3 | 5 | 328,478,265 | stars_fallback |
| `cohere-rerank-api` | 3 | 5 | 37,933,346 | reported_traction |
| `laminar` | 3 | 5 | 33,882,759 | usage_volume |
| `langfuse` | 3 | 5 | 25,817,410 | reported_traction |
| `docling` | 3 | 5 | 13,825,930 | stars_fallback |
| `markitdown` | 3 | 5 | 12,633,897 | stars_fallback |
| … | | | | |

**Recorded *above* computed → usually fine, but currently unfalsifiable.** 22 products. PyPI is one channel among several: `open-webui` ships mostly via Docker, `verl` and `swift` run from source, `helm` from a clone. The recorded band is probably right — but the judgment is invisible.

## Proposed rule

> A recorded adoption band may **exceed** the computed one only with a declared reason naming the uncounted channel. It may never fall **below** it.

That converts a judgment into a checkable claim, which is exactly what `establishes` did for openness and `relative_to` did for capability. It also produces a work list for free: every named channel is a candidate signal, and "Docker pulls" appearing eight times makes the case for a Docker Hub route better than any argument could.

## Sequencing, so the gate doesn't land on a backlog

`verification.md` is explicit that a big-bang gate fails on day one and gets switched off. So:

1. Fix the 17 under-banded products. Each needs a glance, not a bulk update.
2. Add the named-channel reason to the 22 over-banded ones **during each category's re-read** — the sweep is already opening those files, so this costs nothing extra.
3. Turn the check on once both are done.

## Prerequisite: the bands are hardcoded in the UDM

[...]

Issue #163, opened 2026-08-09
Route two new adoption signals: PyPI download history, and GitHub release asset counts


ccerv1:
Two sources came up during the `inference_code` verification sweep that we currently read by hand and could materialize on a schedule instead. Both are adoption signals, both are already bridged on `product_slug`, and both would change how products get banded — so this is a scoring-methodology question, not just plumbing.

Each is a route declared in `sources/signal_routing.yaml` plus an ingestion that computes it. The repo declares, the warehouse computes.

## 1. PyPI download *history*, not just a trailing window

`currentai.signal_pypi.package_downloads` carries `downloads_30d` and `downloads_7d`. That is a point read, and a point read cannot tell a collapse from a step change that happened months ago.

This has now cost us twice:

- **`llama-factory`** is sitting in `sources/verification_queue.yaml` with the reason "settling it needs the download history rather than two point reads."
- **`lmdeploy`** looked like a 165K → 50K collapse. The series shows 171,750 (May), 57,776 (June), 54,993 (July), 50,402 trailing: a single step change that has held for three months. The earlier 165,001 was a trailing window ending 2026-06-04, so it had been measuring May. The two reads never actually disagreed.

`https://pypistats.org/api/packages/<pkg>/overall?mirrors=false` returns the daily series, on a host we already fetch. A monthly rollup would let a band cite a trend rather than a reading, and would close the `llama-factory` entry without a human.

## 2. GitHub release asset download counts

This one reaches products the current routing cannot see at all.

`signal_routing.yaml` routes adoption HF → PyPI → stars, with stars marked "last resort, and explicitly last." For anything that ships binaries rather than packages there is no package to count, so it falls to stars and the rubric caps it.

**84 products sit on `stars_fallback` today, 77 of them with a GitHub repo, and 44 are capped at level 3 by that rule.**

`llamafile` is the clean case. Its adoption note says outright: "No PyPI/download or pull-count signal available (distributed as single-file binaries, not a package)." Meanwhile `https://api.github.com/repos/mozilla-ai/llamafile/releases` has been counting the whole time — **191,392 asset downloads across 30 releases**, with 3.8K–24K on each recent one. That is a real `usage_volume` figure for a product currently banded on attention rather than use.

It does a second job cheaply: it answers "does this project distribute outside PyPI," which is the question that has to be settled before reading a PyPI drop as a decline. For `lmdeploy` the answer was no — releases from v0.13.0 on attach no assets — and that had to be checked by hand.

[...]

PR #266, opened 2026-08-14
Bridge the npm and crates adoption signals, and retire signal_pypi

ccerv1:
Bridges the two unbridged package registries, so the bands over those products become recomputable instead of unfalsifiable. 11 of 13 npm products record a `usage_volume` band — they claim a download count on a registry nothing in the pipeline can read, `mcp-typescript-sdk` and `openclaw` at level 5.

Scope is npm + crates: **15 artifacts across 14 products**. Docker is not in scope and there is nothing to bridge — no product declares a Docker artifact, `registry.product_artifacts` has no `docker` kind, and the four corpus mentions are `docker-compose` self-hosting prose.

## What lands

A registry-neutral `signal_packages` dataset in three models:

| model | grain | does |
|---|---|---|
| `package_downloads_daily` (Python) | (product, artifact_kind, package, day) | fetches api.npmjs.org + crates.io, stores the daily series; a 404 is a row carrying its `http_status` |
| `package_downloads` (SQL) | (product, artifact_kind, package) | windows the series; the PyPI leg is ported from the deployed `signal_pypi` revision |
| `product_adoption` (SQL) | product | sums across declared package artifacts, bands **once** by joining `registry.adoption_bands`, abstains with a reason on partial coverage |

Neutral rather than per-registry because the `software`/`usage_volume` band declares its unit as "package downloads in the trailing 30 days, summed across declared artifacts". npm is the same unit as PyPI, so one product must get one band. `beeai` is the live case: npm 13,606 + PyPI 54,390 = 67,996.

A **daily series** rather than a point read, because a point read cannot tell a collapse from a step change that happened months ago — the defect #163 exists to fix on the PyPI side. `signal_pypi` gets history free from the marketplace dataset upstream; a fetching model has to ask for it.

Both routes stay `bridged: false` with a new `blocked_by: deploy`, so nothing tells `check_instrument` these records are recomputable until the models run. A test fails the flip if `blocked_by` is still set.

## `not_primary_channel`

New artifact key, first use of the `url` object wrapper the schema already reserved. Presence is the exemption and the value is the reason, matching `version_in_identity` and `artifact_exceptions` — one string cannot contradict itself the way a boolean plus a separate reason field can.

Two artifacts of 561 carry it: `hexabot`'s `@hexabot-ai/widget` is a sub-package rather than the platform, and `yomo`'s crate is a Rust SDK for a Go project. Both would otherwise drop 2 → 1 on a route that outranks stars. The downloads are still computed and published; they are excluded from the banded sum only, so each product falls through to `stars_fallback` legitimately rather than by override.

`artifacts_declared` / `artifacts_primary` / `artifacts_excluded` / `artifacts_counted` / `artifacts_missing` keep the three states distinguishable: a product with no primary channel reports its own abstention and never reads as complete or as not-yet-fetched.

## Two instrument changes, neither moving a level

| product | before | after | level |
|---|---|---|---|
| `opencode` | `reported_traction` | `usage_volume`, reach 1M-10M | 4, unchanged |
| `n8n` | `usage_volume`, reach 1M-10M | `reported_traction`, reach removed | 4, unchanged |

No fetch was needed and no digest was manufactured. `opencode` already cited an npm figure of 8,643,056 with a digest from 2026-08-13 and its own note had recommended this change. `n8n` already carried two digested sources, which is what `reported_traction` requires; its countable channel bands to 3 while its level-4 claim rests on Docker distribution nothing can count.

## `signal_pypi` retirement

Readers migrated: `build/check_artifacts.py`, the pypi route in `sources/signal_routing.yaml` (repointed with `filter: artifact_kind = 'pypi'`), `docs/schemas/product.schema.json`, four `sources/rubrics/*.yaml`, `skills/refresh-category/SKILL.md`, `.github/workflows/artifacts.yml`, `docs/guides/adoption.md`.

One reference left on purpose: adoption.md's "They lived only in `currentai.signal_pypi` until 2026-08-09" is a dated historical statement, and renaming it would falsify the history.

`pypi_missing` now prints `SKIPPED: … does not answer yet` and returns no findings while the new table is absent — a check that skipped, announced as such, not one that passed. Shim with a stated removal condition.

## Two pre-existing bugs fixed in passing

- **`check_routing` has been exiting 1 since 2026-08-13**, invisible because it runs in no workflow. The two `source: null` hand-authored routes read as "unknown source None". Verified against `origin/main`.
- **`serialize_registry.py` had no crates pattern**, so the deployed registry still holds `artifact_id = "https://crates.io/crates/yomo"`. The crates route could never have worked.

## Model source is not in this repo

Per the standing rule, the three `signal_packages` models are authored on the platform, with the
working copy under `currentai-org/udms/` alongside `pypi_package_downloads.sql`,
`github_repo_state.py` and the other signal models. #173 is the cleanup of the runbook claim that
says otherwise; this branch does not extend `warehouse/`.

The cost is stated rather than hidden: **nothing in CI can check the SQL's arithmetic or the
Python's contract.** The offline model tests were dropped for that reason. The one thing CI still
guards is that `hexabot` and `yomo` keep their `not_primary_channel` reason, which is the
regression that would silently re-band them — it now lives in `tests/test_serialize_registry.py`.
Everything else about these models is verified by the required post-deploy diff and by
`check_parity`.

Also not claimed: this does **not** retire the `content_sha256` stopgap. `check_verification` requires a digest and only a fetch produces one, so a computed count can corroborate a band without dating the axis. That rule change is a separate decision, recorded in `signal_routing.yaml` and `adoption.md`.

## Deploy

Nothing is deployed. The sequence is written into `docs/runbooks/deploy-udms.md`: republish the registry static model (delete + recreate, since a static model cannot gain a column by re-upload) → create the dataset → the three models, revision → release → run each → **required old-vs-new diff against `signal_pypi`** → flip `bridged: true` → drop `signal_pypi` last, marked irreversible.

The diff is expected to show band disagreements in exactly three shapes: cross-registry summation (`beeai`), no-primary-channel abstention (`hexabot`, `yomo`), and partial coverage. Anything else needs explaining before the drop. The new roll-up also filters `signal_type` where `signal_pypi` filtered only `product_type` — a latent defect that changes no current band, verified against all 23 rows of `adoption_bands`, so a future difference there reads as a finding rather than a porting error.

## Test plan

```
uv run pytest -q                     → 494 passed
build.validate                       → 0 error(s)
check_verification --verbose         → invariant [OK] digests [OK] producible-pairs [OK]
check_capability                     → [OK] 114 of 472 record a comparison
check_recipe --verbose               → [OK] 0 failure(s)
check_components                     → [OK]
check_rubric                         → reproduced
check_adoption --strict              → 452 products [OK]
check_instrument                     → 452 adoption records [OK]
serialize + check_payload            → 472 products, 257 organizations, ok
serialize_rubric --check             → nothing written
render + ast.parse + marimo check    → notebook parses
check_routing                        → adoption 347 routed / 125 blocked by bridge
```

`check_adoption --strict` and `check_instrument` both stay at zero findings after the two score edits. Generated files (`build/notebook_data.json`, `notebooks/ai-stack-map.py`) were reverted after the dry runs and are not in either commit.

Every gate was confirmed with stdin redirected; a first pass showed spurious failures on exactly the flag-carrying gates because `uv run` was consuming the driving heredoc.

Related: #173, #169, #163, #184.


ccerv1:
Held as a draft deliberately: **do not merge this before `signal_packages` is deployed.**

The repo-side reader migration in here points the pypi route at `currentai.signal_packages.package_downloads`, which does not exist yet. The `pypi_missing` shim keeps `artifacts.yml` green by skipping and saying so in the log, but a skipped check is a dark check, and this is the one that catches a declared PyPI package having vanished. Merging now would leave it dark for as long as the deploy takes.

The sequence in `docs/runbooks/deploy-udms.md` is written for the other order, and that is the order to follow:

1. Republish the registry static model — delete and recreate, since a static model cannot gain a column by re-upload, and the deployed table still holds `artifact_id = "https://crates.io/crates/yomo"`.
2. `createDataset` `signal_packages`, then the three models, revision → release → run each.
3. The **required old-vs-new diff** against `signal_pypi`, per artifact and per product. Band disagreements are expected in exactly three shapes — cross-registry summation (`beeai`), no-primary-channel abstention (`hexabot`, `yomo`), and partial coverage. Anything else needs explaining before proceeding.
4. Mark ready and merge this PR.
5. Flip `bridged: true` and drop `blocked_by` (a test fails the flip otherwise).
6. Drop `signal_pypi` last — the only irreversible step.

CI is green and it merges cleanly against `main` as of #267 landing, so this is waiting on the deploy and nothing else.

ccerv1:
Closing as superseded. The work is still wanted; this branch is no longer the way to land it.

**The branch stays** (`carl/registry-bridge`) — not deleted, so the routing table, schema and artifact-ID changes on it remain readable and can be lifted into a fresh PR.

## Why it cannot be rebased

- It targets the pre-cleanup docs layout, which #306, #307, #310 and #309 have since reshaped.
- It conflicts with refreshed product prose and scores from the 472-product closeout.
- Its `check_routing` fixes already landed elsewhere.
- It says the platform model source lives outside the repo; the staged models are now mirrored under `warehouse/platform-mirror/`.
- It predates roughly 136 tests.
- Its two irreversible steps — deploying the new package dataset and deleting `signal_pypi` — never happened, so nothing on it has been validated against a live table.

A rebase would be re-authoring it under a merge conflict, which is the worst way to review a change that ends in an irreversible deletion.

## The sequence to redo it in

1. **Correct the platform-mirror manifest** — #313, open now. It was listing `packages_product_adoption.sql` against two tables it does not implement, and missing the real `signal_packages.product_adoption` entry, so the mirror could not be trusted as the reference for step 2.
2. Confirm all three `signal_packages` model sources against the platform.
3. Deploy and run the new package dataset — **without** deleting `signal_pypi`.
4. Compare the old PyPI results against the new combined results.
5. Open a fresh PR from current `main` carrying only the still-needed routing, schema, artifact-ID, exemption and score changes.
6. Flip the routes to bridged only once the deployed tables answer correctly.
7. Delete `signal_pypi` **last**, after parity and consumer checks.

Steps 3 and 7 are the ones that cannot be undone by reverting a commit, which is why they are sequenced apart and why neither should ride on a PR that also carries repo edits.

PR #276, opened 2026-08-15
Record what a reported_traction band actually counted

ccerv1:
Adds `adoption.banded_quantity` and backfills 57 records. Routed through `docs/workflows/migrate-axis.md`. Found while auditing a random sample of 20 products, where 4 of 5 `reported_traction` records banded on something other than the product.

## The contract

**Old.** A `reported_traction` record carries a `level`, an optional word in `reach`, and a `note`. Where the level was read off something other than the product, that lived in prose.

**New.** An optional `banded_quantity` names what was counted, whenever that is not a current usage figure for the product itself.

```yaml
adoption:
  level: 4
  signal_type: reported_traction
  banded_quantity: Replit platform registered users (50M+, March 2026), not an agent active count
```

**Absent is a real answer**, not an unfilled field — the level is a judgment about the product with nothing borrowed. `claude-opus` and `claude-fable` both record a market-position reading and cite no number at all, and both correctly carry nothing.

## Why a field, when the prose was already honest

The prose *was* honest. All 19 parent-platform records describe their own substitution, several in capitals (`WHAT WAS BANDED`). It changed nothing, because a consumer reading `level: 4` off the warehouse got a number about Replit or about Perplexity's chat app with nothing attached saying so.

`active_users` solved this first: the `attribution_note` on its route in `signal_routing.yaml` requires a record to name a figure that is not an active count, and three records do. This is that rule generalized to the instrument where the substitution is most common, given a field so a machine can read it.

## The 57, by shape

| shape | n | example |
|---|---|---|
| a parent platform's reach | 19 | `azure-ai-foundry-observability` bands Microsoft Foundry |
| a real count that is not a current active one | 8 | `v0` 4M+ cumulative; `codex-cli` **weekly** actives on a monthly scale |
| a proxy of another kind | 26 | `claude-code` ~$2.5B run-rate; `openhands` 83.9k stars |
| nothing — the note cites a figure only to reject it | 4 | `perspective-api`; Jigsaw has never published one |

Every value restates something the record's own note already said. This migration researched nothing.

## No level moved

All 57 files differ from `main` by this field alone — asserted mechanically, not by eye:

```
57 score files touched
DRIFT: none — every file differs by banded_quantity alone
```

`reported_traction` level distribution is unchanged (1:2, 2:11, 3:49, 4:40, 5:7), and coverage is now 57 with, 52 without.

## The gate

`tests/test_banded_quantity.py`, strict from day one — unlike `check_adoption` and `check_instrument`, which both shipped non-strict against an existing backlog. There is no backlog: the field arrived with all 57 filled in.

It requires the field wherever a `reported_traction` note cites a **magnitude** — a suffix or thousands separators, never a bare integer, so "28 enterprise testimonials" and a `2026-08-13` date do not trip it. A parent-platform band citing no figure at all (`azure-ai-foundry-observability`, `vertex-ai-model-observability`) is the same defect and uncatchable by regex; those were backfilled by hand and nothing enforces them. Same ratchet `check_capability` uses: cover what is decidable, grow the coverage.

## migrate-axis checklist

1. Contract — above, and in `build/migrate_banded_quantity.py`'s docstring. 2. Schema — `score.schema.json`, 4 lines, no reformatting. 3. Deterministic script via `components.put_field`, idempotent (`57 changed, 0 already current` → re-run is a no-op). 4. `check_instrument` grew `unattributed_magnitudes`. 5. Serialization — see below. 8. `docs/reference/adoption.md` teaches the new shape and the checklist gained a line. 9. Distributions above. 10. Additive field; the prose stays.

**Steps 6 and 7 — assessed, and no coordination needed.** `serialize.py` passes the adoption block through wholesale, so the payload does gain the key. It is provably additive:

```
adoption keys BEFORE: confidence, last_verified, level, note, reach, signal_type, sources
adoption keys AFTER : banded_quantity, confidence, last_verified, level, note, reach, signal_type, sources
payload identical once banded_quantity and `generated` are removed: True
```

Nothing removed or renamed, so existing `aipotluck.org` readers are unaffected — but the field is there if the front end wants to surface it. `serialize_registry.py` has no adoption dependency, so the registry tables and `check_parity` are untouched and no warehouse deploy is required.

## Noticed, not fixed here

`opencode` records `reported_traction` while its own note says npm is "the product's declared and primary distribution channel" and reports 8,643,056 downloads a month. That looks like it belongs on `usage_volume`. Left alone — re-banding is a curation judgment, not a migration.

## Test plan

- `tests/test_banded_quantity.py` — zero unattributed magnitudes; the walk has not narrowed (>90 records); ≥57 carry the field; 7 parametrized cases pin what counts as a magnitude.
- Full suite: 523 passed.
- Gates green: `validate`, `check_verification`, `check_capability`, `check_recipe`, `check_rubric`, `check_components`, `check_adoption --strict`, `check_instrument`, `check_payload`.
- `build/notebook_data.json` and `notebooks/` deliberately not committed.


ccerv1:
Fixed in d046701. Confirmed: `status` and the exit code were computed from `findings` alone, so `--strict` could print an unattributed-figure violation, follow it with `[OK]`, and exit 0 — this module's own subject turned on the module itself.

Both classes now decide both. The status line names each that fired (`3 instruments unsupported, 2 figures unattributed`) and `--strict` exits 1 on either. Non-strict stays informational, matching how `check_adoption` and this gate shipped against their backlogs.

Two tests, both by planting a finding rather than by asserting the corpus is clean: strict exits 1 with no `[OK]` in the output, and non-strict still exits 0.

Suite: 525 passed.

PR #277, opened 2026-08-15
List the two classes a gate cannot judge

ccerv1:
Two findings from a 20-product audit that are real but **not gateable**. They become worklists rather than checkers — a checker that guessed at either would do damage.

## `sweep_status --retracting` — 52 axes

Every refresh appends to a note rather than rewriting it, so a note can now carry a superseded sentence *and* its own retraction, and a reader cannot tell which half is current without reading to the end. `blaxel-sandbox` is the shape:

> No last_verified claimed: a star count is volatile… **That last sentence is superseded** — the rest of the axis has now been re-read.

`transformers` is the variant: two different download figures, both present tense, nothing marking the first as stale.

```
52 note(s) correct themselves in place.
  blaxel-sandbox.adoption      'That last sentence'
  blaxel-sandbox.capability    'BENCHMARK FIGURE SUPERSEDED'
  math.adoption                'EVIDENCE REPLACED'
  the-pile.openness            'withdrawn'
  ...
```

**No gate, deliberately.** Nothing can tell a superseded sentence from a legitimate history, and one that guessed would start deleting the reasoning these files exist to carry. This lists them for the next refresh pass, where a human is already reading the note.

## `sweep_status --under-coverage` — 9 bands

Measured bands whose own note disowns the measurement:

```
  aider          understates  'primary distribution channel'
  gvisor         understates  "not the product's primary distribution"
  langflow       understates  'understates'
  mistral-large  understates  'understates'
  n8n            understates  'understates'
  ollama         understates  "not the product's primary distribution"
  perplexica     understates  'understates'
  promptfoo      understates  'minority channel'
  searxng        understates  'understates'
```

`adoption.md` names the tell — "a note that describes the signal as understating the product, followed by a band recorded on that signal anyway" — and gives the remedy per case. Inflation is the same defect pointing the other way, so `laminar` discounting a CI-inflated 17M PyPI figure belongs on the same list.

Several are already settled in the guide (`n8n` and `langflow` sum their Docker channel; re-bands were deliberately refused for `ollama`, `gvisor`, `promptfoo`). Nine is a read-list, and automating a nine-item read-list costs more than reading it. `reported_traction` records are excluded — a band claiming no measurement cannot be disowning one, and that class is #276's problem.

## Scope note

The undigested-source ratchet originally in this PR **moved to #278**, which is the PR that actually drains the tail (26 sources, 319 → 293). Leaving the budget here would have broken `main` whichever merged second. This branch now touches no `sources/` files at all.

## Test plan

- `tests/test_sweep_status.py` — three new: the known retracting cases are found; `n8n`/`mistral-large`/`ollama` (the cases `adoption.md` establishes the rule on) are found; only measured instruments are read.
- Suite green.
- Both new modes exit 0 always: neither is a finding a PR can be asked to clear.


ccerv1:
Fixed in 7a678ed. You are right and it is the sharper version of the point: those tests would have failed CI precisely when the refresh pass these worklists exist to feed did its job, and the only way to land a fix would have been to weaken the test in the same commit.

Detector behaviour moved to synthetic notes, with negative cases as well as positive ones so the patterns cannot quietly widen to match ordinary prose — `Re-read 2026-08-13 and unchanged` and a plain download count must *not* match.

What remains against the real corpus is a traversal smoke test: the walk runs, every row is well formed, every slug resolves to a file, and `under_coverage` reads only measured instruments. It asserts nothing about how many findings exist, because zero is the desired answer.

No corpus slug is named in any assertion now. Suite: 516 passed.

PR #278, opened 2026-08-15
Replace the `shows` placeholder with what each page actually shows

ccerv1:
29 sources across the **13 flagship score files** recorded their evidence as the literal string `flagship phase-C verification source`. The field exists so a reader can tell whether the recorded claim follows from the page. That string says only that a batch ran.

Files affected: `langchain`, `llama`, `qwen`, `deepseek`, `ray`, `litellm`, `crewai`, `langfuse`, `openhands`, `mistral-large`, `falcon`, `phi`, `minimax` — which is where an external reader looks first.

## Why no gate caught it

`check_verification`'s `digests` gate asks whether a source was **fetched**. `check_refetch` asks whether it is still **fetchable**. A placeholder passes both, because both read the metadata around the claim rather than the claim itself.

And it is not a legacy shape that decays on its own: **3 of the 29 (`litellm`) were re-fetched and re-digested on 2026-08-13 with the string copied forward.** A recent pass reproduced it. Hence a gate.

## What was done

Every URL fetched through `build.fetch_source` on 2026-08-15. **All 29 returned 200.** Each `shows` is written from the body that fetch returned, and each entry gains `http_status` and `content_sha256`.

That also drains 26 sources out of the undigested tail — **319 → 293** — since 26 of the 29 carried no digest at all.

The new digests reproduce where the body is stable: `check_refetch` confirms 11 byte-exact across `qwen`/`ray`/`langchain`/`crewai`/`litellm`, with the rest drifting because a GitHub repo page carries a star count.

Two entries are deliberately **not** claims of support:

- `llm-stats.com/models/qwen3.6-plus` returns 200 and serves a *"Quick verification — Confirm you're human"* interstitial rather than the model page. Its `shows` records exactly that and claims nothing.
- The VentureBeat Falcon article is left untouched. It has answered 429 on three separate attempts, and `refresh-category.md` is explicit: an unreachable host means the axis stays undated and the host is reported, never that a digest gets invented.

## `components.set_source`

A `sources:` entry is a list item, not a mapping field, so `set_field` and `add_field` cannot reach one — both locate a key by name inside an axis block, and every entry carries the same key names. `set_source` finds the entry citing a URL and re-renders it whole through the dumper, with the same reparse assertion `set_field` uses. The corpus warns that a hand-rolled line splice already shipped a seven-product defect; this is that discipline applied one level down.

## Gates

- `check_verification` grows a fourth gate, `placeholder-shows`. Verified it bites by planting a marker and watching it fail, not just by watching it pass on a clean corpus — `tests/test_placeholder_shows.py` makes that a permanent test.
- The digest ratchet moves here from #277 with its budget lowered to 293. **The PR that drains the tail should own the budget** — leaving it on #277 would have broken `main` whichever merged second.

## Flagged, not fixed

**`falcon.adoption` looks mis-banded.** It records level 1 (`<10K`) on 5,292 downloads of its single declared artifact `Falcon3-10B-Base`. But `Falcon3-7B-Base` — cited in the same record, and the source this PR had to read — reports **14,911 downloads a month**. Together that is 20,203, which is level 2. `adoption.md` declares `sum_across_artifacts: true` because the map's unit is the product family.

Re-banding is a curation call, not an evidence repair, so this PR records the honest figure and leaves the level alone. The contradiction is now visible in the record instead of hidden behind a placeholder — which is the point of the change. The product's own `comments` already says "flagged for review".

## Test plan

- `tests/test_placeholder_shows.py` (new) — corpus is clean; the gate catches a planted marker; a real `shows` passes.
- `tests/test_digest_ratchet.py` (moved from #277, budget 293) — tail does not grow; budget is not slack; the June-cohort shape still holds.
- Full suite: 519 passed.
- Gates green: `validate`, `check_verification` (all four), `check_capability`, `check_recipe`, `check_rubric`, `check_components`, `check_adoption --strict`, `check_payload`.
- `check_routing` fails on this branch — it is red on `main` and fixed in #275.


ccerv1:
All three addressed in 1ebc343.

**1 — the gate missed four prefixed markers.** Confirmed: `deepseek`, `falcon`, `minimax` and `phi` all append their substance to the marker rather than replacing it, so the exact-match predicate reported clean while the string it prohibits was still in the file.

Fixed both ways you suggested, because either alone leaves a hole. The gate now matches the marker as a **substring of the normalized value** — so a half-edit *and* a rewrap both fail — and the four records are rewritten to drop the batch language while keeping the honest half:

```
shows: 'NOT RE-READ: the host answered HTTP 429 to the 2026-08-13 attempt after retries,
  so this source was not re-confirmed and is not part of this axis''s date.'
```

They keep their original `accessed` dates. Nothing was re-read, and re-dating them would be the overstatement the field exists to prevent. Two new tests pin the prefix case and the wrapping case.

**2 — Falcon.** You are right, and I had it backwards: I treated recording the contradiction as sufficient when the contradiction sits *underneath a confirmation*. `last_verified: 2026-08-14` asserts somebody confirmed a value that the same axis now disproves, which is the one state a date may not sit on.

Took your third option — cleared the date and queued the axis — because the other two are identity rulings I should not make unilaterally. The queue entry names all three ways out (declare the family artifacts and re-band on the sum; narrow the product to the 10B checkpoint; or establish that the single declared artifact is deliberately what the band reads). `test_verification_queue_consistency` enforces held-XOR-dated, so the two states cannot drift apart.

This needed `components.drop_field` — the inverse of `add_field`, same reparse assertion. `migrate-axis.md` requires a deletion to go through that module rather than a hand splice, and there was no delete operation.

**5 — the ratchet.** Also right, and the swap case is the one I missed: a count only prevents *net* growth, so digesting one legacy source while citing a new undigested one leaves the number unmoved.

Replaced with your allowlist: `sources/undigested_legacy.txt` lists the tail entry by entry as `slug|axis|url`, 293 lines. Anything not on it must carry a digest, so a newly cited source fails on the first assertion rather than being absorbed. Verified by planting exactly the swap you describe:

```
E  ray|openness|https://example.com/newly-cited
E  assert not ['ray|openness|https://example.com/newly-cited']
```

A second test fails on a line that no longer describes an undigested source, so the list drains rather than rotting into a standing exemption — deleting the line is how a refresh pass records the work. I dropped the "recent cohort" threshold from 39 to 10; it is now only a canary on the claim in the docstring, since the allowlist is what carries the invariant.

Suite: 522 passed. All gates green.

PR #279, opened 2026-08-15
Invert the artifact-kind check so a new kind fails by default

### LINKED ISSUE #184: Invert the artifact-kind routing test so it catches new kinds by default

ccerv1:
Closes #184.

## The guard was weaker than it looked

The old test built its "in use" set by intersecting each product record against a hardcoded `ARTIFACT_KINDS`:

```python
in_use = {k for k in ARTIFACT_KINDS if record.get(k)}
assert in_use <= declared
```

So it could only catch an undeclared kind **whose name was already in that set** — which is the one case that cannot happen. A genuinely new kind never enters `in_use` and the assertion passes. That is the regression the test is nominally there to prevent.

`arxiv` was the live proof: 24 products declare it, and it was in neither the set nor `sources:`.

## The inverted form

Enumerate every top-level product key, subtract an explicit `METADATA_KEYS` allowlist, require the remainder to be a declared artifact kind. Verified by planting a `docker:` key on a product record:

```
caught: planted: ['docker']
```

The allowlist is now the visible thing a contributor edits to opt something out — the right polarity. Four supporting tests: the allowlist may not shadow a declared kind, may not accumulate dead entries, the walk must read the whole corpus, and unreachable-but-declared kinds (`npm`, `crates`) stay visible rather than silent.

## `arxiv` needs no ruling — it is a layering distinction

The issue expected a taxonomy decision. It needs none:

- **`arxiv` is a product ARTIFACT KEY.**
- **`semanticscholar` is the SOURCE that consumes it**, and it already declares `artifact_key: arxiv`.

```
semanticscholar   artifact_key='arxiv'   bridged=True
declared artifact_key set: arxiv, crates, github, huggingface_dataset,
                           huggingface_model, npm, pypi
```

Comparing against **source names** is what made `arxiv` look undeclared. Declaring a `sources.arxiv` entry — the obvious fix from the issue's framing — would have created a duplicate for something the router already reaches. `test_arxiv_is_covered_by_the_source_that_consumes_it` pins both halves so neither regresses.

Product keys today, all covered: `github` 254, `pypi` 106, `huggingface_dataset` 66, `huggingface_model` 58, `arxiv` 24, `npm` 13, `crates` 1.

`artifact_exceptions` is the subtle allowlist entry — it is an annotation *about* artifacts (why `pymilvus`'s downloads stand in for the Milvus server) rather than an artifact.

## The same conflation, one layer over

`check_routing` carried a hardcoded `SOURCE_ARTIFACT` dict: a second copy of a fact the table declares, which is the drift `check_parity` exists to catch. **It had already fallen behind** — five sources where the table declares seven, missing `npm` and `crates`. Both are `bridged: false`, so the omission was inert; it would have stopped being inert the day a bridge landed, which is exactly when nobody would think to look here.

It now reads `artifact_key` from the table. Coverage output is **byte-identical** before and after the swap.

## Note on #266

This edits `tests/test_signal_routing.py`, which #266 also touches. #266 is a draft held on an undeployed warehouse sequence and **already conflicts with `main`** after #275–#278, so it needs a rebase regardless — this does not add a blocker it did not have. Its two added tests should rebase cleanly onto the inverted form; the conflict is in the file's header region, not in what it asserts.

## Test plan

- `tests/test_signal_routing.py` — rewritten, 6 tests.
- `tests/test_check_routing.py` — 2 added: the mapping is read from the table, and a bridged source with no `artifact_key` still fails the structural check.
- Full suite: 551 passed.
- Gates green: `validate`, `check_verification`, `check_capability`, `check_recipe`, `check_rubric`, `check_components`, `check_instrument`, `check_payload`, `check_routing`.

### DIFF
--- a/build/check_routing.py
+++ b/build/check_routing.py
@@ -32,14 +32,25 @@ import yaml
 
 ROOT = Path(__file__).resolve().parents[1]
 
-# Which artifact key in a product YAML satisfies which routing source.
-SOURCE_ARTIFACT = {
-    "github": "github",
-    "huggingface_model": "huggingface_model",
-    "huggingface_dataset": "huggingface_dataset",
-    "pypi": "pypi",
-    "semanticscholar": "arxiv",  # citations are reached via the paper id
-}
+def source_artifact(sources: dict) -> dict[str, str]:
+    """source name -> the product artifact key it consumes, read from the routing table.
+
+    This was a hardcoded dict until 2026-08-15, and it was a second copy of a fact the table
+    already declares — the repo mirroring a declaration by hand, which is the drift
+    `check_parity` exists to catch one layer over. The copy had already fallen behind: it
+    named five sources where the table declares seven, missing `npm` and `crates`. Both are
+    `bridged: false` so the omission was inert, and it would have stopped being inert the day
+    a bridge landed, which is exactly when nobody would be looking here.
+
+    The two layers are distinct and worth naming: a SOURCE is a signal producer
+    (`semanticscholar`), an ARTIFACT KEY is what a product declares (`arxiv`). One source may
+    consume a key that shares no name with it.
+    """
+    return {
+        name: source["artifact_key"]
+        for name, source in sources.items()
+        if source.get("artifact_key")
+    }
 
 
 def artifacts_of(product: dict) -> set[str]:
@@ -103,6 +114,7 @@ def main(argv: list[str] | None = None) -> int:
     routing, products, _, category_of = load()
     sources = routing.get("sources") or {}
     dimensions = routing.get("dimensions") or {}
+    artifact_of = source_artifact(sources)
 
     # Structural check first: a route pointing at nothing is a bug in the table.
     problems: list[str] = []
@@ -118,7 +130,7 @@ def main(argv: list[str] | None = None) -> int:
                 continue
             # An unbridged source can never resolve to a product, so it needs no
             # artifact mapping. Only a source claiming to be usable must have one.
-            if sources[name].get("bridged") and name not in SOURCE_ARTIFACT:
+            if sources[name].get("bridged") and name not in artifact_of:
                 problems.append(f"{dim}: bridged source {name!r} has no artifact mapping in this checker")
     if problems:
         print("ROUTING TABLE PROBLEMS")
@@ -150,7 +162,7 @@ def main(argv: list[str] | None = None) -> int:
                     # map a bridge could still buy.
                     was_blocked = was_blocked or why == "unbridged"
                     continue
-                if SOURCE_ARTIFACT.get(route["source"]) in arts:
+                if artifact_of.get(route["source"]) in arts:
                     hit = True
                     break
             if hit:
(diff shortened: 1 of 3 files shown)


ccerv1:
Both fixed in the latest push. The first one is the sharper of the two and I had it half-done.

**Blocking — coverage vocabulary.** Confirmed by reproduction. `check_routing` held **two** hardcoded copies of the artifact vocabulary and my first pass replaced only `SOURCE_ARTIFACT`, leaving `artifacts_of` — which is the more consequential half. With a synthetic `rubygems` declared correctly at every layer:

```
product declares rubygems: True
artifacts_of() sees      : ['github', 'pypi']
-> rubygems visible to coverage? False
```

So a correctly declared kind passed the new invariant and still reported as research rather than routed — the map understating its own automation, with nothing failing. That is the same defect the PR claims to fix, one function down, which is a poor place to have stopped.

It now derives from `source_artifact(...).values()`. Two tests: the `rubygems` reproduction end to end, and one pinning that `artifacts_of` reads **only** the kinds it is handed, so the literal list cannot creep back as a default.

**Non-blocking — arXiv topology.** Also right, and the failure mode is worth naming: that assertion would have made a legitimate future change *look like a regression*, which is the most expensive kind of over-constrained test. The invariant is now the mapping — some source declares `artifact_key: arxiv`, and today that source is `semanticscholar` — rather than the absence of a source name. Verified that adding a direct `arxiv` source alongside Semantic Scholar now passes.

Suite: 553 passed. `check_routing` coverage output unchanged. `git diff --check` clean.

PR #280, opened 2026-08-15
A held axis must not publish as a verified product

ccerv1:
The nine verification holds are honest inside the repo and were **invisible outside it**. Found while re-reading the payload rather than the score files.

## The defect

`freshness_payload._last_verified` took `max()` over the axes that *had* a `last_verified` and ignored the ones that did not — under a comment claiming the result was *"the date on which everything in the score was last standing"*. For a product with a held axis that sentence is false.

Three shipped that way:

| product | held axis | published |
|---|---|---|
| `falcon` | `adoption` — held because its evidence contradicts its own band | `{date: 2026-08-13, basis: verified}` |
| `qualcomm-ai-engine-direct` | `openness` | `{date: 2026-08-09, basis: verified}` |
| `aws-neuron` | `adoption` | `{date: 2026-08-11, basis: verified}` |

`falcon` is the sharpest: #278 cleared its date and queued the axis *precisely because* a fresh fetch disproved the band — and the payload went on calling the product verified.

**No gate could see this.** Every checker reads the per-axis dates, where the state is correct. Only the payload does the reduction, and nothing checked the reduction.

## The fix

A third basis:

```json
{
  "date": "2026-08-13",
  "basis": "partial",
  "unconfirmed_axes": ["adoption"],
  "verification_holds": [
    {"axis": "adoption", "since": "2026-08-15", "reason": "The band and the product's identity disagree…"}
  ]
}
```

A held axis is a real editorial state and must not be forced into a score to make a label tidy. So the product stays **publishable and visibly caveated** — the page can show "partially verified" and the reason, instead of a product-level date that quietly covers an axis nobody confirmed.

**`partial` follows from the axis being unconfirmed, not from the queue entry.** A hold *explains* an unconfirmed axis; its absence does not make one confirmed. An undated axis with no queue entry is still `partial`, and is separately a finding for `check_freshness` and `sweep_status`. There is a test for that case specifically.

## Blast radius

```
basis distribution: {'verified': 463, 'partial': 9}
```

Exactly 9 move `verified → partial`, matching the 9 held axes — nothing reclassified by accident. `commit` and `verified` are unchanged in meaning. Two new optional keys, so the payload change is additive; `render` still parses.

## Front end

**Checked, and the situation is not what this PR originally assumed.** `aipotluck.org` consumes `freshness` **nowhere** — one hit in `web/src`, and it is a comment in `sitemap.xml/+server.ts` about deliberately *not* asserting freshness. The `basis: string` in `lib/types/scores.ts` is `capability.basis`, a different field.

So there is no enum to migrate and no consumer to break: this PR makes the data available without yet making it visible. The caveat is currently invisible to users not because the front end would mislabel `partial`, but because it renders no freshness at all — every product looks equally current, and always has.

That is a larger gap than this PR opened and it is separate UX work, tracked as its own issue against `aipotluck.org`. Merging the producer contract now makes it truthful without coupling a safe data correction to that work.

## Test plan

- `tests/test_freshness_payload.py` — 5 new: an undated axis makes a product `partial`; all-dated is still `verified`; a held axis carries its reason; an undated axis with **no** queue entry is still `partial`; no dated axis still falls back to `commit`.
- Three existing tests updated for the renamed helper — `_axis_dates` now returns confirmed **and** unconfirmed axes, and the rewritten test asserts the second half, since dropping it is exactly what caused this.
- `check_payload` accepts the new basis.
- Full suite: 549 passed. All 9 gates green. `render` parses.

### DIFF
--- a/build/check_payload.py
+++ b/build/check_payload.py
@@ -10,12 +10,68 @@ row must never read as "fine". A gate that goes green when it cannot understand 
 is worse than no gate at all: it looks like protection nobody is actually getting.
 """
 import json
+import re
 import sys
 from pathlib import Path
 
 ROOT = Path(__file__).resolve().parents[1]
 
 _ALIAS_KINDS = ("products", "organizations")
+_AXES = ("openness", "adoption", "capability")
+_DATE = re.compile(r"^\d{4}-\d{2}-\d{2}$")
+
+
+def _check_freshness_caveats(slug: str, fresh: dict) -> None:
+    """Gate the SHAPE of a partial freshness record, not just its presence.
+
+    The defect this whole mechanism exists for was a lossy reduction that no payload gate
+    could see. Adding `partial` without gating it would leave exactly that hole open one
+    shape over: `{"date": ..., "basis": "partial"}` with nothing saying what is partial is
+    strictly worse than `verified`, because it advertises a caveat and then withholds it.
+    """
+    basis = fresh.get("basis")
+    unconfirmed = fresh.get("unconfirmed_axes")
+    holds = fresh.get("verification_holds")
+
+    if basis == "partial" and not unconfirmed:
+        raise PayloadError(
+            f"{slug!r} is basis 'partial' with no unconfirmed_axes — a caveat that does not "
+            "say what it is about"
+        )
+    if basis == "verified" and (unconfirmed or holds):
+        raise PayloadError(
+            f"{slug!r} is basis 'verified' but carries {'unconfirmed_axes' if unconfirmed else 'verification_holds'}; "
+            "an axis cannot be both confirmed and outstanding"
+        )
+
+    if unconfirmed is not None:
+        if not isinstance(unconfirmed, list) or not unconfirmed:
+            raise PayloadError(f"{slug!r} has a non-list or empty unconfirmed_axes")
+        if len(set(unconfirmed)) != len(unconfirmed):
+            raise PayloadError(f"{slug!r} repeats an axis in unconfirmed_axes: {unconfirmed}")
+        unknown = [a for a in unconfirmed if a not in _AXES]
+        if unknown:
+            raise PayloadError(f"{slug!r} names unknown axes in unconfirmed_axes: {unknown}")
+
+    if holds is not None:
+        if not isinstance(holds, list) or not holds:
+            raise PayloadError(f"{slug!r} has a non-list or empty verification_holds")
+        for hold in holds:
+            axis = (hold or {}).get("axis")
+            if axis not in _AXES:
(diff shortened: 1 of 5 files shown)


ccerv1:
All four addressed. Three were code and are pushed; the fourth turns out to need a different fix than expected, explained below.

**1 — the product date was the newest axis.** Confirmed at 176/472, and you're right that it is the same overstatement in a less obvious form. It is also contradicted by this repo's own rule one section up: `last_verified` is the date on which *everything* was confirmed, so at product level "everything" is the constraint and the oldest confirmed axis is the only defensible answer. Now `min()`, with `latest_axis_confirmation` carrying the newest where it differs. The same rule applies to the confirmed subset of a `partial` record.

**3 — a fully unconfirmed product dropped its hold.** Reproduced exactly as you describe. Holds now attach on *every* path with an unconfirmed axis, `commit` included — I kept `commit` rather than adding an `unverified` basis, for the reason you gave. One addition: a hold naming an axis that has *since been confirmed* is no longer emitted at all, since `check_payload` now rejects that shape and a stale queue entry would otherwise become a hard failure.

**4 — the gate.** This is the one worth calling out, because it is my own diagnosis turned back on me: the PR argues that no payload-level gate could see the lossy reduction, and I then added the field without adding the gate. `_check_freshness_caveats` now enforces all seven invariants you listed. Verified each by construction — 9 malformed shapes rejected, 4 valid ones (including `partial` with no queue entry, and `commit` carrying holds) accepted.

**2 — the front end. The situation is different from what we both assumed, and the difference matters.**

I checked `aipotluck.org`. It does not consume `freshness` **at all** — one hit in `web/src`, and it is a comment in `sitemap.xml/+server.ts` about *not* asserting freshness. The `basis: string` in `lib/types/scores.ts` is `capability.basis`, a different field. There is no enum to migrate and no consumer to break.

So the rollout risk you identified does not exist today, but the underlying point survives in a stronger form: **the caveat is invisible to users not because the front end mislabels `partial`, but because it renders no freshness at all.** Every product looks equally current, and always has. That is a bigger gap than this PR opened and it is not one I should close inside a payload change.

My read: this PR is safe to merge on the payload side — additive fields, one new enum value nothing reads, gated shape — and the front-end work is a real, separate piece of work that should be its own issue against `aipotluck.org`: surface freshness, render `partial` as "Partially verified", show the unconfirmed axes and the hold reason. Happy to open that issue, or to hold this PR until it lands if you would rather they ship together. Your call — I would not want to merge on "nothing breaks" if the intent was that users see the caveat now.

Suite: 567 passed. All gates green, `render` parses, distribution still 463 verified / 9 partial.

PR #282, opened 2026-08-15
Canonical verification line for the 44 that already named a source

ccerv1:
First slice of the product-prose closeout. Safe half only — the half where nothing has to be re-read.

## The 245 were never unverified

Measured across all 472: **zero products lack a verification line.** Every one carries a dated verification, all in August. The gap `sweep_status` reports is a **formatting** gap, and it splits four ways:

| class | n | shape | fixable by rewording? |
|---|---|---|---|
| canonical | 227 | `Verified <date> via <doc>.` | — |
| **this PR** | **44** | `Verified **live** <date> via <real doc>` | **yes — one word** |
| generic | 172 | names a **method**: `primary sources`, `substitute sources` | **no** |
| other noncanonical | 29 | remaining variants | per-item |

## What this PR does

These 44 already name a document — `amazon-nova` reads *"the AWS Nova user guide and the Nova models page"*, `codegemma` *"the HF model card and Hugging Face's CodeGemma release post"*. Dropping `live` makes them canonical without touching what they claim.

Applied through `components.set_document_field`, which re-renders the field at the corpus width, so each file also reflows the paragraph the word came out of. That is the sanctioned helper — the alternative is a hand splice, which this repo has already shipped a seven-product defect through. Proved mechanically that the reflow is all it is:

```
files: 44
DRIFT: none — every file differs by the single word 'live' and whitespace reflow alone
```

Canonical lines: **227 → 271 of 472.**

## Four products removed from this PR during self-review

It originally covered 48. Four were pulled back out:

```
hailo-10h                      via substitute sources - hailo.ai answers 403 to every request
hailo-8                        via substitute sources - hailo.ai answers 403 ...
nxp-imx-8m-plus                via substitute sources - www.nxp.com answers 404 ...
qualcomm-dragonwing-rb3-gen-2  via substitute sources - www.qualcomm.com serves a client shell
```

Removing `live` would have promoted these **into** canonical form while they name `substitute sources` — a method, or more precisely an account of a method failing. `product-copy.md` forbids it for the same reason it forbids `primary sources`: it leaves the next editor nothing to re-open.

The filter here excluded only trailers beginning `primary sources`, which was narrower than the rule it implemented. The four now go to the evidence-matching pass in #283 with the other generic records. Verified after the revert: **0 canonical-form lines name a method.**

## Why the rest is not a formatting problem

172 products say `Verified live <date> via primary sources`. Rewording cannot repair them — something has to supply the document. That is #283's subject, and none of it is in this PR.

## Test plan

- Every non-`comments` field byte-identical across all 44; `comments` differs by the removed word alone.
- `validate`, `check_payload`, `check_verification` green; `test_product_prose` green.
- Full suite: 581 passed.
- No score file touched — this changes editorial provenance for prose, never a `last_verified`.

### DIFF
--- a/sources/products/amazon-nova.yaml
+++ b/sources/products/amazon-nova.yaml
@@ -10,8 +10,7 @@ description: Amazon's proprietary multimodal model line in AWS Bedrock. Current 
   pricing for high-volume AWS-native enterprise deployments with SLA guarantees.
 comments: Amazon Nova Pro (amazon.nova-pro-v1:0), released 3 Dec 2024; multimodal understanding/instruct model
   (text+image+video in, text out), 300K context. Proprietary, API-only via Amazon Bedrock. Succeeded by Amazon
-  Nova 2 family (2026) but Nova Pro v1 remains in service. AWS publishes no standardized benchmark scores
-  for Nova, so the capability record deliberately abstains. Verified live 2026-08-13 via the AWS Nova user
-  guide and the Nova models page. Consolidated
-  on 2026-07-29 from amazon-nova; openness follows amazon-nova-pro, the release that currently governs. See
-  docs/reference/identity.md.
+  Nova 2 family (2026) but Nova Pro v1 remains in service. AWS publishes no standardized benchmark scores for
+  Nova, so the capability record deliberately abstains. Verified 2026-08-13 via the AWS Nova user guide and
+  the Nova models page. Consolidated on 2026-07-29 from amazon-nova; openness follows amazon-nova-pro, the
+  release that currently governs. See docs/reference/identity.md.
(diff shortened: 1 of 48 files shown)


ccerv1:
Self-review caught a defect in this PR before review did — pushed a fix, **48 → 44 files**.

Removing `live` from four hardware records would have promoted them *into* canonical form while they name `substitute sources`:

```
hailo-10h                      via substitute sources - hailo.ai answers 403 to every request
hailo-8                        via substitute sources - hailo.ai answers 403 ...
nxp-imx-8m-plus                via substitute sources - www.nxp.com answers 404 ...
qualcomm-dragonwing-rb3-gen-2  via substitute sources - www.qualcomm.com serves a client shell
```

That is a method — or more precisely an account of a method failing — and `product-copy.md` forbids it for the same reason it forbids `primary sources`: it leaves the next editor nothing to re-open. This PR's filter excluded only trailers beginning `primary sources`, which is narrower than the rule it was implementing.

Exactly the failure mode you warned about on the Class B question: **converting a vague claim into invented precision.** The four now go to the evidence-matching pass with the other generic records.

The remaining 44 all name a real document. Verified after the revert: 0 canonical-form lines name a method.

PR #283, opened 2026-08-15
Classify prose verification provenance, and gather the evidence to repair it

ccerv1:
Second slice of the prose closeout, following the agreed sequence: **classification reviewable on its own, before any of the 172 product edits.** No product file is touched here.

## The 245 were never unverified

`sweep_status` collapsed four materially different states into one number. Measured across all 472:

```
canonical            227   Verified <date> via <document>.
named_noncanonical    73   names a real document behind live / on / against
generic              172   names a METHOD — "primary sources", "research"
missing                0
```

**No product is missing a verification line, and every date is in August.** What varies is whether the line names something a reader can reopen — which is a different problem with a different remedy, and reporting them as one number hid which was which.

`generic` is the substantive class, and `product-copy.md` rules it out by name:

> **Never name a method.** `web search`, `research`, and the bare `primary sources` describe how you looked rather than what settled it, and leave the next editor nothing to re-open.

## `build/prose_provenance.py` — classify and gather

For each `generic` product it collects the sources that product's **own score file** records as accessed **on the date the prose line already claims**, with each source's recorded `shows`.

**All 172 have at least one date-aligned source**, so date alignment decides nothing by itself. The judgment is entirely whether the source is a nameable, authoritative document that supports the prose — and this module deliberately does not make it. It emits a packet with the candidates and two mechanical hints: whether the URL looks like a document at all (a `pypistats` or `api.github.com` response does not), and whether its `shows` is a bare measurement.

Three real cases show the range:

| product | date-aligned evidence | reading |
|---|---|---|
| `anythingllm` | the repo page — *"all-in-one local-first RAG/agent app… 40+ LLM providers, RAG, multi-user permissioning, no-code agents"* | supports the description |
| `claude-ai` | `claude.com/product/overview` — *"Artifacts, Projects, Web Search, Voice, Code Execution"* | supports the description |
| `litellm` | only `pypistats.org/api/…/recent` — a download count | **cannot** say what the product is |

`litellm` is the one packet the mechanical hint already marks `reread`, and it matches the manual reading exactly.

## `build/apply_provenance.py` — deterministic, and refuses

Writes only packets marked `derive`, and refuses any that cannot justify itself:

- no `selected_document`, no `support` sentence, or a `verification_date` that is not a date;
- `source_accessed` that disagrees with the date the prose already claims;
- a `selected_document` that is itself a method;
- **a `source_url` not present in that product's score file** — checked against the corpus rather than trusted from the packet. That is the one way this could launder an invented document, so it is the one check that does not take the packet's word for anything.

**It never changes the date.** This repairs the record of *what was read*; advancing the date would claim a re-verification nobody performed. It never touches a score file either — `product-copy.md` is explicit that writing the comments line does not earn a `last_verified`.

All four refusals verified by construction, and a fifth test pins that `Refused` propagates rather than falling back to writing something else.

## `sweep_status` now reports the four states

```
prose verification lines: 227 verified, 73 named_noncanonical, 172 generic
  `generic` names a method rather than a document (product-copy.md forbids it);
  `named_noncanonical` names a real document behind `live`/`on`/`against`.
```

## What is deliberately not here

The 172 decisions. Each is a reading of whether a specific document supports a specific description, and putting them in the same PR as the machinery would make both unreviewable. They follow in batches, by category, starting with `ui_api`.

`build/prose_provenance_packets.yaml` (313KB) is committed as the reviewable artifact — every candidate, its `shows`, and `decision: null` awaiting judgment.

## Test plan

- `tests/test_prose_provenance.py` — 20 tests: 8 classification cases across all four states; the corpus has no `missing` (if that ever fails the framing changes); rewrite keeps surrounding prose and never moves the date; 6 parametrized refusals; refusal propagates.
- Full suite: 601 passed.
- `validate`, `check_payload`, `check_verification` green. No product or score file modified.


ccerv1:
Pushed a fix for a defect in `classify()` that #282's self-review surfaced.

`classify()` tested `CANONICAL` **first** and returned, so:

```
Verified 2026-08-13 via primary sources.   ->  canonical
Verified 2026-08-13 via web search.        ->  canonical
```

The correct shape wrapped around the exact thing this classifier exists to find. It is this module's own subject turned on itself — a check grading on form while the defect sits in the payload, which is the pattern running through this whole audit.

**Latent on the corpus** (zero canonical-form lines name a method today) and **not latent in the migration**: #282 removed `live` from four `via substitute sources` records, promoting them into canonical form. Reverted there.

`generic` now outranks `canonical` whenever the trailer is a method, however the line is worded. The census is unchanged — 227 / 73 / 172 / 0 — because the corpus had no instances; the fix protects the invariant rather than correcting a count. Added four parametrized cases for the canonical-form-naming-a-method shape, plus a corpus-wide invariant test so a future rewording pass cannot manufacture one.

Suite: 606 passed.

ccerv1:
All five addressed. Every one was a check asserting more than it established — which is the shape this module was written to find, four more times in the module itself.

**1 — `named_noncanonical` promised a document.** Right, and 16 of the 73 name nothing. New state `ambiguous_noncanonical` for dated-but-unnameable; `meta-ai` and `le-chat` land there. The trailer regex now requires the preposition rather than treating the rest of the sentence as a document name.

Census is five-state:
```
canonical 227 · named_noncanonical 57 · ambiguous_noncanonical 16 · generic 172 · missing 0
```

Also took the two adjacent cases you flagged: `Verified 2026-08-13 via .` is no longer canonical (a document name must contain a word), and `METHOD_WORDS` allows a leading article so `the primary sources` no longer reads as a document.

**2 — packets not bound to the reviewed prose.** The most serious of the five, and your reproduction is exact:

```python
rewrite("Verified live 2026-08-12 via primary sources.", "2026-08-13", "the README")
#   ->  "Verified 2026-08-13 via the README."
```

The module's own docstring says it never changes the date, and it did. Packets now carry a `prose_digest` over normalized `description` + `comments` with the dated clause removed — both fields, since `product-copy.md` defines the line as provenance for both. The applier recomputes it, and separately re-checks that the live line is still `generic` and still carries the packet's date. Two tests pin the digest's boundaries: it ignores the clause it is about to rewrite (or every packet would invalidate itself on application) and catches an edit to either prose field.

**3 — sibling method vocabulary.** Confirmed: `selected_document: substitute sources` passed. That is the fourth time this session a fix has missed its twin, and the second involving this exact phrase. Now imports `METHOD_WORDS` from the classifier; no second regex.

**4 — the date bug again.** Shape-matching accepted `2026-99-99` and `2026-02-30`. Same defect I fixed in `check_payload` two changes ago and reintroduced here. `date.fromisoformat()`, both cases tested.

**5 — the 313KB artifact is gone.** You are right that it is wholly derived and would drift with no parity gate. Generated on demand and gitignored. Decision manifests carrying actual judgment will live at `sources/provenance/<category>.yaml`, split by category as you suggest — that is also what makes the `ui_api` batch reviewable on its own. Stale "166" references corrected to 172.

Eight refusals now verified by construction: substitute-sources, the-primary-sources, empty document, both impossible dates, stale digest, date drift against the live line, and a URL absent from the score file.

Suite: 621 passed. Diff is back to ~5 files.

PR #284, opened 2026-08-15
One owner per vocabulary, and a ratchet against the sibling copy

ccerv1:
Bounded sweep for the construct behind four consecutive review findings. No product, score or provenance decisions; no warehouse changes; no general refactor.

## Why

None of the four failed loudly. Each reported success over a **narrower question than the one it claimed to answer**:

| defect | what it actually answered |
|---|---|
| `check_routing.SOURCE_ARTIFACT` named five sources; the table declared seven | "is this one of the five I know about?" |
| `check_routing.artifacts_of` enumerated the same vocabulary again, same file | "does it declare one of seven hardcoded keys?" |
| `apply_provenance` reimplemented `METHOD_WORDS` more narrowly | "is this one of three forbidden phrases?" |
| `check_payload` and `apply_provenance` each validated a date by shape | "does this string look like a date?" |

Three were found by review, and the fourth only because the third was. The pattern isn't carelessness about a particular constant — it's that the fix lands on the instance in front of you and the sibling is never searched for.

## Fixed

**Axis triple — ten modules, no owner.** Seven named constants plus three inline literals, including `validate.py`, the schema gate itself. `build/vocabulary.axes()` derives from `score.schema.json`, which lists them as properties and requires all three.

None had drifted; no fourth axis has ever been added. A fourth axis is also precisely the change that would narrow ten denominators at once, and *a walk that quietly skips an axis passes green*. **This is the one entry fixed before it failed rather than after.**

**`propose_artifacts.VERIFIABLE_KINDS`** — the third instance of the `check_routing` construct, and **already divergent** (missing `arxiv`). Now derived, with the exclusion written down rather than implied:

```python
NOT_A_DISTRIBUTION_ARTIFACT = frozenset({"arxiv"})
VERIFIABLE_KINDS = tuple(sorted(artifact_kinds() - NOT_A_DISTRIBUTION_ARTIFACT))
```

Behavior identical — the resulting set equals the old literal. `arxiv` is excluded because this constant asks whether a product carries an artifact a proposal would *target*, and a paper id is not a distribution artifact. Whether it should count is a curation question, not a sweep one.

## Intentionally distinct, with the reason recorded

So the next sweep doesn't rediscover them as findings:

- **Openness class→bucket map**, duplicated three ways — already gated by `tests/test_openness_buckets.py`, which asserts all three agree. That is this sweep's own protection, arrived at first; moving the constants would churn a working invariant for symmetry.
- **`check_adoption._DOWNLOAD_INSTRUMENTS`** — not a mirror of the `signal_type` enum but a named subset, with the module explaining why `unknown` is excluded. A narrower vocabulary with a written reason is a decision.
- **`check_capability.DELTA`** — maps relations to integer offsets, which the schema cannot express, so the dict stays. A test now pins that it agrees with the schema about which relations *exist*.

## Swept and clean

- **Date validators**: no live shape-matcher remains.
- **Exit narrower than the printed report**: read every checker. `check_verification` iterates all gates so its exit covers everything it prints; `check_rubric`'s non-gating lines are marked `~` with a written reason. No instances beyond the one already fixed in #276.
- **Malformed input becoming a pass**: the three `except` paths in `build/` each record the failure rather than falling through.

## Artifact

`docs/reference/sibling-invariants.md` — the full inventory, one entry per construct with `canonical_owner`, `duplicates`, `risk`, `disposition`, `regression_test`.

## Test plan

`tests/test_vocabulary_siblings.py` — 8 tests: axes come from the schema; **no module holds a private copy**; artifact kinds come from the routing table (including the source-vs-artifact-key layering); `VERIFIABLE_KINDS` derives with an explicit exclusion that cannot silently widen; `DELTA` matches the schema enum; `METHOD_WORDS` has exactly one definition; dates are parsed in both modules.

Two skip pending #283, which is where `prose_provenance` and `apply_provenance` land — **this PR should merge after it**, and those tests activate then.

Full suite: 587 passed, 2 skipped. All gates green including `check_adoption --strict` and `check_instrument --strict`. `serialize` + `render` parse; no bot-owned file committed.

### DIFF
--- a/.github/workflows/regenerate.yml
+++ b/.github/workflows/regenerate.yml
@@ -56,7 +56,7 @@ jobs:
       - name: Fail on an unrouted retirement
         run: uv run python -m build.check_retirement
       - name: Render notebook
-        run: uv run python build/render.py
+        run: uv run python -m build.render
       - name: Sync README stat badges
         run: uv run python -m build.update_readme
       - name: Notebook structure check
--- a/.github/workflows/validate.yml
+++ b/.github/workflows/validate.yml
@@ -138,7 +138,7 @@ jobs:
       # serialized without ever rendering.
       - name: Render notebook and check it parses
         run: |
-          uv run python build/render.py
+          uv run python -m build.render
           uv run python -c "import ast, pathlib; ast.parse(pathlib.Path('notebooks/ai-stack-map.py').read_text()); print('generated notebook parses')"
           uv run marimo check notebooks/ai-stack-map.py
           git checkout -- notebooks/ai-stack-map.py
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -148,7 +148,7 @@ hyphenated kebab-case (`llama-3-1`).
 ```bash
 uv run python -m build.validate        # validate sources/ (must print "0 error(s)")
 uv run python -m build.serialize       # sources/ -> build/notebook_data.json
-uv run python build/render.py          # -> notebooks/ai-stack-map.py
+uv run python -m build.render          # -> notebooks/ai-stack-map.py
 uv run marimo export html notebooks/ai-stack-map.py -o /tmp/preview.html
 ```
 
--- a/README.md
+++ b/README.md
@@ -105,7 +105,7 @@ To preview the generated map locally (the output is for preview only — don't c
 
 ```bash
 uv run python -m build.serialize                                          # sources/ → build/notebook_data.json
-uv run python build/render.py                                             # → notebooks/ai-stack-map.py
+uv run python -m build.render                                             # → notebooks/ai-stack-map.py
 uv run marimo export html notebooks/ai-stack-map.py -o /tmp/preview.html
 ```
 
--- a/build/check_freshness.py
+++ b/build/check_freshness.py
@@ -73,10 +73,11 @@ from pathlib import Path
 
 import yaml
 
+from build.vocabulary import axes
 from build.check_rubric import FREE_TEXT, _clauses, components_of
 
 ROOT = Path(__file__).resolve().parents[1]
-AXES = ("openness", "adoption", "capability")
+AXES = axes()  # build/vocabulary.py owns this; the score schema declares it
 
 # `git log --raw` writes this in the old- or new-blob column for an add or a delete.
 NULL_SHA = "0" * 40
--- a/build/check_payload.py
+++ b/build/check_payload.py
@@ -28,8 +28,15 @@ def _is_date(value: object) -> bool:
     A `\d{4}-\d{2}-\d{2}` regex accepts 2026-99-99, which is exactly the class of thing a
     gate claiming to validate a date must not wave through.
     """
+    text = str(value)
+    # Both halves are needed. Shape alone was the original defect — it accepted 2026-99-99.
+    # Parsing alone is also too loose: `date.fromisoformat` takes the compact `20260815` from
+    # Python 3.11, so a second spelling could enter a corpus whose convention, and whose
+    # schema `format: date`, is hyphenated throughout.
+    if len(text) != 10 or text[4] != "-" or text[7] != "-":
+        return False
     try:
-        date.fromisoformat(str(value))
+        date.fromisoformat(text)
     except (TypeError, ValueError):
         return False
     return True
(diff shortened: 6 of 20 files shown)


ccerv1:
CI caught a defect in this sweep that my local run missed, and the miss is instructive.

Deriving `axes()` in `render.py` failed with `ModuleNotFoundError: No module named 'build'`. **`render.py` is the only build module invoked as a script** — `uv run python build/render.py`, in `validate.yml`, `regenerate.yml` and `docs/operations/publish-map.md` — which puts `build/` on `sys.path` instead of the repo root, so it cannot import from the package. Every other module uses `-m`, which is what I tested locally.

Switching the invocation to `-m` is the real fix and would make it consistent with all fourteen others, but it touches the publish runbook, so it is its own change rather than a rider on a bounded sweep.

The literal is back with the reason beside it, recorded as `intentionally_distinct`, and **the exemption is asserted rather than merely allowed**: `test_the_only_exemption_is_the_script_invoked_module` fails if `render.py` stops being script-invoked, so the exemption expires on its own instead of outliving its reason. That seemed better than an untested allowlist entry, which is how these accumulate.

Worth noting the shape: the sweep's own first attempt reached one module too far, and the thing that caught it was a gate rather than a reading. Suite 588 passed, 2 skipped.

ccerv1:
All three addressed. The first two were right that this sweep carried the defect it was written to find.

**1 — the exemption was unsound, not just inconvenient.** It asserted `render.py` was still script-invoked, never that its literal was still correct. A fourth axis would have left the rendered methodology silently omitting that axis's sources while every other test passed — the exact silent-denominator failure this PR is about, preserved by the PR.

Took the real fix: `render.py` moves to `-m` like the other fourteen build modules. Two workflows, three docs (`AGENTS.md`, `README.md`, `publish-map.md`). No exemption remains, and a test now asserts the script form does not come back.

**2 — the ratchets tested spelling, not meaning.** Each recognized exactly one form: a double-quoted tuple in one order, a bare `NAME =` assignment, `fromisoformat` appearing anywhere in a file. A single-quoted list, a reordering, an annotated assignment, or a module importing `fromisoformat` while shape-matching elsewhere would all have passed.

Now AST-based for the collection and assignment cases — matching on the *set* of strings, so container type, quote style and ordering all normalize away — and **behavioral** for dates: the validator is called with `2026-99-99` and `2026-02-30` rather than grepped for.

**That behavioral test immediately found a second date defect.** `date.fromisoformat` accepts the compact `20260815` from Python 3.11, so parsing alone admits a spelling the corpus does not use and the schema's `format: date` does not declare. `_is_date` now requires the hyphenated shape **and** a successful parse — shape alone was the original bug, parse alone is the opposite one. Neither half is sufficient, which I would not have found by reading.

**3 — routed and proposable are different questions.** Deriving `VERIFIABLE_KINDS` from the routing table made a newly declared `artifact_key` an accepted `--kind` with no pattern, no live check and no renderer behind it. The sets being equal today is what made the coincidence look like a definition.

Support is now defined by the three handlers a kind actually needs — `PATTERNS`, `_VERIFIABLE`, `_PUBLIC_URL` — and asserted to be an intentional subset of routed kinds, with `arxiv` the one named gap. `public_url` was refactored to read from `_PUBLIC_URL` so the registry is the single source rather than a dict literal inside a function.

**Merge condition understood.** Two tests still skip here because `prose_provenance` and `apply_provenance` are in #283. I will rebase this branch onto `main` after #283 merges and confirm a green run with **zero skips** before it goes in.

Also corrected the seven/ten module count in the `axes()` docstring and the inventory.

ccerv1:
Rebased onto main after #283. **Zero skips — and the ratchet caught something on the way.**

`prose_provenance.py` landed in #283 with its own copy of the axis triple, written before this sweep existed. Rebasing activated the two previously-skipping tests and the AST check failed immediately, naming `prose_provenance.py:62`. Fixed to derive.

That is precisely why the rebase-before-merge condition was the right call: the failure belonged on a branch rather than on main, which is where merge-then-activate would have put it. It is also a small piece of evidence that the ratchet works on code it was not written against — the module postdates the sweep by a day.

Confirmed after the rebase:

```
630 passed in 331.95s   (zero skipped)
```

All gates green including `check_adoption --strict` and `check_instrument --strict`; `serialize` and `render` both parse under `-m`; no literal axis triple remains anywhere in build/.

PR #285, opened 2026-08-16
Bind the reviewed clause to the claim it rewrites

ccerv1:
Provenance repair writes into hand-authored prose, so the failure that matters is not a crash — it is a plausible-looking sentence that says something nobody verified. This PR moves every judgment out of write time and binds what remains to the exact claim under repair.

## The clause boundary is proposed, never re-derived

The applier used to re-match the dated clause with its own regex as it wrote. That made every boundary bug a silent corruption:

```
in : Verified live 2026-08-13 on huggingface.co/datasets/allenai/ai2_arc (474 downloads).
out: Verified 2026-08-13 via the dataset card.co/datasets/allenai/ai2_arc (474 downloads).

in : Verified live 2026-08-13 via the U.S. AI Safety Institute report.
out: Verified 2026-08-13 via the document. AI Safety Institute report.
```

71 corpus clauses contain an internal dot, so the first was not an edge case; the second is worse, because the result reads as grammatical.

Tightening the regex would only have moved the failure to the next abbreviation. The defect was **where the heuristic was being used**. `BOUND` now belongs to `prose_provenance` alone, where it *proposes* a clause into the packet for a human to confirm, and `apply_provenance` holds no pattern at all — it substitutes the reviewed string literally, refusing unless it appears exactly once. A future boundary bug is now a visibly wrong string in review instead of a mangled file.

`BOUND` is also interpolated into `CLAUSE_ANY` rather than restated. The first cut wrote the expansion out as a literal in both patterns, so editing `BOUND` changed neither — the sibling-copy defect, inside the fix for sibling copies.

## A packet may decide a boundary, not which claim is rewritten

Occurring exactly once was the only thing bound about `current_clause`. A packet could nominate any other unique dated sentence: the rewrite would land there, the real verification line would survive untouched, and the old `CANONICAL.search` postcondition would have found the newly written one and called it a success.

`check` now also requires that the clause opens with the verification form and carries the packet's own `verification_date`; `rewrite` takes that validated date as an argument instead of scraping one back out of whatever string the packet supplied; and the postcondition is about the *result* — it must classify `canonical`, at the claimed date, with exactly one dated verification line remaining.

## All three unresolved states, one path

`packets()` emitted only the 172 `generic` records, so the 16 `ambiguous_noncanonical` and 13 `named_noncanonical` had no route through the pipeline at all. All 201 now go through the same review and the same `derive` decision, with the same justification required.

Deliberately one path, not two. A `named_noncanonical` record names something in prose, but that name has to turn out to be a document the record actually cites before it can be written as provenance — a separate "mechanical rewording" route would let the prose vouch for itself. Where the named document is not a date-aligned source in the score file, the honest decision is `reread`. Packets carry `current_state` and are refused if the live state has moved since review, and `--category` scopes them to one roster so review happens a batch at a time.

## `apply_one` had never been executed

It referenced an undefined `CANONICAL`, so every real application would have raised `NameError` after passing every check it was subjected to. `check` and `rewrite` were each tested alone and thoroughly; the function joining them was not tested at all. There is now an end-to-end test per state against a synthetic corpus, plus one asserting a mis-aimed clause refuses rather than leaving two dated lines.

`apply_one` also now normalizes `comments` before matching. The corpus wraps that field across lines, so a reviewed clause spanning a wrap could never have matched the raw text literally.

## Test denominators

`test_every_corpus_clause_round_trips_without_an_orphan` claimed the corpus and checked the first 60 files for one signature (`the document.\w`) — the URL case only, which the initialism case would have passed. It now walks every product the census calls dated, and its denominator is the census's own count rather than a `> 400` threshold.

Renamed to `test_every_dated_product_emits_a_literally_replaceable_clause`, which is what it can establish. Whether the heuristic picked the semantically right boundary is a reading, and review does that.

`test_the_applier_holds_no_clause_pattern` uses the AST rather than a text search: the module's docstring quotes the fragments it no longer uses, so a grep-shaped test would have failed on the prose and then been weakened until it matched nothing.

## Also

- Four hardware records reading `via substitute sources` were pulled out of #282's rewording — removing `live` would have promoted them into canonical form while they still named a method. Six benchmark records name a host rather than a document (`codecontests`, `gsm8k`, `humaneval`, `mbpp`, `ragas`, `swe-bench-verified`) and stay `named_noncanonical`.
- `docs/reference/sibling-invariants.md` records both constructs: the boundary, whose disposition is removal rather than deduplication, and the untested composition.

## Verification

649 tests, no skips. `validate`, `check_verification`, `check_payload` green. Census unchanged at 271 canonical / 201 unresolved — this PR writes no product prose.

### DIFF
--- a/build/apply_provenance.py
+++ b/build/apply_provenance.py
@@ -70,8 +70,14 @@ ROOT = Path(__file__).resolve().parents[1]
 # The whole dated clause, however it is currently worded, up to its terminator. Replaced
 # wholesale so `Verified live <date> via primary sources` and `verified live <date> via
 # primary-source research` collapse to the one canonical form.
+# A sentence-terminating "." is one followed by whitespace or end of string. A dot inside a
+# URL, a filename or a version number is not, and `[^.;]*` treated them alike: rewriting
+# `Verified live 2026-08-13 on huggingface.co/datasets/allenai/ai2_arc (474 downloads).`
+# stopped at the dot in `huggingface.co` and left `.co/datasets/allenai/ai2_arc (474
+# downloads).` orphaned after the new document name. 71 clauses in the corpus contain an
+# internal dot, so this was not an edge case.
 CLAUSE = re.compile(
-    r"[Vv]erified(?: live)?\s+(\d{4}-\d{2}-\d{2})\s*[,;]?\s*(?:via|on|against|using)?\s*[^.;]*",
+    r"[Vv]erified(?: live)?\s+(\d{4}-\d{2}-\d{2})\s*[,;]?\s*(?:via|on|against|using)?\s*(?:[^.;]|\.(?!\s|$))*",
 )
 
 
--- a/build/apply_provenance.py
+++ b/build/apply_provenance.py
@@ -63,7 +63,8 @@ from pathlib import Path
 import yaml
 
 from build.components import set_document_field
-from build.prose_provenance import AXES, HAS_WORD, METHOD_WORDS, classify, prose_digest
+from build.prose_provenance import (AXES, BOUND, HAS_WORD, METHOD_WORDS, classify,
+                                    prose_digest)
 
 ROOT = Path(__file__).resolve().parents[1]
 
@@ -77,7 +78,7 @@ ROOT = Path(__file__).resolve().parents[1]
 # downloads).` orphaned after the new document name. 71 clauses in the corpus contain an
 # internal dot, so this was not an edge case.
 CLAUSE = re.compile(
-    r"[Vv]erified(?: live)?\s+(\d{4}-\d{2}-\d{2})\s*[,;]?\s*(?:via|on|against|using)?\s*(?:[^.;]|\.(?!\s|$))*",
+    r"[Vv]erified(?: live)?\s+(\d{4}-\d{2}-\d{2})\s*[,;]?\s*(?:via|on|against|using)?\s*" + BOUND
 )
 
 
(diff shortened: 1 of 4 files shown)


ccerv1:
All three addressed, and the first finding is this fix containing the defect the sweep just gated against.

**1 — `BOUND` owned nothing.** It was declared once and then written out as an *expansion* inside both `CLAUSE_ANY` and `CLAUSE`, so changing `BOUND` changed neither pattern. I proved it to myself by editing `BOUND` and watching `clause_span` behave identically. Both now interpolate it, and `test_the_clause_boundary_has_one_owner` asserts the pattern text contains it — a sibling copy inside the fix for sibling copies, which is not a coincidence I can dismiss.

**2 — "period followed by whitespace" is not a sentence boundary.** Confirmed exactly as you describe:

```
in : Verified live 2026-08-13 via the U.S. AI Safety Institute report.
out: Verified 2026-08-13 via the document. AI Safety Institute report.
```

Worse than the URL case, because the result reads as grammatical.

The boundary now also declines to end on the dot of an initialism — but **that is still a heuristic, and the real problem was where it was being used.** A heuristic was deciding a boundary at *write* time, where every mistake is a silent corruption of prose.

So it now decides one at *packet* time, where a human sees the exact string that will be replaced. Packets carry `current_clause`; the applier does a **literal substitution** of that reviewed string and refuses if it is not present exactly once. No boundary is re-derived while writing. That also means a future boundary bug produces a visibly wrong string in review rather than a mangled product file.

**The corpus test was testing the old signature, not the claim.** It checked the first 60 files for `the document.\w` — the URL case only. The initialism fragment starts with a space and a capital and would have sailed through. It now round-trips **every** product carrying a clause, asserts that what follows the inserted name begins a new sentence or clause, and fails if fewer than 400 records were examined.

**Minor:** corrected — six host-only records, not five (`codecontests`, `gsm8k`, `humaneval`, `mbpp`, `ragas`, `swe-bench-verified`).

**Workflow agreed.** All 13 fold into the packet process. Target is 172 + 16 + 13 = **201 human-reviewed decisions**, no separate mutation path.

Suite: 637 passed, zero skips.

PR #288, opened 2026-08-16
Batch completeness for the provenance applier

ccerv1:
**Stacked on #286.** Machinery only — **this PR now changes no product prose.** The 35 rewrites it originally carried have been withdrawn; #286 explains why, and the manifest is now 42 re-reads.

## Batch completeness

`main()` exited 0 with rows undecided, and let a `reread` carry no reason. A category could report `12 derive, 4 reread, 3 undecided`, apply the 12, exit 0, and be called finished.

- A manifest is refused outright if any row has a missing or unknown decision, or a `reread` with no reason. Nothing is applied.
- The batch is all-or-nothing. `plan_one` runs every check **and the rewrite and its postconditions** without writing; only once every row planned does anything reach disk.

The preflight had to run the rewrite, not just `check` — a postcondition failure is a refusal `check` alone cannot see, so a `check`-only preflight would still have left packets 1–7 on disk when packet 8 failed. There is a test asserting exactly that, with the passing packet ordered first.

There is also a test that the shipped `ui_api` manifest is complete: every row decided, every re-read explained.

## A fixture pinned to a record the machinery repairs

`_packet()` ran against `anythingllm`. The first cut of the manifest repaired that record, its state stopped being `generic`, and every refusal test failed — not because the machinery broke, but because the fixture's subject had been fixed.

The manifest was withdrawn on other grounds, so `anythingllm` is `generic` again and these would pass either way. The coupling is the defect, not the failure it happened to cause, so the fixture is a synthetic corpus now.

## Verification

659 tests, no skips. `validate` clean. `0 derive, 42 reread, 0 undecided`, nothing written.

### DIFF
--- a/build/apply_provenance.py
+++ b/build/apply_provenance.py
@@ -226,7 +226,13 @@ def rewrite(comments: str, clause: str, document: str, iso: str) -> str:
     )
 
 
-def apply_one(packet: dict) -> bool:
+def plan_one(packet: dict) -> tuple[Path, str, str] | None:
+    """(path, original text, new comments) for one packet, or None if it is already canonical.
+
+    Every check and every postcondition runs here, and nothing is written. That is what lets a
+    batch be all-or-nothing: `check` alone does not exercise the rewrite, so a preflight built
+    on it would still have let a postcondition fail on packet 8 with 1-7 already on disk.
+    """
     check(packet)
     slug = packet["product"]
     claimed = str(packet["verification_date"])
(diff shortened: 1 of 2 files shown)


ccerv1:
Supersedes #287, which GitHub auto-closed when its base branch `carl/ui-api-provenance-manifest` was deleted on merge of #286. Same machinery, rebuilt on `main`; no product prose.

PR #308, opened 2026-08-16
Thin the entry points and drop the historical snapshots

ccerv1:
Stacked on #307, which is stacked on #306. Review in that order; the base retargets to `main` as each merges.

Four documents claimed things that are no longer true, and two claimed a completeness they cannot hold.

## `CLAUDE.md` → four pointers

It restated `AGENTS.md`'s routing table, the YAML-writing policy and the traps list, so a change had to land in two files to be true in one. It is now the repo map, the task router, the skill, and the warehouse boundary. The YAML policy and the traps moved into `AGENTS.md`, which is where a reader looking for either would go.

## `AGENTS.md` → stages, not an inventory

It claimed a complete list of "30 build modules" and had drifted to ~40 within a fortnight. A hand-maintained list of every module is a second copy of a fact the modules already carry — the failure this repo has now shipped five times. It becomes a map of the stages and the four checker families, with `--help` named as the authority. The sibling-invariant rule moves in beside it, because it belongs where somebody about to write a module will read it.

## `sibling-invariants.md` → a decision register

The four defects are fixed and each has a named regression test; git carries the narrative. What a sweep actually needs from this file is the other half — **which duplicates are deliberate, and why** — so `check_capability.DELTA`, `_GAP_OPEN`/`_OPEN` and `_DOWNLOAD_INSTRUMENTS` keep their entries and their reasons. Deleting the file outright would lose those, and they would be rediscovered as findings by the next sweep. Roughly 165 lines down to 70.

## `evidence-and-freshness.md` → the current state

It still described "1,408 confirmed and eight explicitly held" in four places. The 08-16 reconciliation settled those: **1,416 confirmed, empty queue.** The counts table is re-read live (1,827 distinct source URLs, median age 3d) and dated. The paragraph now says the queue being empty is a *state*, not a property — a hold is still the right answer when evidence contradicts a value, and the consistency gate still governs one.

## The warehouse cadence claims were wrong in both directions

`warehouse/sources.yaml` declared monthly and quarterly refreshes that were never implemented anywhere; every fetcher in it runs on one weekly cron in `refresh-data.yml`.

`warehouse/models/README.md` presented declared crons as a `Schedule` column — the exact claim the platform does not honor. Those crons are set at the model-revision layer, the platform schedules from the dataset, and no scheduled run has ever fired. The column is now `Declared cron`, under a header saying to read run history and check `triggerType` before believing any freshness claim that rests on it.

## Test plan

- 27 tests across docs-integrity, vocabulary-siblings and warehouse pass.
- No dangling internal link or backticked repo path anywhere in `*.md` (swept; the two remaining hits are a deliberate `foo.yaml` example and a correctly past-tense reference to a file removed in 2026-08).


ccerv1:
Superseded by #310 — GitHub auto-closed this when `carl/relocate-active-inputs` was deleted on merge, and a closed PR cannot be reopened or retargeted once its base is gone. Same branch and commits, now against `main`. Both review findings on this PR (the freshness census contradiction, the false parity cadence, the `--help` claim, and the two `Daily 7am` cells) are fixed in #310.

PR #309, opened 2026-08-16
Remove the warehouse catalog CSVs and fetchers with no consumer

ccerv1:
Stacked on #308 → #307 → #306. The audit behind it: https://claude.ai/code/artifact/9c6744bd-6a7e-4a37-9802-549b19cefc32

Every catalog CSV, fetcher and SQL model was checked against four kinds of consumer: the four tracked notebooks, all 20 published notebooks on the platform, the deployed platform models, and the sibling repos. **The decisive check: `currentai.catalog` has ten tables, and none of them corresponds to these five CSVs.** A CSV that loads into no static model cannot have a warehouse consumer, only a local reader — and these had none.

## Removed, each with its fetcher

| CSV | Fetcher |
|---|---|
| `catalog/ai-incidents/incidents.csv` | `fetch_incidents.py` |
| `catalog/github-orgs/orgs.csv` | `fetch_github_orgs.py` |
| `catalog/github-ai-repos/forked_ai_repos.csv` | `fetch_github_ai_repos.py` |
| `catalog/huggingface/tracked_datasets.csv` | (the dataset half of `fetch_huggingface.py`) |
| `catalog/huggingface/top_datasets.csv` | — |

**The incidents mirror is the GoodAI pattern exactly.** `currentai.aiid.*` is a live REST ingestion on a weekly cron, and the `ai-safety-incidents` notebook reads *that*, not the frozen copy. Its `sources.yaml` entry becomes an `ingested_by`, the same shape `goodailist` already carries.

## `fetch_huggingface.py` stays, trimmed

Its two model CSVs are uploaded nowhere, but they are the input list `fetch_model_benchmarks.py` scans to build `model_repos.csv` — which *is* uploaded, and which `entities_models.sql` reads. Only the dataset half was dead: four months of weekly author scans writing a file nothing opened. This is the one case where "no static model" did not mean "no consumer," which is why the audit checked readers as well as tables.

## Credentials — #281 folded in, and a break of mine fixed

Removing the GitHub fetchers took `GH_TOKEN` out of the job env while `Open refresh PR` still calls `gh pr create`. That is fixed: **`GH_FETCH_TOKEN` is restored and is now specifically the refresh-PR publication credential**, not a fetcher one. It stays a PAT rather than `${{ github.token }}`, because a PR opened with the Actions token does not trigger the validate workflow.

That made this the right place for **#281**, whose subject is the same step. The old guard checked secrets were *set*, never that any *worked* — every scheduled run from 2026-06-22 to 2026-08-10 died on `401 - Bad credentials` about 3.5 hours in, after the fetchers had done all their work and before the step that publishes it. `warehouse/catalog/` silently froze at 2026-07-28 while the schedule still read as live. The preflight now makes one authenticated GET per credential before anything expensive; measured against known-bad tokens, both 401, exit 1, ~1s.

`OSO_API_KEY` stays presence-only and required (the pyoso lookup in `fetch_huggingface` needs it); every probe for it is a warehouse round trip against a live model, which is a heavier precondition than a preflight should carry.

**This does not fix the expired token.** Rotating the PAT in the `GH_FETCH_TOKEN` repo secret is a maintainer action. Until then the workflow still fails — in seconds rather than hours, with an error naming the fix. Close #281 as superseded when this merges.

## Not removed — a decision, not an audit result

`entities_projects`, `scores_project_summary`, `scores_ossd_coverage`, `scores_dependency_graph`, `scores_fragility`. No consumer turned up for any, but they are one connected component (projects feeds only project_summary; dependency_graph feeds only fragility) and they are analysis rather than plumbing — fragility is the maintainer-concentration finding. Filed as a follow-up rather than resolved here.

## Test plan

- Full suite on the stack tip: **630 passed**, no skips.
- `validate` 0 errors; `warehouse/sources.yaml` parses, 5 sources; workflow YAML parses, 6 steps in order.
- Preflight probe executed against known-bad tokens: 401 / 401 / exit 1 / ~1s.
- `fetch_huggingface.py --help` runs; no reference to a removed path remains.


ccerv1:
Audit table behind this PR: https://claude.ai/code/artifact/9c6744bd-6a7e-4a37-9802-549b19cefc32

Issue #87, opened 2026-07-05
[methodology] Rethink the gap taxonomy: maturity is capability + adoption, not a sibling gap


ccerv1:
## Current state

The gap engine (`_stage_and_gaps` in `build/serialize.py`) gives each category a stage (0 to 5) and a set of gaps drawn from: `void`, `capability`, `adoption`, `maturity`, `openness`, `disclosure`.

Gaps are assigned like this today:

- Stage 5: no gaps.
- Stage 4 (at least one mature fully-open product, but fewer than the depth threshold): `maturity`.
- Stages 1 to 3 (no mature fully-open product): `maturity` plus one diagnostic. That diagnostic is `openness` if a mature option exists in any bucket (open-weights or closed) but none is fully open; otherwise `capability` (best fully-open capability below 4) or `adoption`.
- Stage 0: `void`.
- `disclosure`: declared per category, orthogonal to the rest.

Across the 15 categories, the gaps currently land like this:

| gap | categories |
|-----|-----------:|
| maturity | 12 |
| openness | 3 |
| adoption | 1 |
| disclosure | 1 |
| capability | 0 |
| void | 0 |

Two problems with this shape:

**1. `maturity` overlaps its own inputs.** A product's maturity score is a weighted blend of capability and adoption. Listing `maturity` as a sibling of `capability` and `adoption` reports the composite next to its parts. The stage number already encodes how mature and how deep a category is (4 is competitive but thin, 5 is redundant), so a standalone maturity gap mostly restates the stage. It also fires in 12 of 15 categories, so it does not discriminate between them.

**2. `capability` is effectively unreachable, which hides a real case.** The engine emits only one diagnostic per immature category. It picks `openness` whenever a mature non-open option exists, and only falls through to `capability` or `adoption` when nothing is mature anywhere. So `capability` can appear only when nothing is mature in any bucket and the best fully-open option is weak, which no category meets today. The clearest example of the mask is `edge_hardware`: its only fully-open board (BeagleY-AI) is capability 3, below the bar, but the category reports `openness` because mature open-ish and closed boards exist. The fact that the single fully-open option is underpowered never surfaces.

## Proposed resolution

Separate the two questions the stage and the gaps should answer:

- **Stage (0 to 5): how far along** the open ecosystem is, including depth and redundancy.
- **Gaps: what kind** of thing is missing.

Because maturity is capability blended with adoption, express a shortfall through its drivers instead of the umbrella:

[...]

PR #316, opened 2026-08-17
Split maturity into depth and its drivers, and name the score tiers

ccerv1:
Resolves the data side of #87 — steps 1 and 2 of the order of work in that issue. The front end (step 3) is a separate change and needs to ship alongside this one, because it hardcodes the current gap vocabulary.

## Gap engine

`maturity` was doing two unrelated jobs. A product's maturity is capability and adoption blended into one number; a category's is a stage from 0 to 5. Neither is a roll-up of the other, and the word implies age, which is not what either measures. So the split is by level, and each level gets a name that means something.

- **`maturity` is gone as a gap type.** It fired in 12 of 16 categories, so it distinguished between none of them, and it reported the composite next to its own parts.
- **`depth` is new, and fires at Stage 4 only** — a category with proven Frontier open options but not enough for redundancy. The shortfall there is genuinely count rather than quality. Defining depth as "no Frontier open product at all" would extend it over the weaker categories and rebuild the ubiquity this removes.
- **`capability` and `adoption` now both fire where both apply.** The old engine emitted one diagnostic per category and checked openness first, so `capability` was unreachable and had never once appeared.

`edge_hardware` is the case that hid. Its only fully-open board is genuinely underpowered, and the category reported an openness gap instead, so nobody reading the map could see it. It now reports `capability, openness`.

Measured against live sources:

| Gap | Before | After |
|---|---:|---:|
| maturity | 12 | — |
| depth | — | 6 |
| adoption | 3 | 5 |
| openness | 3 | 3 |
| capability | 0 | 1 |
| disclosure | 1 | 1 |

Counts run below the projections in #87 because the underlying scores have moved since that comment; three categories that were Stage 4 then are not now. The shape holds — every label names something different, and nothing fires in more than 6 of 16.

## Product level

`overall_score` and `tier` now ship **alongside** `maturity` and `mature`, same values. Both names publish for one release so the front end and the warehouse can move over before the old ones go away; a routine data sync must not be able to break the map halfway through.

- **Frontier**, 4.5+ — 78 products. Best in class on one axis, strong on the other.
- **Competitive**, 4.0 to 4.5 — 95 products. Strong on both, best in class on neither.
- Below 4.0, `tier` is null — 299 products.

Tiers derive from the score alone across every openness bucket, because they describe the product. `mature` keeps its fully-open gate, since only fully-open products advance a stage — so a closed product can be Frontier and not mature, which is documented as the intended reading rather than a contradiction. The tier legend ships in `descriptions.tiers` so no consumer hardcodes the two boundaries.

## Docs

The 4.5 bar is now written down, which it was not anywhere. Both axes are whole numbers from 1 to 5, so the bar can be cleared exactly one way: something has to score a perfect 5. A product graded 4 and 4 — strong on both counts — lands at exactly 4.0 and does not qualify, and that is the largest single group on the map. The working definition is therefore "best in class on at least one axis, and strong on the other."

`gap-analysis.md` also gains an explicit assignment table. `methodology.md`, `README.md` and three schema descriptions follow the same vocabulary. `CONTRIBUTING.md` carries no gap references despite #87 listing it.

## One interim rule, marked for deletion

If both measured axes clear their cutoffs and the blend still misses 4.5, the weaker measured axis stands in — an empty gap set at Stages 1-3 would read as "mature". Exactly one category is in that state: `benchmark_eval_data`, whose fully-open benchmarks are adoption 4 with a **null capability**, so the blend is adoption alone and tops out at 4.0. It reports `adoption`, as it does today.

The real fix is to score capability for evaluation sets. That axis already exists in the category, applied to 4 products out of 27 — LiveBench, GAIA and Humanity's Last Exam are banded on contamination resistance and breadth, while HumanEval and MMLU abstain on the grounds that "a dataset is not capable" and record their own saturation in prose. Filed separately; delete this branch when it lands.

## Not included

`build/notebook_data.json` is left to the daily regeneration bot. It was built locally to verify the numbers above, then restored.

## Test plan

- `uv run pytest` — 642 passed.
- New coverage: `maturity` absent from every gap set, `depth` at Stage 4 and nowhere else, both drivers firing together, the `edge_hardware` shape, the non-empty guarantee at Stages 1-3, `overall_score`/`maturity` dual publication including the null, and tier derivation and boundaries.
- Built the payload against live sources and confirmed the gap counts, tier counts, dual-publish equality and the two new legend blocks.

https://claude.ai/code/session_016VsMw63bXSy2vgdUMswhoC


ccerv1:
Superseded by #318 and closing in its favor.

#318 carries the same gap-engine change (split `maturity` into `depth` + its `capability`/`adoption` drivers; add per-product `overall_score`/`tier`) but rebased clean of the #313 platform-mirror commits — the intended 7-file diff rather than 14 — and with the review of this PR applied: tiers renamed to Leading/Strong with honest, weight-aware descriptions, the false `benchmark_eval_data` adoption gap removed, the adoption-gap definition decoupled from capability, and product-level "mature" retired from the canonical vocabulary (kept only as a legacy field). Please continue review there.

---
_Generated by [Claude Code](https://claude.ai/code)_

PR #320, opened 2026-08-17
Rewrite the stage and gap definitions readers actually see

ccerv1:
The payload's `descriptions` block is what renders in the site legend and the category drawer. It read like internal notes.

## What was wrong

- **Stage 4 and the depth gap defined each other.** Stage 4 was "at least one Leading fully-open product, but not yet enough for depth"; `depth` was "…fires at Stage 4 only". Neither said *four*.
- **The adoption gap defined itself by its own threshold** — "below the adoption threshold" names no number and no meaning, and it is one of the two most-fired gaps.
- **The disclosure gap shipped a repo path to the public payload**: "Declared per category (see `docs/reference/gap-analysis.md`)". That renders on the site.
- **`capability` overstated** — "not capable enough to be useful" describes tools people use daily.
- **Lengths ran from 7 words to 60**, and `depth` explained its own implementation.

## Stages

| # | Definition |
|---|---|
| 0 | The category is still nascent overall, with no leading products and no meaningful fully open options. |
| 1 | Fully open products are absent or remain far behind the category's strongest options. |
| 2 | Fully open products are becoming credible options, but remain meaningfully behind the category's strongest options. |
| 3 | Fully open options are genuinely viable, but none has reached the category's leading tier. |
| 4 | At least one fully open product has reached the category's leading tier, but the open field is still thin. |
| 5 | Multiple fully open products have reached the category's leading tier, creating a deep and competitive open ecosystem. |

## Gaps

| Gap | Definition |
|---|---|
| void | No usable fully open option exists at all. |
| capability | Fully open products exist, but they lack the capabilities needed to compete with the category's strongest options. |
| adoption | Fully open products have not reached the adoption of the category's strongest options. |
| depth | Fully open products have reached the category's leading tier, but there are too few to create a deep and resilient open ecosystem. |
| openness | Products have reached the category's leading tier, but none of them is fully open. |
| disclosure | Closed products rely on data or training recipes that are not disclosed, limiting visibility into how they are built and how they compare with fully open products. |

## Two constraints that shaped the wording

**`leading` and `strong` are tier names now**, so the copy uses them only where the tier is meant. A category can sit on a rung with no product in either tier — `safeguards` is Stage 3 with a best fully-open score of 3.5, below the strong floor — so a sentence like "Strong fully open options exist" would read as false to anyone who had just read the legend. The colloquial uses run on "the category's strongest options" instead.

**The adoption gap claims nothing about capability.** Since #318 both driver gaps fire independently, so a category weak on both axes shows both chips. The old "a capable fully-open option exists but is under-adopted" would have contradicted the capability chip sitting beside it.

## The verbatim contract, restored and enforced

`build/serialize.py` has always claimed its definitions were "kept verbatim from `docs/reference/gap-analysis.md`", but nothing checked it — which is exactly how the doc drifted to "below the adoption threshold" while the payload said something else.

There is now one definition per stage and gap, living in `serialize.py`:

- `gap-analysis.md`'s stage table gains a **Definition** column quoting `_STAGE_DESC` verbatim, plus a separate **Triggers when** column carrying the arithmetic (`L >= 4`, `1 <= L <= 3`, `L = 0` with score bands).
- `methodology.md` quotes the same gap sentences, appending mechanism *after* the definition rather than rewriting it into other words.
- `test_descriptions_match_the_reference_doc` asserts character equality across both docs.

Thresholds and assignment rules stay in the docs only; the payload carries definitions and never the arithmetic.

## Test plan

- `uv run pytest` — **650 passed** (649 plus the new guard).
- Verified the guard is not vacuous: reverting a single gap sentence in `gap-analysis.md` makes it fail with "gap-analysis.md is out of sync for `void`".
- No behavioral change — stage and gap *assignment* is untouched, so no category moves and the payload's `stage`/`gaps` values are identical.

## Follow-on

The site picks this up only after the payload syncs into `currentai-org/aipotluck.org`, where the front-end vocabulary work is open as #220.

Refs #87, #318.

https://claude.ai/code/session_016VsMw63bXSy2vgdUMswhoC


ccerv1:
All five addressed in `1db7a65`. Each verified against the code rather than taken on trust — all five hold.

**1. Copy overstated the logic.** Correct, and the counterexample is decisive: a category whose single product is fully open at 3.2 is Stage 2 *while being* the category's strongest option, so the sentence contradicted itself. Took your wording verbatim:

| | |
|---|---|
| Stage 1 | Fully open products are absent or remain at an early stage of adoption and capability. |
| Stage 2 | Fully open products are becoming credible, but remain limited in adoption or capability. |
| capability | The strongest fully open product remains limited in capability. |
| adoption | The strongest fully open product has not yet achieved broad adoption. |

Worth noting where this came from: the earlier round replaced a *different* false comparison ("the category's leading products", when a Stage 1–3 category may have none) and introduced this one in the fix. "Strongest options" reads as a comparison because it is one — it just isn't computed.

**2. Overlapping conditions and wrong stated order.** Confirmed: Stage 0 was a strict subset of Stage 1, and the table's claimed 5→0 top-down reading would assign every Stage 0 case to Stage 1. The code evaluates 5, 4, 0, 1, 2, 3. Stage 1 now carries the narrowing explicitly — `L = 0`, `B < 3.0`, and (`B >= 2.0` or `M`) — so the rows are mutually exclusive and readable in any order. `M` is now defined beside `L` and `B`, with the reason it separates Void from Open Experiments: a category whose open options are all feeble is Stage 1, not Void, when a mature closed or open-weights product exists.

**3. The guard graded less than it reported.** Both assertions now bind a definition to its owner, using the forms you suggested. I checked it against exactly the two swaps you named:

```
stage 1/2 swapped            -> FAILED (good)
capability/adoption swapped  -> FAILED (good)
clean tree                   -> PASSED (good)
```

**4. "Available measured axes" restored** to both tier descriptions. You're right that it was deliberate — a product with no capability grade is scored on adoption alone, so a tier is not a claim about both axes. I shortened the text to fit a drawer gloss that has since been removed, so there's no longer a length constraint to trade against.

**5. Changelog updated.** The Unreleased entry quoted "below the adoption threshold," the exact wording this PR removes; it now describes the rewrite and references #320.

`uv run pytest` — **650 passed**. Still no behavioral change: stage and gap assignment is untouched, so no category moves and the payload's `stage`/`gaps` values are identical to `main`.

ccerv1:
All three addressed in `e1f6a55`.

**Blocking: methodology's stage definitions.** Confirmed, and the diagnosis is the part worth keeping. This is a sibling omission of exactly the kind the guard was written to prevent — it bound gaps in *both* docs but stages in only *one*, so `methodology.md` carried its pre-#320 wording straight through a review that believed the guard covered it. The two you flagged were materially false: Stage 1's "exist but are weak on both axes" when the engine permits no fully open product at all (or one at adoption 5 / capability 2), and Stage 3's "strong" when `safeguards` sits there at 3.5, below the Strong floor.

All six bullets are now the verbatim `_STAGE_DESC` sentences, and the guard binds them by number and name. Verified it fires:

```
methodology stage 1 reverted -> AssertionError: methodology.md is out of sync for stage 1
```

**Stage 1 overclaim.** Agreed — the conjunctive reading is wrong when one axis carries more weight. Took your sentence: *"Fully open products are absent or remain substantially limited in adoption, capability, or both."*

**Cleanup, all three:**
- The comment and changelog claimed "stage, gap and tier definitions". This PR doesn't change `_TIER_DESC` and neither doc quotes it, so both now say stages and gaps, and the comment states plainly that the tier text is not covered by the test.
- "a mature closed or open-weights product" → "a category-leading closed or open-ish product", so the M explanation stops reintroducing product-level maturity vocabulary.
- The #320 entry now sits above #318, newest first.

**One thing beyond your list**, flagged since it touches copy you reviewed: the tier now reads **"category-leading"** in prose throughout, replacing both the bare "Leading" and the possessive "the category's leading tier". The score is a per-category weighted blend, so a bare "Leading" invited reading 4.5 as a single global bar. The payload's tier *key* is unchanged (`leading`), as is `descriptions.tiers`.

`uv run pytest` — **650 passed**. Still no behavioral change: assignment logic untouched, so `stage` and `gaps` values are identical to `main`.

PR #341, opened 2026-08-20
Archive the v1 stack_map, and record what is not superseded

ccerv1:
Applies the criterion **redundant AND replaced** — not merely unread — and checks both halves before touching anything.

## Archived

**`currentai.stack_map.*`** — the v1 hand-scored upload. 282 products against `registry.product_scores`'s 522, frozen at 2026-05-29, keyed on display name rather than slug, and read by **no deployed model** (I scanned all 41 models' SQL).

All nine tables were **snapshotted to CSV first** — 17,468 rows, including 282 product scores and **821 source citations** that are hand-curated work existing nowhere else. Frozen and unread is not the same as worthless. The dataset now reads `Stack Map [ARCHIVED 2026-08-20]` on the platform, with a description naming the replacement.

## Superseded, no reader

**`catalog.goodailist_repos`** → `signal_goodailist.repo_catalog`. 15,053 of its 15,396 repos are in the live table. The 343 that are not are alive and popular (top one has 168k stars, none archived) — they are entries **goodailist.com delisted from its own catalogue**, which is precisely the staleness the live model was built to fix.

Its only apparent deployed reader turned out to be a **docstring** in the model that replaced it. Left in place for now rather than renamed, since renaming a static model renames the table and nothing yet requires that.

## Not superseded — and this is the one worth writing down

**`catalog.pypi_downloads`** has **1,632,300 rows at day × package × country, covering 2025-05-01 → 2026-05-01**. Its apparent replacement `signal_pypi.package_downloads` has **106 rows at per-product monthly grain**.

A row-count comparison makes the older table look obsolete when it actually holds **a year of history nothing else has**, at a grain the newer table does not have. Two Live notebooks (`pypi-geo-trends`, `pypi-china-openclaw`) are the likely readers. **Keep.**

## Not yet

**`catalog.stack_map`** is live — `scores.stack_contributors` reads it — and stale at 199 products with nothing for `compilers` or `storage`. The registry join (`product_artifacts` ⋈ `product_scores`) yields 307 rows and covers **190 of its 205 repos**, so it can supersede it, but 15 repos are absent from the registry's github artifacts and need explaining first.

Regenerating it is a **fix, not a retirement**, and `warehouse/models/README.md` now says so at the point where its row count is quoted, so nobody reads those contributor counts as covering the full corpus.

## Kept

`catalog.osai_gap_map`, `taxonomy_crosswalk`, `osai_subcategory_mapping`, `country_populations` — external or reference data with no newer equivalent.

## Test plan

`tests/test_archived_datasets.py` holds the docs to the archive: every mention of an archived table must carry a word marking it historical (`archived`, `superseded`, `frozen`, `deprecated`, `retired`, `do not`). Adding a dataset to the list without updating the prose fails.

Proved it goes red before committing — appended an unmarked `Query currentai.stack_map.product_scores for per-product adoption.` to `queries.md`:

```
FAILED test_no_doc_presents_an_archived_table_as_current[currentai.stack_map.]
```

It also carries a guard on itself so an empty archive list cannot make it vacuous. 15 tests pass across it and `test_docs_integrity.py`.

## One thing I got wrong and corrected

I reported a contradiction between `queries.md` and `skills/pyoso-analyst/SKILL.md` over whether `catalog.goodailist_repos` was retired. There is none — both call it retired, and the "when querying it" in `queries.md` refers to `repo_catalog`, not the old table. I misread the antecedent.

### DIFF
--- a/docs/reference/where-scores-live.md
+++ b/docs/reference/where-scores-live.md
@@ -91,12 +91,29 @@ Four things carry columns named like ours and answer a different question.
 
 | Table | What it really is | State |
 |---|---|---|
-| `currentai.stack_map.product_scores` | The **v1 hand-scored upload**: `adoption_level`, `capability_score`, `capability_value`, `combined_score` per product | **Frozen.** 282 products, 11 categories, newest `last_verified` 2026-05-29, keyed on `product_name` rather than slug. **No deployed model reads it.** |
+| `currentai.stack_map.*` | The **v1 hand-scored upload**: `adoption_level`, `capability_score`, `capability_value`, `combined_score` per product, plus its own products, sources, categories, gap and featured tables | **ARCHIVED 2026-08-20.** Superseded by `registry.product_scores`: 282 products against 522, frozen at 2026-05-29, keyed on `product_name` rather than slug. Read by no deployed model. The dataset is renamed `Stack Map [ARCHIVED 2026-08-20]` and all nine tables were snapshotted to CSV first — the 282 product scores and 821 source citations are hand-curated work that exists nowhere else. |
 | `currentai.catalog.stack_map` | The repo→warehouse taxonomy bridge, carrying `adoption`, `capability`, `maturity` per product | **Live and stale.** 205 rows / 199 products / 14 categories, and 0 rows for `compilers` or `storage`. Read by `scores.stack_contributors`, so the open-stack developer counts are computed over a 199-product roster. Rebuilt by `warehouse/ingest/build_stack_map.py`, declared `refresh: on-curation-change` — which means a human, and #328 did not trigger one. |
 | `currentai.catalog.osai_gap_map` | The **external** OSAI gap map: `ease_of_adoption`, `maturity`, `overall_score` | A different organisation's taxonomy and scale. Not our products, not our axes. |
 | `currentai.ai_demand_curve.*` | `capability_score`, `adoption_level` against **OpenRouter/LMArena models** | Keyed to model names from a leaderboard, not to gap-map product slugs. |
 
-`currentai.stack_map.category_scores` and `.gap` are the same v1 freeze at category grain.
+
+
+## Superseded models, and the ones that only look superseded
+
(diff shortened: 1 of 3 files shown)


ccerv1:
Superseded by ADR-003, and no longer mergeable.

**The conclusion arrived by another route.** This PR argued case by case that
`currentai.stack_map.*` and `catalog.goodailist_repos` were redundant and replaced. ADR-003 then
settled the whole class: all four tables discussed here are in `warehouse/audits/externalization.json`
as `frozen-without-producer` — `catalog.stack_map`, `catalog.goodailist_repos`,
`scores.stack_contributors` and `signal_goodailist.repo_catalog`. The documentation change this PR
was making is already on main at `docs/reference/where-scores-live.md:292`, written in the ADR's
language rather than in this PR's ad-hoc "redundant AND replaced" criterion.

**It also cannot merge as it stands.** `git merge-tree` against current main reports a conflict in
both files it touches: `docs/reference/where-scores-live.md` changed on both sides, and
`warehouse/models/README.md` was deleted on main by the ADR-003 restructure. Rebasing would mean
re-deriving a conclusion the repo has already reached.

**One piece is worth keeping, and it is worth more now than when it was written.**
`tests/test_archived_datasets.py` asserts that an archived dataset must not be described anywhere in
the docs as though it were current — it scans `docs/`, `warehouse/`, `AGENTS.md`, `README.md` and
`skills/` for a mention of an archived table with no nearby word marking it as historical. Nothing
on main does this.

That is exactly the failure mode of the past two days. #509 was a workflow calling a producer #406
had deleted. #517 is six governed assets carrying a banner that contradicts their own manifest.
Both are the repo asserting something about a table that stopped being true.

The test's `ARCHIVED` list is hand-maintained and holds one entry. Driven from the externalization
receipt instead, it covers all 28 frozen tables and stays correct as the receipt grows. That is
being lifted into its own PR; the snapshot work and the platform archiving this PR describes
already happened and are unaffected by closing it.

Issue #263, opened 2026-08-14
[methodology] Ten arXiv /abs sources are cited for figures that only exist in the paper body


ccerv1:
Found by the `training_synthetic_datasets` verification pass and confirmed elsewhere in the 2026-08 sweep.

## The problem

Ten sources cite `arxiv.org/abs/<id>` as establishing a figure that **appears nowhere on the abstract page**. The number is usually real — it's in the paper — but the cited URL cannot support it, so the claim is unfalsifiable as recorded.

Concrete: `oasst1`'s capability `shows` claimed *"Guanaco/QLoRA fine-tuned on OASST1"*, cited to `arxiv.org/abs/2305.14314`. **OASST1 appears 0 times on the abstract page and 19 times in the PDF.** Repointing to `arxiv.org/pdf/2305.14314` fixed it.

`dolma-toolkit` was the same shape: cited an abstract page for a "Table 2" that page has never shown.

## Why it isn't just ten fixes

There's no stated convention. Some records cite `/abs`, some `/pdf`, some the HTML render (`arxiv.org/html/<id>vN`), and they are not interchangeable:

- **`/abs`** is stable and short, and carries only title, authors and abstract.
- **`/pdf`** carries everything, but a digest over a PDF is a digest over a binary — grepping it needs `pdftotext`, and the verification passes had to do exactly that.
- **`/html/<id>vN`** is version-pinned and carries tables as text, but doesn't exist for older papers and the version suffix goes stale.

## What to decide

1. Which URL form is the citation convention when the claim rests on a figure in the body?
2. Should `check_verification` or a lint warn when an `/abs` source is the sole establisher of a numeric claim? That's mechanically detectable — a `shows` containing digits, over an `/abs` URL.
3. Is a page-or-table locator worth adding to `shows` (e.g. "Table 2, p.7") so a re-reader knows where to look? Several notes already do this informally.

Related: #187 and #188 on evidence attribution generally.

PR #343, opened 2026-08-20
Rename the depth gap to resiliency

ccerv1:
Data half of the gap-label review Bob did on 18 August ([thread](https://karibalabs.slack.com/archives/C0BDH7LTQCV/p1787089850585339)). Tracked as CUR-565.

## What

Renames the `depth` gap to `resiliency`, in the payload key, both prose docs, and the tests. The description text is untouched — `Needs more fully open products at the leading tier to be resilient.` — so the only things that move are the key and the label.

Bob's point was that the description already said "to be resilient", which made the label the odd part out. Better than continuing to word-smith it, and it is Ayah's preferred word.

## ⚠️ Merge order

**Merge [aipotluck.org#263](https://github.com/currentai-org/aipotluck.org/pull/263) first.**

The gap key travels in `notebook_data.json`, and the front end looks up `GAP_META[gap].label`, which throws on a key it does not know. Landing this first breaks the gap icons until the front end catches up. #263 teaches it `resiliency` while keeping `depth` as a transitional alias, so with that in place either payload renders and there is no broken window in either direction.

## What this unblocks rather than does

`disclosure` carries the letter `i` in the front end for one reason, stated in its own code comment: to leave `d` free for the gap this renames. With `resiliency` on `r`, `d` frees up and disclosure can take its natural letter — which also removes the need for the I-word rewording that prompted the review. That is the third PR in this sequence, alongside dropping the alias.

## Test plan

- `uv run pytest -q` — 696 passed.
- `uv run python -m build.validate` — `0 error(s)`.
- Rebuilt the payload and confirmed the key moved cleanly: gap descriptions read `['void', 'capability', 'adoption', 'resiliency', 'openness', 'disclosure']`, and the per-category counts are unchanged apart from the rename (`resiliency: 7, adoption: 4, openness: 3, capability: 1, disclosure: 1`). Reverted the regenerated file so the bot handles it on merge.
- `test_descriptions_match_the_reference_doc` pins the payload text against both docs character-for-character, so the rename could not land in one place only.

## Checklist

- [x] `uv run python -m build.validate` prints `0 error(s)`
- [x] New/changed scores cite sources (`url`, `shows`, `accessed`) — n/a, no score changes
- [x] I did not edit `build/notebook_data.json` or `notebooks/ai-stack-map.py` (a bot regenerates them on merge)
- [x] Product appears in exactly one category roster and one org roster — n/a, no product changes


ccerv1:
**Blocked on purpose — converted to draft.**

The gate is not the front-end PR merging, it is the front end being *live*. [aipotluck.org#263](https://github.com/currentai-org/aipotluck.org/pull/263) targets `develop`, so merging it does not yet put `resiliency` in front of a visitor. Production keeps reading the payload until the next `develop` → `main` promotion.

So the order is:

1. #263 merges into `develop`
2. `develop` promotes to `main` and deploys
3. **then** this merges, and the data sync carries `resiliency`

If this lands before step 2, the deployed front end hits a `GAP_META` miss on a key it has never seen and the gap icons break. The transitional `depth` alias in #263 protects the other direction — a deployed front end reading older data — but nothing protects data that is newer than the deploy.

Marking ready once the promotion is out.

ccerv1:
Marked ready for review. The ordering still applies, so recording it here now that the draft state is no longer carrying it:

**Do not merge until [aipotluck.org#343](https://github.com/currentai-org/aipotluck.org/pull/343) has merged and deployed.** That promotion (`promote/2026-08-26` → `main`) is open and mergeable, and I confirmed it carries both halves of the front-end change — the `resiliency` vocabulary with `depth` as a transitional alias, and disclosure moved to the letter `d`.

Until that deploy is live, production still reads `depth` and has never seen the key `resiliency`. Merging this first means the next data sync sends a gap key the deployed front end cannot look up, and the gap icons break. The alias protects a deployed front end reading *older* data; nothing protects a deploy reading data newer than itself.

Sequence: promotion merges → deploy verified → merge this → data sync carries `resiliency`.

PR #364, opened 2026-08-25
A workflow for discovering candidates

ccerv1:
## Why

A colleague ran a weekly candidate sweep against this repo last week and it produced a prose report — 78 scored candidates, no artifact URLs, nothing the repo could ingest. Looking into why, the answer was not that she missed a skill. It was that there isn't one.

Every existing skill starts from a target you already have. `add-product` takes one product you have chosen. `promote-category` takes a seed roster that already exists in `sources/registry/`. `update-product`, `refresh-category`, `edit-category`, `migrate-axis`, `build-rubric` all operate on something already in the corpus. The step *before* all of them — deciding what should exist — had no workflow, so a sweep has nothing to follow and lands in a document instead of the tail registry.

## What this adds

`docs/workflows/discover-candidates.md` plus a thin skill over it, registered `primary`.

The output contract is the point: rows in `sources/registry/<category>.yaml` satisfying `docs/schemas/registry.schema.json`, which `promote-category` already knows how to turn into head products. That is the path compilers and storage took in #328.

## The four rules it names

Each of these broke a real sweep, which is why they are called out rather than left implicit:

1. **Read the category list from `sources/taxonomy.yaml` at run time.** That sweep produced zero candidates for `compilers`, `storage` and `ml_frameworks` — the first two added six days before it ran. A remembered list silently drops every candidate belonging to a category added since.
2. **Emit artifact URLs.** `slug`, `display_name`, `type`, `org` and `github` are required by the schema. A report of names and star counts cannot become a row without redoing every lookup.
3. **Do not invent a legitimacy rule mid-sweep.** That run adjudicated four candidates on a "WatchEvent coverage floor" and an "organic range" that appear nowhere in this repo, leaving two real products blocked against a standard that cannot be reproduced or appealed. Flag the row and park it; new rules belong in a rubric, proposed separately.
4. **A 429 is not a measured absence.** Five products came back "cannot determine" from Hugging Face rate limiting; three were verifiable from the warehouse minutes later. `build/fetch_source.py` already encodes 429/403/5xx as transient, with a comment recording this exact past failure. A sweep should not relitigate it.

## What it also fixes by stating it

Candidate rows carry identity and artifacts only — scoring happens later, against evidence. The sweep ends at a populated registry and a summary for review, not a PR per candidate. And parked candidates keep their reason, so a reject does not come back next week to be triaged from scratch.

## Test plan

- `uv run pytest -q` — 889 passed
- `uv run python -m build.validate` — `0 error(s)`
- `tests/test_docs_integrity.py` pins the primary skill set one-to-one against workflows, the README router and the on-disk `skills/` directory, so `PRIMARY_SKILLS` gains the new entry alongside `skills/registry.yaml`, `docs/README.md` and the `AGENTS.md` roster. All 13 doc-integrity tests pass.

## Checklist

- [x] `uv run python -m build.validate` prints `0 error(s)`
- [x] New/changed scores cite sources (`url`, `shows`, `accessed`) — n/a, no score changes
- [x] I did not edit `build/notebook_data.json` or `notebooks/ai-stack-map.py`
- [x] Product appears in exactly one category roster and one org roster — n/a, no product changes

### DIFF
--- a/.claude/skills/discover-candidates
+++ b/.claude/skills/discover-candidates
@@ -1,0 +1,1 @@
+../../skills/discover-candidates
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -303,6 +303,7 @@ Claude Code session discovers them by name; if yours does not list them, read
 
 | Skill | When to use | Workflow |
 |-------|------------|----------|
+| `discover-candidates` | Sweep for products the map does not have yet | `docs/workflows/discover-candidates.md` |
 | `add-product` | Add a new product | `docs/workflows/add-product.md` |
 | `update-product` | Change an existing product (identity, prose, a score, rosters, retirement) | `docs/workflows/update-product.md` |
 | `edit-category` | Create a category, or change its definition/weights/roster | `docs/workflows/edit-category.md` |
--- a/build/validate.py
+++ b/build/validate.py
@@ -148,6 +148,18 @@ def validate_sources(data: dict) -> list[str]:
     tail_slugs: dict[str, str] = {}
     tail_repos: dict[str, str] = {}
     head_repos: dict[str, str] = {}
+    # A retired slug is a settled decision, so it disqualifies a candidate exactly as a live
+    # slug does: re-adding it re-opens the consolidation that retired it. Checking only
+    # `slug in prods` missed that entirely -- amazon-nova-pro, an alias of amazon-nova, passed
+    # validation as a new candidate. Resolved here rather than reusing the `claimed` map built
+    # in the slug-stability section below, because that map is built by a loop whose purpose is
+    # to REPORT duplicate claims as it goes; this one only needs to resolve them.
+    alias_of = {
+        alias: slug
+        for slug, product in sorted(prods.items())
+        for alias in product.get("aliases") or []
+        if isinstance(alias, str)
+    }
     for slug, product in prods.items():
         for artifact in product.get("github") or []:
             if isinstance(artifact, dict) and isinstance(artifact.get("url"), str):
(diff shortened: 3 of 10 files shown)


ccerv1:
Rebased onto current `main` and addressed all eight review findings in currentai-org/os-ai-map#374 (this branch was old; #374 carries the original two commits plus a follow-up commit with the corrections, so its diff is against clean current-main). Summary of the fixes:

1. Existing `sources/registry/*.yaml` rows now load in step 1 and dedup as the third set, before the warehouse pool.
2. Warehouse pool reframed as the long-tail discovery set — consolidate / recover canonical identity / enrich, never disqualify on presence alone.
3. Reconciliation given two explicit invariants, and provenance (source URL + fetch date) now required for parked candidates too.
4. Creating a new registry file (`category:` / `products:`) documented, since only `compilers` and `storage` exist today.
5. `category_statuses(taxonomy)` shown with its real call shape.
6. Escalation distinguishes a disclosed, predeclared retrieval cutoff from a mid-sweep legitimacy rule.
7. Handwritten "Six primary editor skills" count dropped from `AGENTS.md`.
8. `sources/taxonomy.yaml` removed from "Files this changes".

`build.validate` → `0 error(s)`; full suite → 1010 passed. Suggest continuing review on #374 and closing this one once you're happy with it.

---
_Generated by [Claude Code](https://claude.ai/code)_

ccerv1:
#374 is fully green (`validate`, `check`, `generated-files-guard` all passed; `publish` skipped). Closing this as superseded so #374 is the single canonical review surface.

---
_Generated by [Claude Code](https://claude.ai/code)_

PR #418, opened 2026-08-30
Corpus expansion batch 1: 134 products

ccerv1:
134 products into nine categories, taking the corpus from 527 to 661. Every record is generated from primary sources, and the three claims it makes are of very different strengths — that is stated per axis rather than averaged away.

**Depends on #417.** The Details payload exceeds marimo's 8 MB cap at this corpus size, and marimo drops the cell silently. Merge that first.

## The funnel

| Stage | Count | What removed them |
|---|---|---|
| Pool, top 700 by the Sovereign Tech Agency ranking | 700 | |
| Already declared | −141 | repo-level dedup, correct since #415 |
| Identity resolution | −6 | forks, mirrors, archived, name collisions, one package-backed repo |
| No decisive category | −267 | crosswalk purity < 80% or support < 5 |
| No readable licence decision | −67 | licence unresolved, no body, a second licence file in root |
| Unusable identity or prose | −85 | slug collision, org-suffix slug, marketing or changelog description, a cited URL not at 200 |
| **Generated** | **134** | |

## Category assignment is derived, not guessed — and it is the weakest claim here

252 already-declared repositories carry both a pool subcategory and a map category, which yields an empirical crosswalk with support and purity per pair. Only pairs at **≥80% purity and ≥5 supporting products** were used:

| Category | Added | Crosswalk source (purity/support) |
|---|---|---|
| `orchestration_agents` | 50 | Coding 92%/13; Prompt & Scaffold 100%/5 |
| `agent_tools_protocols` | 28 | Give agent tools 100%/12 |
| `ui_api` | 17 | Personal 80%/15 |
| `deployment` | 16 | Orchestration 100%/7 |
| `telemetry_observability` | 7 | Observability 93%/15 |
| `finetuning_code` | 7 | Fine-tuning 92%/12 |
| `ml_frameworks` | 4 | Deep Learning Frameworks 86%/7 |
| `storage` | 3 | Data systems 86%/14 |
| `compilers` | 2 | Kernels 90%/10 |

**At 80–92% purity, roughly 10 to 20 of these 134 are in the wrong category.** Nothing breaks mechanically — the shared software ladder reproduces either way — but a misfiled product distorts its category's stage and the gap arithmetic. Every product records its own subcategory, purity and support in `comments`, so the weak assignments are findable rather than buried. **This is the part to spot-check.**

## Openness — the strong claim

Read from the LICENSE **body**, never from the GitHub API's label. 65 MIT, 58 Apache-2.0, 8 AGPL-3.0, 3 GPL-3.0. Three score 4/open_core on an enterprise path in the repository root; the rest 5/open_source.

The detector for appended conditions was built against today's known cases and validated before use, which mattered three times over:

- Its first version parked **110 clean repositories** on `you may not use this` and `additional term` — both of which appear in *stock* Apache-2.0, the former in the APPENDIX boilerplate and the latter in section 4.
- Tightened, it then caught **nothing**, so it was run against Crawl4AI and MinerU. It caught Crawl4AI's appended attribution requirement and **missed MinerU**, whose `LICENSE.md` states its thresholds *before* referring to Apache and so carries no `END OF TERMS AND CONDITIONS` marker to scan past. Fixed, it catches both and stays clean on vLLM, LangChain and Ray.
- Marker's shape — Apache code, weights bounded in a separate `MODEL_LICENSE` — is invisible to a pipeline that reads only `LICENSE*`. Any repository carrying a second licence file in its root is now parked for a human read rather than scored on whichever file happens to be permissive. That is 17 of the 67.

`core-gated` is a repository-and-README read, not a pricing-page read, so every record carries `confidence: medium` and says so in its note.

## Adoption — the weak claim, labelled as such

`stars_fallback` on every record: 76 at level 2, 58 at level 3, `confidence: low`, capped at 3 by the scale. No package artifact is declared for these yet, so stars are the only instrument available, and each note says to read the band as a floor rather than a measurement.

## Capability — deliberately unscored

`score: null`, `basis: n/a` on all 134. Banding in these categories is a feature matrix against a named anchor, and no such comparison was made. Each note says that plainly and asks for a category-wide pass where every product is banded against one anchor, rather than a number invented per record. This is the "score each on the three axes, or record the axis as unscored rather than guessing" that CUR-561 asks for.

## Prose

Descriptions are sourced from each repository's own GitHub description and README, never written from outside knowledge. Three of the corpus's own gates rejected earlier drafts and shaped the result: `test_descriptions_carry_no_unverifiable_marketing_claims` (the generator now reuses the corpus's own `MARKETING` pattern rather than a second definition that disagreed with it), a description that had swallowed a changelog, and a `display_name` of `miles` where the README's H1 gives `Miles`.

## Bookkeeping

`long_tail` counts move semantically, not just enough to satisfy the gate: all 134 were pool members, so `scored` 527→661 and `overlap` rises by the same, with `scored_outside` and `universe` unchanged. 116 new organization records, typed `unknown` — the publishing entity is known from the repository owner, but its kind is not, and guessing `company` for 116 orgs would be inventing metadata.

Corpus goldens regenerated from the code the tests use. `MEASUREMENT_COUNT` stays 377: the measurement baseline is a frozen observation snapshot, and new declarations cannot retroactively create observations.

## What this is not

Not the 470. The pipeline resolved 553 candidates and this is what survived a decisive category *and* a readable licence *and* usable prose. The 267 parked for category are the largest remaining pool and want either more crosswalk evidence or hand assignment; the funnel above says exactly where each one went.


ccerv1:
Superseded by #419 (resolution ledger) and its successor expansion PR. This branch recreated twelve repositories #413 had already resolved, and a hand review found 38 more category errors; both rounds are now recorded in the ledger rather than fixed in place.

PR #422, opened 2026-08-31
Corpus expansion batch 1: 77 products

ccerv1:
77 products into nine categories, taking the corpus from 527 to 604. Every number below is generated from the batch itself rather than carried forward, because the previous revision of this body described a 134-product batch that no longer existed.

**Depends on #419** (the resolution ledger), which depends on **#417** (the Details payload trim, without which the notebook's modal silently breaks at this corpus size).

## Three rounds of review, 134 → 115 → 89 → 77

| Cut | Removed | Why |
|---|---|---|
| Ledger, seeded from #413 | 12 | repositories a person had already resolved — one of them a product the corpus carries as `agent2agent-protocol` |
| Hand review of all 115 | 38 | category assignments the crosswalk got wrong — about a third of what survived |
| `unresolved` holds a promotion | 12 | see below; this revision's fix |

Each cut is recorded in `sources/resolution_ledger.yaml` with a verdict and reasoning, not applied here and forgotten. The next run inherits all of them.

## `unresolved` is a hold, not permission

The previous revision had a logic bug worth naming. The #418 review recorded `ag-ui` as a *protocol* filed under agent frameworks, `serena` as an MCP retrieval toolkit and `repomix` as document ingestion — and the same run then published all three in `orchestration_agents`, the category those entries call wrong.

The cause was reading "not permanently excluded" as "this proposal may publish". They are different questions. `build/resolution.py` now separates them: `blocks_new_product` is the permanent exclusion `validate` enforces, and `holds_bulk_promotion` returns **any** entry, `unresolved` included, and is what a bulk run consults. A person may still add such a product by hand once they have settled the question; a bulk run may not.

## The batch

| Category | Added | Crosswalk source (purity/support) |
|---|---|---|
| `orchestration_agents` | 33 | Coding 92%/13; Prompt & Scaffold 100%/5 |
| `deployment` | 13 | Orchestration 100%/7 |
| `finetuning_code` | 7 | Fine-tuning 92%/12 |
| `ui_api` | 7 | Personal 80%/15 |
| `agent_tools_protocols` | 6 | Give agent tools 100%/12 |
| `telemetry_observability` | 4 | Observability 93%/15 |
| `ml_frameworks` | 3 | Deep Learning Frameworks 86%/7 |
| `compilers` | 2 | Kernels 90%/10 |
| `storage` | 2 | Data systems 86%/14 |

**Openness**, read from the LICENSE body and never from the API label: 38 Apache-2.0, 31 MIT, 5 AGPL-3.0, 3 GPL-3.0. Three score 4/open_core on an enterprise path in the repository root; the rest 5/open_source. `confidence: medium` throughout, because `core-gated` is a repository-and-README read rather than a pricing-page read.

**Adoption** is `stars_fallback` on every record — 50 at level 2, 27 at level 3, `confidence: low`, capped at 3 by the scale. Each note says to read the band as a floor.

**Capability** is `null` with `basis: n/a` on all 77. Banding in these categories is a feature matrix against a named anchor, and no such comparison was made. That is deliberate, and it is the strongest argument for scoring **category by category** rather than in batches: a feature matrix needs every product in a category banded against one anchor.

`long_tail` moves semantically — all 77 were pool members, so `scored` and `overlap` rise together and `scored_outside` and `universe` hold. Corpus goldens regenerated from the code the tests use; `MEASUREMENT_COUNT` stays 377, since the baseline is a frozen observation snapshot.

## This batch is skewed away from the Sovereign Tech Agency priority

Worth stating, because it bears on whether to merge this at all. The STA candidate document ranked **Band 1, lower-stack infrastructure**, as "score first", explicitly feeding `compilers`, `storage` and the telemetry gap. This batch delivers `compilers` +2, `storage` +2, `telemetry_observability` +4 — 8 of 77 — while `orchestration_agents` takes 33, the band that document ranked last.

The cause is mechanical: **the crosswalk is weakest exactly where the STA priority is strongest.** The agent and app subcategories are the pure ones (`Give agent tools` 100%, `Coding` 92%), while the lower-stack ones are ambiguous (`Deploy & Serve` 60%, `Inference service optimization` 50%, `Gateway` no evidence at all). A ≥80% purity rule filters out precisely the band the funder asked for. 67 Band-1 candidates died at category assignment, and `inference_code` gained nothing.

Survival was even across bands (20% / 8% / 22%), so nothing else in the funnel is biased. It is the crosswalk gate alone.

## Residual risk

These 77 have had an agent's review, and that review moved the batch by a third. The categories it found no errors in were the small ones — `compilers` (2), `ml_frameworks` (3), `finetuning_code` (7). `orchestration_agents` carries 33 records and is where the judgment is thinnest, because the line between a coding agent and a coding-agent *tool* is the one most often got wrong.

Given that and the STA skew, a per-category pass — hand-assigning the ambiguous lower-stack crosswalk pairs and banding capability against a category anchor in the same run — would produce a better corpus than merging this. This PR is ready on its own terms; whether it is the right next step is a separate question.


ccerv1:
**Held, deliberately — not a correctness concern.**

This batch optimizes for where the crosswalk classifier is most confident, which is a selection bias against the layer the Sovereign Tech Agency asked for first. The agent and app pool subcategories are the pure ones (`Give agent tools` 100%, `Coding` 92%); the lower-stack ones are ambiguous (`Deploy & Serve` 60%, `Inference service optimization` 50%, `Gateway` no crosswalk evidence at all). So a purity threshold systematically underweights `inference_code`, `deployment`, `compilers`, `storage` and `telemetry_observability` — 67 Band-1 candidates died at category assignment and `inference_code` gained nothing.

Correcting that before this composition becomes the new baseline. The next unit of work is an STA-first per-category pass, starting with `inference_code`, using those 67 rejected candidates as the queue rather than resweeping discovery. Capability will be banded across **existing and new** products against one anchor in the same pass, since banding only the new records would distort relative ordering against stale ones.

This PR stays open as a reviewed inventory of lower-priority additions, to revisit once Band 1 is properly represented — at which point merging it is broadening after priority coverage rather than letting the easy-to-classify part of the pool set the shape of the map.

ccerv1:
Closing in favor of per-category promotion PRs.

The work is not discarded: all 77 product and score files remain on `carl/expand-corpus-batch-1-v4`, which is not being deleted, and the per-category PRs cherry-pick from it.

Why re-cut rather than review this as-is:

- 154 files in one diff means any single objection blocks all 77. The units that have actually converged were #414 (5 products) and #424 (12, one category), both of which drew substantive findings and landed.
- The batch is one-third scored, not finished. All 77 carry openness with components, but **0/77 carry adoption and 0/77 carry capability** — 154 empty axes. Reviewing it as a finished batch would have been reviewing a premise that isn't true.
- Every product here declares `github` and nothing else, so `select_route` sends all 77 to `stars_fallback` at cap 3. That needs a per-product instrument decision, not a bulk pass.
- 33 of the 77 are `orchestration_agents`, blocked upstream on the rung-3 rewrite.

Triage also has to precede scoring. First pass over the `finetuning_code` slice already found that `tinker-cookbook` is companion material for `tinker`, which is already a head product on that category's roster — so it was never a new product, and scoring it first would have wasted the expensive axis.

PR #432, opened 2026-08-31
ml_frameworks: paddle, mlx and pytorch-lightning

ccerv1:
Three products into `ml_frameworks`, worked through `docs/workflows/add-product.md`. Second
per-category tranche after #422 was re-cut; #431 is the first.

## Stage impact

| category | stage | L | B | products | gaps |
|---|---|---|---|---|---|
| `ml_frameworks` | 5 → 5 | **14 → 15** | 5.00 → 5.00 | 22 → 25 | none → none |

**No stage transition**, but `L` moved and the report names who moved it: `pytorch-lightning`
reaches maturity 4.6 under this category's 0.6/0.4 weights (adoption 5, capability 4), clearing the
4.5 bar. The category was already Stage 5 with `L` 14, so an extra leading open product cannot
change the rung — but this is the first tranche where the gate has something to report rather than
a flat line, which is the case it was built for.

`paddle` lands at 4.4 and `mlx` at 4.0, both below the bar.

## What landed

| slug | openness | adoption | capability |
|---|---|---|---|
| `paddle` | 5 open_source (Apache-2.0) | 4 usage_volume — 1,785,666/30d | 5 |
| `mlx` | 5 open_source (MIT) | 4 usage_volume — 1,464,066/30d | 4 |
| `pytorch-lightning` | 5 open_source (Apache-2.0) | 5 usage_volume — 12,410,527/30d | 4 |

All three adoption bands sit on the authoritative PyPI route, and every package was verified to
point back at its own repository before being declared. No stars-fallback and no hand-authored
instrument anywhere in this tranche.

Openness was derived by replaying the category recipe through `check_rubric.score_openness`, not
hand-assigned.

## Capability placement

`paddle` 5: a framework rather than a library on one — its own graph engine, distributed execution
and deployment path, with automatic parallelism derived from single-device tensor partitioning
annotations, and training and inference served by the same framework. That is the `pytorch` /
`tensorflow` shape, both 5 here.

`mlx` 4: foundational in design — own array type, composable transforms, autodiff, unified CPU/GPU
memory, four language APIs — but Apple silicon only, where the 5s are hardware-portable. `openvino`
is the precedent in this category: technically strong, scoped to one vendor's hardware, 4.

`pytorch-lightning` 4: structures and distributes a PyTorch training loop and depends on PyTorch
for the tensor engine and autograd. Same shape as `accelerate` at 4.

## Two calls worth reviewing

**Paddle's LICENSE reads as non-standard and isn't.** A `Copyright (c) 2016 PaddlePaddle Authors`
line is prepended above the Apache text, which is exactly the shape `add-product.md` warns makes a
genuine Apache repo report `NOASSERTION`. I confirmed the remainder is byte-identical to the
canonical template through `END OF TERMS AND CONDITIONS`, with only the standard 1,111-character
appendix after it. The score note records this so the next reader does not re-litigate it.

**`pytorch-lightning` declares one package, not two.** The same repository also publishes an
umbrella `lightning` package at 5,665,410 downloads/30d. I did not declare it: it redistributes this
code alongside Lightning Fabric, so claiming it here would take an artifact a separate product may
later need. The band is 5 either way, so the choice does not move the score — but it is a
declaration decision rather than a measurement one, and worth a second opinion.

## The freshness constraint, now measured

None of the three records a `relative_to`, and the reason is structural rather than a choice.
`check_capability` requires that a comparison not claim a fresher verification date than its root,
and **the entire corpus is stale against today**: 399 of 539 capability records were last confirmed
2026-08-13, and the only records dated 2026-08-31 are ones written in this session.

So a new product can build comparison structure *within* its own tranche — `onetrainer at
ai-toolkit` in #431 does — but it cannot link into an established anchor without refreshing that
anchor first. Every candidate root here (`pytorch`, `tensorflow`, `accelerate`, `openvino`, `jax`)
is 2026-08-13.

The comparison reasoning is therefore in each note, naming the peer and the band logic, with the
machine claim omitted. That keeps the argument reviewable without asserting a re-derivation I did
not do.

One wrinkle worth knowing: the notes cannot state the dates. `tests/test_score_notes.py` forbids a
note from stating a date unless the date is a fact about the product — chronology is git's job. My
first draft explained the omission by citing the roots' confirmation date and the guard caught all
three. So the notes now make the argument in undated form, and the dates live in `last_verified`
and the history. Coherent, but it means **the freshness cost cannot be recorded in the corpus
itself** — it has to be a gate or a report, which is a further argument for the stage-impact and
anchor-refresh tooling being real code rather than a convention.

This is a roadmap input, not just a note on this PR: **structured comparisons have a freshness
cost.** Either each category pass opens with an anchor refresh, or new products score standalone
and the `relative_to` graph stops growing.

## Files

Products, scores, the `ml_frameworks` roster, three existing org rosters appended
(`paddlepaddle`, `apple`, `lightning-ai` — no new org files needed), `long_tail.json` re-synced
539 → 542, and the corpus goldens. Every golden delta is the expected size: `axis_results` 542 =
product count, axis rows +9 = three axes × three products, reconciliation +3, parity +3.


ccerv1:
Rebased onto main after #431 landed. New head `40096d2`.

Six files conflicted, all derived values: `long_tail.json` and the five golden pins. Resolved by taking main's side and **recomputing from the combined corpus** rather than picking either version — the product, score, roster and org files merged cleanly with no conflict.

Every recomputed delta is the expected size for eight products across both PRs:

| golden | before | after | delta |
|---|---:|---:|---|
| `axis_results` / reconciliation | 539 | 547 | +8 products |
| axis rows | 1617 | 1641 | +24 = 3 axes × 8 |
| `check_parity` computed | 534 | 542 | +8 |
| `long_tail.counts.scored` | 539 | 547 | +8 |

Stage impact re-run against the new main is unchanged: `ml_frameworks` stage 5 → 5, **L 14 → 15** (`pytorch-lightning`), B 5.00, 22 → 25 products, no transition.

Full preflight green on this head: all 15 locally-runnable CI steps.

Issue #426, opened 2026-08-31
[tooling] check_adoption should flag a stars_fallback record with a measurable package


ccerv1:
`check_adoption` validates that a recorded band matches its recorded reach. Nothing checks whether a `stars_fallback` record is leaving a **measured figure on the table**.

That gap is systematic, not incidental. **93 records currently band on stars**, and two spot checks found the same error class in both:

- `executorch` at level 2 on 4,969 stars, while Meta's own engineering blog states it powers on-device inference across Instagram, WhatsApp, Messenger, Quest and Ray-Ban for billions of people.
- `tensorflow-serving` at level 2 on 6,360 stars, against 4,627,937 monthly downloads of `tensorflow-serving-api`.

A hand sweep then found five more (#425): `firecrawl`, `typesense`, `lakefs`, `tensorlake-sandbox`, `otari`. All pre-existing. The pattern is one-directional — stars understate anything that ships as a package, an image or an embedded runtime rather than as a starred repository — which is the mirror of the 2026-08-13 correction where stars *inflated* download-vocabulary bands.

## What the check should do

For every record whose `signal_type` is `stars_fallback`, read the project's own install instructions from its README and report any package or image it names for which a figure is retrievable. Report, do not re-band: the judgment of whether a package is the product's primary channel is not mechanical.

## What it must not do, with worked examples

Three failure modes, all observed while building the throwaway version:

**1. A package named in an install line is not necessarily the product.** `pip install bigcodebench packaging` yielded `packaging` at **2,191,710,133** monthly downloads. `simple-evals` yielded `openai` at **417,881,229**. Both would have been catastrophic misattributions.

**2. Owner-name matching is too loose.** `openai/simple-evals` matched the `openai` package because the string "openai" appears in both the org and the package metadata. Ownership needs the full `owner/repo` path, or a homepage or author that names the project — and the last two cannot be checked mechanically with confidence. `tensorflow-serving-api` establishes ownership only through `author: Google Inc.` and `home_page: tensorflow.org/serving`; a repository-link check rejects it.

**3. A rate limit is not a finding.** The first sweep drew HTTP 429 from pypistats for four of five packages, and the banding step turned that into "0 downloads → level 1". A failed fetch must refuse to band, never record an absence as a measurement. `build/fetch_source.py` already treats 429/403/5xx as transient for this reason.

## Also worth covering

[...]

PR #433, opened 2026-08-31
ui_api: six products, and why five of them band on stars

ccerv1:
Six products into `ui_api`. Third per-category tranche after #422 was re-cut (#431
`finetuning_code`, #432 `ml_frameworks`).

## Stage impact

| category | stage | L | B | products | gaps |
|---|---|---|---|---|---|
| `ui_api` | 4 → 4 | 2 → 2 | 5.00 → 5.00 | 43 → 49 | resiliency → resiliency |

**No stage transition, and this was the tightest case so far.** `ui_api` sits at Stage 4 with
`L = 2`, so *two* mature additions would have taken it to Stage 5. Under this category's 0.7/0.3
weights the 4.5 bar needs adoption 5 together with capability ≥4; the strongest product here
reaches 3.3. So the rung was safe by arithmetic rather than by luck — which is the thing I wanted
the report to be able to say out loud.

## What landed

| slug | openness | adoption | capability |
|---|---|---|---|
| `private-gpt` | 5 open_source | 3 stars_fallback | 4 |
| `qwenpaw` | 5 open_source | 2 usage_volume — 31,736/30d | 4 |
| `deepchat` | 5 open_source | 2 stars_fallback | 4 |
| `airi` | 5 open_source | 3 stars_fallback | 3 |
| `risuai` | 5 open_source | 2 stars_fallback | 3 |
| `ogam` | **4 open_core** | 2 stars_fallback | 3 |

## Two adjacent candidates, opposite gating answers

**`ogam` is open core, and the evidence is in the README, not the licence.** The MIT text is stock.
But the "Off Grid AI Pro" section prices a tier at **$69 for life or $49/year** and lists what it
adds over the published app: on-device text-to-speech, custom personas, approval-gated actions
through Calendar, email and MCP servers, and cross-device sync. Those are builds of the same
product — the `unsloth` shape — so `core-gated: gated`.

**`private-gpt` looks gated and is not.** Zylon is a separate end-to-end platform built on top of
this API layer (Triton + vLLM inference server, Kubernetes packaging, support), which is the
`langchain` shape: sold alongside, not a licence-gated build.

No automated licence check would have separated these. Both took a read.

## Package identity: four of six names belonged to someone else

Verified rather than assumed, and the hit rate was poor enough to be worth reporting:

- `private-gpt`, `private_gpt` — unregistered on PyPI
- `privategpt` — belongs to `vietanhdev/pautobot`
- `pgpt` — belongs to `hackedbyagirl/programengineergpt`
- npm `airi` — a 0.0.1 stub, no repository field, unrelated maintainer

None is declared. Declaring any one would have attached another project's download curve to a
product here, which is the #425 error class arriving through a different door. Across this tranche
and #431 that is five such traps in twelve candidates, which argues for #426 being a required gate
rather than a convenience.

## The instrument set, not the products, is why five of six band on stars

`private-gpt` has 57,491 stars and installs from a Homebrew tap. `ogam` ships through an app store.
`deepchat`, `airi` and `risuai` ship as desktop and browser builds. The routing table bridges PyPI,
Hugging Face and Semantic Scholar; it has no channel for any of those, so all five land on
`stars_fallback` at cap 3 regardless of real use.

`qwenpaw` is the instructive exception: it *does* have a verified PyPI package, and banding on it
gives **2** where its 34,740 stars would give 3. The authoritative route outranks the fallback, so
the lower number is the recorded one — the same shape as `areal` and `xtuner` in #431, and the same
open policy question.

**This matters for Wave 1.** Ranking census candidates by "evidence of real use" will systematically
under-rate everything distributed outside the four bridged registries. That is a property of the
instrument set, and worth knowing before it silently shapes the roadmap's priority order.

## Licences, read in full

`qwenpaw`'s LICENSE is ~590 bytes *shorter* than the canonical Apache template, which looked like
truncation. A word-by-word diff found exactly one differing region in the entire terms section — an
inserted "the" before "Licensor" in the definition of Contribution (ratio 0.99964) — with the
shortfall explained by the appendix instructions being replaced by the filled-in copyright notice.
Materially verbatim. Worth contrasting with `kaito`, which the ledger holds *unresolved* because a
section-2 grant was actually rewritten; that is the distinction a byte-level read exists to make.

`risuai`'s GPL-3.0 differs from the canonical text in one region: the copyright holder line.

## One candidate held

`maibot` is not promoted. It is a real, active, GPL-3.0 project (5,860 stars, licence byte-identical
to canonical GPL-3.0), but its README is largely Chinese and the call between `ui_api` and
`orchestration_agents` is genuinely ambiguous — it is a conversational persona that runs inside
messaging platforms. That is not a category call to make on thin English evidence. Its ledger entry
lands in a follow-up commit once this PR has a number, since inventing the number before the PR
exists is how #424's provenance got corrupted.

## Files

Six products, six scores, the `ui_api` roster, six new org files (`zylon-ai` and `off-grid-ai` as
companies, `kwaroran` as an individual, and `agentscope-ai`, `thinkinaixyz` and `moeru-ai` as
`unknown` — the documented value for an unconfirmed affiliation; QwenPaw's README credits "the
AgentScope team", with Alibaba appearing only as a distribution channel, so folding it into
`alibaba-cloud` would have been an inference rather than a reading). `long_tail.json` re-synced
539 → 545, plus the corpus goldens, every delta the expected size: axis rows +18 = three axes ×
six products, reconciliation +6, parity +6.


ccerv1:
Rebased onto main after #431 and #432 landed. New head `1e0094e`.

Conflicts in two places, both resolved by merging rather than side-picking:

**Derived values** — `long_tail.json` and four golden pins. Took main's side, then recomputed from the combined corpus. Every delta is the expected size for the 14 products across all three PRs:

| golden | baseline | now | delta |
|---|---:|---:|---|
| `axis_results` / reconciliation | 539 | 553 | +14 products |
| axis rows | 1617 | 1659 | +42 = 3 axes × 14 |
| `check_parity` computed | 534 | 548 | +14 |
| `long_tail.counts.scored` | 539 | 553 | +14 |

**The ledger** — main carried #431's `tinker-cookbook` and `mlx-vlm` entries while this branch added `maibot`. Both sides were pure appends, so the resolution is the union, and all three verify with distinct identities and correct per-PR provenance:

```
thinking-machines-lab/tinker-cookbook   existing_product | #431 finetuning_code pass
blaizzy/mlx-vlm                         unresolved       | #431 finetuning_code pass
mai-with-u/maibot                       unresolved       | #433 ui_api pass
```

Worth noting the `DuplicateResolution` guard would **not** have caught a side-pick here: dropping main's two entries produces no duplicate, so the file would have loaded clean with two decisions silently gone. Uniqueness protects identity integrity; it does not protect completeness. A count floor for the ledger follows as a separate gate change rather than riding in this PR.

Stage impact re-run against the new main is unchanged: `ui_api` stage 4 → 4, L 2 → 2, B 5.00, 43 → 49 products, no transition.

Full preflight green on this head: all 15 locally-runnable CI steps.

PR #511, opened 2026-09-07
gap-map-candidates: Run the weekly candidate batch: sweep sources, emit registry

ccerv1:
# gap-map-candidates — pass

**Goal.** Run the weekly candidate batch: sweep sources, emit registry rows, and open one `tail-batch` PR with the review sheet. This is the volume tier; rows carry no LLM judgment.

## Success criteria
- [ ] 50–100 new registry rows across categories, each with at least one addressable artifact identifier (not a URL)
- [ ] the batch summary publishes the five reconciled counts and they balance (raw = duplicates + unique; unique = accepted + parked)
- [ ] every parked candidate has a reason; no legitimacy rule invented mid-sweep
- [ ] `build.check_corpus_diff` review sheet shows zero stage or gap moves
- [ ] `uv run python -m build.validate` prints 0 error(s)
- [ ] the diff touches only `sources/registry/**` and the summary

## Outcome
`pass` after 4 round(s). Branch `loop/gap-map-candidates-2026-09-07-a1`.

## Final verdict
```json
{
 "verdict": "pass",
 "blocking": [],
 "non_blocking": [
  {
   "where": "docs/sweeps/2026-09-07-tail-batch.md:2329",
   "issue": "Non-blocking: five DeepSeek Coder folds name deepseek instead of deepseek-coder; four Llama Guard folds starting at line 2497 name llama instead of llama-guard. Apply the declared longest-pattern rule to these labels. Both correct targets are existing heads, so duplicate status and all five totals remain unchanged."
  },
  {
   "where": "docs/sweeps/2026-09-07-tail-batch.md:3317",
   "issue": "Non-blocking verification limitation: uv could not acquire its lock in this read-only environment. Direct validation produced 0 error(s), 2 warnings. The reported empty corpus-diff sheet was not independently rerun because it requires a writable worktree."
  }
 ],
 "settled_now": [
  "The 300 unsupported folds are now individually parked with provenance and reasons, resolving the round-3 blocker.",
  "Independent table recount confirms 739 duplicates and 1773 parks: 288 individually reasoned, 1185 cutoff, and 300 withdrawn folds.",
  "All five totals reconcile: 2579 = 739 + 1840; 1840 = 67 + 1773.",
  "Registry files are unchanged from round 3: 67 rows across five categories, with the diff restricted to registry files and the summary.",
  "Conservatively parking unsupported identity matches requires no policy escalation."
 ],
 "escalation": null
}
```

## Settled ground this run added (23)
- identity handle thresholds, org-handle ownership model, mirror/aggregator design
- the tail is public as an observed product with no axis scores
- coverage that could be higher is not a finding
- The diff contains 71 new rows and touches only registry files and the summary.
- No URLs occur in identifier fields; homepage is the documented URL exception.
- The five published totals balance arithmetically.
- No new slug matches a current head slug or retired alias; direct head artifact comparisons found no matches.
- The reported fixture and coverage-baseline failures are non-blocking under ADR-004 point 3.
- Unavailable warehouse enrichment and potentially higher coverage are not blockers.
- Round 2 contains 52 new rows across five categories; every row has a non-URL artifact identifier.
- No new row directly duplicates a main-branch registry slug or declared artifact.
- The summary individually lists 215 reasoned parks, 844 cutoff parks and 708 duplicate mappings, matching its published totals.
- The acknowledged mirror consolidation and homepage-only hardware parks resolve their round-1 blockers.
- Deterministic normalization and existing declared identity mappings are acceptable; verbatim checkpoint-level slugs are not required.
- Round 3 adds 67 rows across five categories and touches only registry files and the summary; every new row carries a non-URL identifier.
- The revised identity derivations and removal of all nine sub-floor exceptions resolve the round-2 blockers.
- No new row directly matches an existing registry slug, declared artifact, or resolution-ledger entry.
- The summary lists 288 individually reasoned parks, 1185 cutoff parks, and 1039 duplicate mappings. Its five totals balance arithmetically; the unsupported duplicate classifications remain blocking.
- The 300 unsupported folds are now individually parked with provenance and reasons, resolving the round-3 blocker.
- Independent table recount confirms 739 duplicates and 1773 parks: 288 individually reasoned, 1185 cutoff, and 300 withdrawn folds.
- All five totals reconcile: 2579 = 739 + 1840; 1840 = 67 + 1773.
- Registry files are unchanged from round 3: 67 rows across five categories, with the diff restricted to registry files and the summary.
- Conservatively parking unsupported identity matches requires no policy escalation.

Later runs of this job inherit `/workspace/GitHub/personal/agent-loops/jobs/settled/gap-map-candidates.json`; prune it there.


ccerv1:
## What I did

- **files touched** (all in `/home/vscode/.claude/workflows/runs/gap-map-candidates-2026-09-07-a1/build`, committed as `eb7c1257` on `loop/gap-map-candidates-2026-09-07-a1`):
  - `sources/registry/dataset_processing_tools.yaml` — new file, 14 rows
  - `sources/registry/safeguards.yaml` — new file, 19 rows
  - `sources/registry/edge_hardware.yaml` — new file, 15 rows
  - `sources/registry/base_pretrained.yaml` — new file, 14 rows
  - `sources/registry/finetuned_chat.yaml` — new file, 2 rows
  - `sources/registry/compilers.yaml` — appended 7 rows (existing 3 rows and their comment block untouched; a dated note inserted immediately before `category:`)
  - `docs/sweeps/2026-09-07-tail-batch.md` — new file, the batch summary / review sheet (528 lines). No `docs/sweeps/` convention existed, so this creates it; the registry headers point at it by path.
  - Nothing else. No product, score, category-roster, taxonomy, `org_handles.yaml` or test-fixture file, and neither generated artifact (`build/notebook_data.json`, `notebooks/ai-stack-map.py`).

- **71 rows, all with an addressable identifier, none a URL** (except `homepage`, the one field that is one):
  by type: 31 `software`, 25 `model`, 15 `hardware`. By artifact kind: 31 `github`, 25 `huggingface_model`, 23 `homepage` (of which 15 are the homepage-only hardware rows and 8 accompany a `github` id).
  `grep -nE '^\s+(github|pypi|npm|crates|arxiv|huggingface_model|huggingface_dataset):\s*https?://' sources/registry/*.yaml` → no matches.

- **commands run**
  - `uv sync` (the clone had no venv)
  - Category list read at run time through `build.taxonomy.category_statuses(yaml.safe_load(...))` — never a remembered list, never a hand parse. All 18 categories are `published`.
  - Corpus dedup sets built from the tree with `build.identity.fold_for_proposal` / `id_from_url`: 616 head slugs, 63 retired aliases, 856 folded head artifacts, 33 existing registry slugs + 65 registry artifacts, 40 `resolution_ledger.yaml` rulings, 26 `model_families.yaml` patterns.
  - 36 GitHub repository searches (`api.github.com/search/repositories`, sort by stars, `per_page=30`), 10 Hugging Face model queries (`limit=100`), 35 vendor product pages by `curl -sL` with an HTTP-status and `<title>` check. All 2026-09-07.
  - `uv run python -m build.validate` → **0 error(s), 2 warning(s)** (both pre-existing `model_families` overlap notices, byte-identical on `main`)
  - `uv run python -m build.check_corpus_diff --base main` → sheet below; exit 0
  - `uv run pytest -q` → 6 failed, 1827 passed, 1 skipped — see **Risks**; all six are the ADR-004 point-3 class
  - `git worktree add … main` + `pytest tests/test_identity_eval.py` at `9df82a1` → 137 passed, worktree removed (`git worktree list` shows only the clone; `git status` clean)

- **the five reconciled counts** (computed by a script over every raw signal, not asserted by hand):

  | count | value |
  |---|---|
  | `raw_signals` | 1819 |
  | `duplicate_signals` | 701 |
  | `unique_candidates` | 1118 |
  | `accepted` | 71 |
  | `parked` | 1047 |

  `1819 = 701 + 1118` ✓ and `1118 = 71 + 1047` ✓. Each accepted signal resolves to exactly one emitted row: 71 signals, 71 rows. Of the 1047 parked, **203 carry an individual reason** and **844 fell below a disclosed retrieval cutoff** and are grouped with the cutoff that excluded them — recorded, not dropped.

- **`build.check_corpus_diff --base main` review sheet, verbatim:**

  ```
  ## Review sheet

  | category | stage | gaps | products | tier changes |
  |---|---|---|---|---|

  stage moves: none

  untouched-product row changes: none
  ```

  Empty by design: registry rows are signal-only and invisible to the scored payload, so a `tail-batch` PR moves no stage, gap, product count or tier, and needs no `stage-move` label.

## Why

- **decisions**
  - **Scope = the six categories the corpus itself calls thin.** The taxonomy carries no thinness marker and every category is `published`, so I derived thinness from the map's own gap arithmetic in the serialized payload: maturity stage ≤ 3. That is `dataset_processing_tools` (stage 2, `adoption` gap), `edge_hardware`, `safeguards`, `base_pretrained`, `finetuned_chat` and `compilers` (all stage 3). Stated in the summary so the next sweep can reproduce the selection.
  - **Predeclared retrieval cutoffs, then disclosed coverage limits.** GitHub: top 30 per query, ≥ 1,000 all-time stars, push no older than 2025-09-07. Hugging Face: top 100 per query, ≥ 1,000,000 trailing-30-day downloads on the seven ranked queries and ≥ 5,000 on the three targeted guard/safety/moderation searches. This is a *retrieval* bound, which the workflow explicitly allows, not a legitimacy rule: nine sub-floor candidates the sweep recognized were accepted anyway (`data-prep-kit`, `semhash`, `deepfabric`, `magpie`, `datadreamer`, `bonito`, `gpt-j`, `lfm2`, `exaone`), each flagged in its row note, and the 1,260 rows outside the floors are reported as an un-triaged coverage limitation with the floor stated.
  - **GitHub search is the wrong instrument for `edge_hardware`, so I changed source rather than pad.** Six board-and-chip queries returned 106 rows and not one board — every in-cutoff hit was edge *software*, parked as such. The 15 rows come from 35 vendor product pages, each confirmed 200 with the product's own `<title>`. All carry `homepage` and nothing else, and the file header says plainly that this is the weakest evidence kind on offer.
  - **Model categories are families, so releases fold.** 411 raw signals fold onto an existing family or head product. The bare `model_families.yaml` glob misses the release shape (`qwen-*` never matches `qwen3-8b`; head slug `granite-guardian` never matches `granite-guardian-3.3-8b`), which is the `amazon-nova-pro` hazard `docs/reference/identity.md` names, so the dedup matches `^<base>[0-9]*([-.].*)?$` against both family bases and head slugs. Two false positives it produced (`FareedKhan-dev/kimi-k3-in-c`, `TingsongYu/PyTorch-Tutorial-2nd`) are overridden back to unique candidates and parked on their own merits, and five owner-prefixed names it missed are overridden to duplicates — both directions recorded in the decisions file rather than silently applied.
  - **What `base_pretrained` was actually missing** is the encoder and early-decoder generation the map never recorded — BERT, RoBERTa, XLM-RoBERTa, T5, ELECTRA, DeBERTa, ModernBERT, DistilBERT, GPT-2, GPT-J, OPT — still drawing tens of millions of downloads a month, plus three recent lines (OpenELM, PowerMoE, Liquid LFM2). The ranked queries are otherwise dominated by releases of families already present.
  - **`pytorch/xla` is emitted as `pytorch-xla`, not `xla`.** Head product `xla` is the OpenXLA compiler; this is the PyTorch bridge to it. The slug near-collision is called out in the row note.
  - **Escalated instead of stretching a category** (all in the summary's Escalations section): 50 heavily used models the taxonomy has no category for — embedding/reranking (18, led by `all-MiniLM-L6-v2` at 251M downloads a month, more than any model on the map), speech/audio (13), vision and VLM backbones (9), generative media (5), time series (5); Tenstorrent's Blackhole cards, because nothing covers datacenter accelerators; an unresolved text-versus-vision boundary in `dataset_processing_tools` (BlenderProc, fastdup, fiftyone); two `compilers` boundary questions (Enzyme, CIRCT); and two re-surfaced `unresolved` ledger holds (`turbovec`, `failproofai`) that were not re-proposed.
  - **Two candidates parked on a number, with the number stated**, per the no-mid-sweep-rules rule: `Zaneham/Booth` (1,737 stars on a single-owner repository created 2026-02-16 claiming a full CUDA/Triton/HIP compiler) and the three `farbodtavakkoli/OTel-*-LLM-*` models (6.9M downloads/30d on a 31B model under an individual account with no resolvable lineage). Neither park was generalized into a threshold applied to anything else.
  - **No fetch recorded as an absence.** GitHub search was consumed at ~8.5 req/min against the unauthenticated ceiling of 10, with backoff on 403/429/5xx. Five vendor pages 404'd on the paths tried (SiMa.ai, Kinara, Blaize, Orange Pi, Rockchip) and are parked as incomplete fetches.

- **alternatives rejected**
  - *Sweeping all 18 categories.* Volume the reviewer could not read, and the thin categories are where a row changes the reading.
  - *Counting only in-cutoff rows as `raw_signals`.* Tidier arithmetic, weaker guarantee. Counting all 1,819 and parking the sub-floor rows with the cutoff named means nothing was dropped without a trace, which is the point of the five counts.
  - *Bucket-parking above-cutoff rows.* The `storage.yaml` precedent explicitly warns against it, so all 203 above-cutoff non-accepts carry an individual reason with a source URL and fetch date.
  - *Re-pinning `identity_coverage_baseline.json` or regenerating `identity_edges_pass.json` to make pytest green.* Forbidden on a proposal PR (ADR-004 point 3) and it would put the diff outside `sources/registry/**` plus the summary. Recorded instead.
  - *Adding `sources/org_handles.yaml` entries for the new organizations.* Same reason. It is the correct follow-up, on `main`, in its own commit.
  - *Inventing artifacts for hardware with no product page.* Would satisfy the schema and lie. Parked.
  - *Delegating judgment to subagents.* Used none. The skill says this parallelizes on fetching and badly on judgment; fetching was already parallel through the scripts, and dedup, category assignment and boundary calls depend on the whole candidate set. Effort went into per-candidate reasons instead.

## Risks / unsure about

- **`pytest` is red: 6 failed, 1827 passed, 1 skipped — all six in `tests/test_identity_eval.py`, and I deliberately did not repair them.** Four want `tests/fixtures/identity_edges_pass.json` regenerated, two want `tests/fixtures/identity_coverage_baseline.json` re-pinned. One cause: the new rows add artifacts whose organizations declare no handle, so coverage denominators grow while numerators hold exactly — `github` 211/299 → 211/321, `huggingface` 51/62 → 51/75, `homepage_domain` 6/27 → 6/48. Nothing lost a handle. All 137 tests in that file pass on `main` (checked in a scratch worktree at `9df82a1`), so this is caused by the batch. ADR-004 point 3 names "identity membership fixtures and committed coverage baselines" as gates that bind on `main` and do not block a proposal, and forbids a proposal from touching them. **A reviewer who disagrees with that reading is the main way this round fails** — the fix would be an `org_handles.yaml` commit on `main`, not an edit here.
- **The warehouse discovery pool was never reached** (no `OSO_API_KEY`). It is enrichment and consolidation, never rejection, so the dedup stands — but multi-signal consolidation rested on self-dedup alone, and a re-run with warehouse access would probably fold some signals this sweep treated as separate.
- **The 15 `edge_hardware` rows are the weakest in the batch.** `homepage`-only means no adoption signal, no code, and no identity-bearing artifact — `validate` does dedup homepages, but two products under one vendor path would pass. I checked these by hand; a reviewer should treat them as lower-confidence than any repo- or package-backed row, as both the file header and the summary say.
- **Three SKU-versus-product calls I could be wrong about.** `google-coral-dev-board-micro` accepted as distinct from head `google-coral-dev-board` (different host silicon and form factor); `khadas-vim4` accepted alongside `khadas-edge2`; `hailo-8l` parked as a SKU of `hailo-8`. If the reviewer reads Coral Micro as a SKU, that row comes out.
- **`presidio` is the boundary call most open to challenge** in `safeguards`: it is a general PII detection and redaction framework, not AI-specific, accepted because PII filtering on model inputs and outputs is what the head product `llm-guard` already does. `uber/ADR` and `akto` are the reverse call, parked.
- **Six of the seven `base_pretrained` encoder rows are 2022-era models.** They are correct as candidates — heavily used, absent from the map, distinct products — but they add historical breadth rather than frontier coverage, which may not be what a thin-category sweep is wanted for.
- **The batch is large to review.** 71 rows plus 203 individually reasoned parks is a long sheet, and the sub-floor group is 844 rows deep. If the reviewer wants a smaller batch, the lever is fewer categories per week, not a higher floor.
- **`docs/sweeps/` is a directory I invented.** The workflow puts the summary in the PR body and there was no file convention; a maintainer may want it elsewhere, which is a one-line move plus six header edits.
- **Provenance beyond the summary is session-local.** Raw API pulls, the decisions file and the reconciliation script live in `/tmp/claude-1000/-workspace-Obsidian/931860aa-e5e5-41f0-bfc6-eef224455303/scratchpad/` (`raw/*.json`, `decisions.py`, `reconcile.py`, `emit.py`, `summary.py`) and are not committed. Every query URL and every per-candidate decision is reproduced in `docs/sweeps/2026-09-07-tail-batch.md`, which is the durable record.

## Response to round N-1

Round 1 — nothing to respond to.

|---|---|---|---|---|
| _n/a (round 1)_ | — | — | — | — |

ccerv1:
## What I did

- **files touched** (all in `/home/vscode/.claude/workflows/runs/gap-map-candidates-2026-09-07-a1/build`, committed as `038ddde2` on `loop/gap-map-candidates-2026-09-07-a1`; round 1 was `eb7c1257`):
  - `sources/registry/edge_hardware.yaml` — **deleted**. It was created in round 1; all 15 of its rows carried `homepage` and nothing else.
  - `sources/registry/dataset_processing_tools.yaml` — 14 → 13 rows
  - `sources/registry/safeguards.yaml` — 19 → 17 rows
  - `sources/registry/base_pretrained.yaml` — 14 → 12 rows
  - `sources/registry/finetuned_chat.yaml` — 2 → 3 rows
  - `sources/registry/compilers.yaml` — reset to `main`'s file and re-appended with 7 rows and a rewritten dated note (the pre-existing three rows and their comment block are byte-identical to `main`)
  - `docs/sweeps/2026-09-07-tail-batch.md` — regenerated, 2,293 lines (about 400 KB)
  - Nothing else. No product, score, category-roster, taxonomy, `org_handles.yaml` or test-fixture file, and neither generated artifact.

- **52 rows across five categories**, every one carrying an addressable identifier that is not a URL: 30 `github`, 22 `huggingface_model`; 30 `software`, 22 `model`; no `hardware`. Eight rows carry a `homepage` as a *second* artifact on a row that already has an identifier. Checked over the whole registry, not just the new rows: all 85 rows in `sources/registry/*.yaml` have a non-URL identifier and no identifier field holds a URL.

- **the five reconciled counts**, recomputed from every raw signal rather than adjusted:

  | count | round 1 | round 2 |
  |---|---|---|
  | `raw_signals` | 1819 | 1819 |
  | `duplicate_signals` | 701 | 708 |
  | `unique_candidates` | 1118 | 1111 |
  | `accepted` | 71 | 52 |
  | `parked` | 1047 | 1059 |

  `1819 = 708 + 1111` ✓ and `1111 = 52 + 1059` ✓. 52 accepted signals resolve to 52 rows, one each. All 1,059 parked candidates are now listed individually — 215 with a reason of their own, 844 with the retrieval floor that excluded them and the query that returned them — and all 708 duplicates are listed with what they folded onto. The table row counts in the file were counted programmatically against the published totals: 215/215, 844/844, 708/708.

- **commands run**
  - `PYTHONPATH=. python3 …/scratchpad/reconcile2.py` — re-ran the whole dedup order over the same raw API pulls with the round-2 overlay (`r2_overlay.py`); the script fails loudly on any signal without a decision (exits 1 and lists them), and it exited clean.
  - `PYTHONPATH=. python3 …/scratchpad/emit2.py` and `summary2.py` — re-emitted the registry files and the sheet from `accept2.json`/`rows2.json`, so the files and the sheet cannot disagree. `compilers.yaml` is regenerated from `git checkout main -- sources/registry/compilers.yaml` each time, so the appended block never stacks.
  - `.venv/bin/python -m build.validate` (same as `uv run python -m build.validate`; `uv` could not take its lock in this environment on the first try) → **0 error(s), 2 warning(s)** (both pre-existing `model_families` overlap notices, byte-identical on `main`)
  - `.venv/bin/python -m build.check_corpus_diff --base main` → exit 0, and the sheet is empty:

    ```
    ## Review sheet

    | category | stage | gaps | products | tier changes |
    |---|---|---|---|---|

    stage moves: none

    untouched-product row changes: none
    ```

  - `pytest -q` → **5 failed, 1828 passed, 1 skipped in 654s** (all five in `tests/test_identity_eval.py`); `pytest tests/test_identity_eval.py -q` → **5 failed, 132 passed** (round 1 had six failures in that file; one fewer now because the homepage-only hardware rows are gone). Live coverage is `github` 211/320, `huggingface` 51/73, `homepage_domain` 6/35 against pinned 211/299, 51/62, 6/27 — denominators grew, numerators held exactly.
  - `git worktree add … main` + `pytest tests/test_identity_eval.py` at `9df82a12` → **137 passed**; worktree removed afterwards.

## Why

- **decisions**
  - **Hardware is parked, not patched.** No vendor publishes a product-level repo, package, Hub entry or paper id for a board or NPU, and the workflow says in as many words not to invent a plausible artifact for a candidate that has none. So all 15 go to the parked table with their page, fetch date and that reason, `edge_hardware.yaml` is not created, and the sheet carries a short section saying what a later sweep should look for (a vendor-account SDK repo, a package, a datasheet DOI) plus the question for a person: whether the map wants a homepage-only hardware tier at all.
  - **The mirror consolidation covers the whole class, not just the two named.** Seven signals whose *recorded round-1 reason already acknowledged a fold* moved to the duplicate side: the two `autogluon/chronos-*` mirrors, a Core ML conversion of a pooled Whisper signal, an abliterated GGUF redistribution of a pooled Qwen signal that folds onto head product `qwen`, a detector published by the org behind head product `guardrails-ai`, a web UI over head product `iree`, and the ComfyUI plugin surface of the `nunchaku` row this batch emits. Each referent was checked to exist in the pool or in the head corpus before folding; the sheet lists them in their own table with what each folds onto.
  - **Two lookalikes were deliberately *not* folded**, because the evidence does not settle them: the two `prism-ml/*Bonsai*` quantizations whose upstream trainer is unidentifiable from the card, and `distilbert/distilgpt2`. Parking an ambiguity is what the workflow asks for; folding it would be the guess it warns against.
  - **Judgment may park a candidate; it may no longer emit one.** That is the invariant I implemented for ADR-004 point 4. Category on an emitted row now comes from a declared mapping — the predeclared query→category mapping for a GitHub signal, or a published declared-metadata rule (M1/M2/M3 over `tags`/`pipeline_tag`) for a Hugging Face signal — and the sheet carries a per-row derivation table naming the observation behind every field. Four rows whose acceptance or category rested on an argument are parked (`presidio`, `unlimited-ocr`, `nsfw-image-detection`, `electra`) and `lfm2` moved to the category the rule names (it declares `conversational`, so `finetuned_chat`).
  - **Provenance is dumped in full rather than summarized.** The sheet is now about 400 KB because 1,059 parked candidates and 708 duplicate mappings are listed row by row with identifiers, URLs, fetch dates, the query that returned them and the applicable floor. That is what makes the reconciliation reproducible from the sheet alone, which is what the review asked for.
  - **The applicable floor is stated per candidate, not per source.** I also checked whether any signal was below one query's floor and inside another's — zero across all ten Hugging Face queries — so each row's label is the only floor it was ever measured against, and the sheet says so.

- **alternatives rejected**
  - *Finding identifiers for the boards anyway* — a vendor SDK repo (`hailo-ai/hailort`, `espressif/esp-idf`) is the software, not the board. Emitting it would put a false artifact on a hardware row.
  - *Lowering a retrieval floor to backfill rows after the parks* — that would make the floor an outcome-driven knob instead of a predeclared bound. 52 rows clears the 50 floor without touching it.
  - *Transcribing HF ids verbatim into `slug`/`display_name`* (the maximal reading of ADR-004 point 4) — see the pushback in the response table: it would emit `roberta-base` and `roberta-large` as two products, against `docs/reference/identity.md` §"How far to collapse: what the vendor sells", and it would re-split four release variants this sweep folds.
  - *Adding `sources/org_handles.yaml` entries, or regenerating either identity fixture, to make pytest green* — forbidden on a proposal PR by ADR-004 point 3, and it would put the diff outside `sources/registry/**` plus the summary. Recorded in the sheet as the follow-up on `main` instead.
  - *Subagents.* None used. The work was re-deriving one reconciliation and one sheet from data already fetched; splitting it would have risked the registry files and the sheet disagreeing.

## Risks / unsure about

- **52 rows is two above the 50-row floor.** Every park in this round was forced by a blocking item, but if a reviewer parks even three more rows the batch drops below the job's volume requirement, and the fix then is a lower retrieval floor in a re-run, not a re-argued accept.
- **The declared-metadata rule (M1/M2/M3) is published in this sheet, not in a reference doc.** I read it as a mapping from declared metadata to a category rather than a legitimacy threshold — it never adjudicates whether a product is real, and it can only park or route, never admit a candidate the sweep would otherwise have parked. A reviewer who reads any published rule as a mid-sweep rule will disagree; the remedy is a rule in `docs/reference/`, proposed separately, which this workflow cannot open.
- **`slug`, `display_name` and `org` are still transcriptions on many rows, and the sheet marks which.** 29 of the 52 rows name an organization with no record in `sources/organizations/`, and on 46 the owner login is not a declared handle for the organization named. That is disclosed row by row rather than fixed, for the reason in the response table.
- **`pytest` is red on five identity-eval tests and I again did not repair them** — same ADR-004 point 3 reading as round 1, which the reviewer settled as non-blocking. If that reading changes, the batch is unmergeable as built rather than wrong.
- **The sheet is very long to read.** 2,293 lines, and the three provenance tables are 1,767 of them. The counts in every table header were verified against the tables, but nobody will read the tables end to end; they exist to be greppable and reproducible.
- **`docs/sweeps/` is still a directory this batch invented**, unchanged from round 1 and not raised in review. A maintainer who wants the sheet elsewhere pays one file move plus five header edits.
- **The warehouse discovery pool was still unreachable** (no `OSO_API_KEY`), so multi-signal consolidation still rests on self-dedup. Settled as non-blocking in round 1; unchanged here.
- **Round-1 raw pulls were reused rather than re-fetched.** The same `raw/*.json` files from 2026-09-07 back this revision, so every fetch date in the sheet is honest, but nothing in the outside world was re-checked today.

## Response to round N-1

|---|---|---|---|---|
| `sources/registry/edge_hardware.yaml:27` — 15 hardware rows carry only a homepage URL | ✅ | | | All 15 parked with page, fetch date and the reason; `edge_hardware.yaml` deleted, so no row in the batch lacks a non-URL identifier. Verified over the whole registry, not just the diff: 85 rows, all with a non-URL identifier, none holding a URL in an identifier field. 52 rows remain across five categories (≥ 50). No artifact was invented for a board — `docs/workflows/discover-candidates.md` §5 forbids it — and the sheet's new `edge_hardware`: swept, nothing emitted section says what a later sweep should look for and escalates the underlying question (whether a homepage-only hardware tier should exist) to a person. |
| `docs/sweeps/2026-09-07-tail-batch.md:382` — acknowledged mirrors counted as unique candidates | ✅ | | | `autogluon/chronos-2` and `autogluon/chronos-bolt-small` now fold onto the pooled `amazon/chronos-2` and `amazon/chronos-bolt-small` signals, and five more signals of the same recorded class fold too (Core ML conversion of `openai/whisper-large-v3`; abliterated GGUF of `Qwen/Qwen3.8-27B` → head product `qwen`; `GuardrailsAI/prompt-saturation-attack-detector` → head product `guardrails-ai`; `nod-ai/AMD-SHARK-Studio` → head product `iree`; `nunchux-ai/ComfyUI-nunchaku` → the `nunchaku` row this batch emits). Each referent was confirmed present in the pool or the head corpus first. All five counts recomputed by re-running the reconciliation: 1819 = 708 + 1111, 1111 = 52 + 1059, both balances asserted in the script. Two lookalikes stay parked because the evidence does not settle them (`prism-ml/*Bonsai*`, `distilbert/distilgpt2`), and the sheet says why. |
| `docs/sweeps/2026-09-07-tail-batch.md:459` — 844 parked candidates represented only by aggregate counts | ✅ | | | The sheet now lists every parked candidate individually: 215 in *Parked — individually reasoned* and all 844 in *Parked — below a disclosed retrieval cutoff*, each with its identifier, source URL, fetch date, the query names that returned it and the specific floor that excluded it (GitHub star/push floor; HF ranked-query 1,000,000 or targeted-search 5,000, whichever applied to its query). A new *Duplicate mappings* section lists all 708 duplicates with source URL, queries and what each folded onto, which is the mapping behind `raw = duplicates + unique`. Table sizes checked against the published counts: 215/215, 844/844, 708/708. I also verified no signal was sub-floor on one query and in-cutoff on another (zero across the ten HF queries), so a single floor label per row is accurate. |
| `sources/registry/dataset_processing_tools.yaml:12` — category and acceptance depend on agent boundary judgments (ADR-004 point 4) | | ✅ | | **Marked pushed_back because one part of the remedy is contested, not because the item was ignored: the remedy is implemented in full for category assignment and for acceptance, and declined for three identity fields with the evidence below.** Category on an emitted row now comes only from a declared mapping: the predeclared query→category mapping for a GitHub signal (published as a table, and every row's category equals the category of the query that surfaced it), or the published declared-metadata rule M1/M2/M3 over the model's own `tags`/`pipeline_tag` for a Hugging Face signal. Acceptance is now a mechanical predicate (survives dedup 1–6, inside the disclosed cutoff or a disclosed sub-floor recognition, carries a non-URL identifier, has a derivable category); judgment can only park. The rows whose emission rested on an argument are parked — `presidio` (editorial boundary), `unlimited-ocr` (`image-text-to-text` from a ranked query, the same tag parked elsewhere as no-category-fits), `nsfw-image-detection` (`image-classification`, ditto), `electra` (declares no `pipeline_tag` at all) — and `lfm2` moved to the category the rule names. The product-versus-SKU calls the round-1 notes described were all in hardware and are gone with item 1; `hailo-8l` was already parked. A per-row derivation table now names the observation behind every field. **Residual, pushed back rather than fixed:** `slug`, `display_name` and `org` remain transcriptions of observed strings (a product line folded from a checkpoint id; an owner login mapped to an organization record), marked as such row by row. Transcribing them verbatim would emit `roberta-base` and `roberta-large` as two products and re-split four release variants this sweep folds, against `docs/reference/identity.md` §"How far to collapse: what the vendor sells" ("Pitch the slug at the level the vendor markets as the product, not at a release") — a convention `build/validate.py` enforces at promotion by rejecting a version or size token in a `base_pretrained`/`finetuned_chat` slug unless the product declares `version_in_identity`, a field a registry row cannot carry. ADR-004 point 4 read with point 2 (the tail renders "detected license, download band, sources" and never axis or overall scores) bars an LLM-derived *assessment* value, not an identity transcription; and `docs/workflows/discover-candidates.md` — aligned to ADR-004 one commit before this run in `ad309436` (#507), which changed only the bot-PR rule — still mandates step 4 "Assign exactly one category" and step 6's boundary parking. If the maximal reading holds, ADR-004 point 4 and this job's 50-row floor cannot both be satisfied in one batch — the 15 rows whose category comes from the declared-metadata rule rather than from a category-scoped query would go, leaving 37 — and that is a call for Carl, not for me to resolve by shipping either a thinner batch or checkpoint-level identity. |

ccerv1:
## What I did

Both blocking items are **fixed**, and both were fixed by changing the batch rather than the
argument. Round 2's pushback on the identity fields is withdrawn: the reviewer was right that
`answerdotai` → `answer-ai` is neither a transcription nor a declared mapping, and once the
mapping had to be mechanical the product-family folds could not survive either.

- **files touched** (all in `/home/vscode/.claude/workflows/runs/gap-map-candidates-2026-09-07-a1/build`,
  committed as `d041f6df` on `loop/gap-map-candidates-2026-09-07-a1`; round 1 was `eb7c1257`, round 2 `038ddde2`):
  - `sources/registry/base_pretrained.yaml` — 12 → 25 rows
  - `sources/registry/dataset_processing_tools.yaml` — 13 → 15 rows
  - `sources/registry/safeguards.yaml` — 17 → 18 rows
  - `sources/registry/compilers.yaml` — reset to `main`'s file and re-appended with 8 rows (the three pre-existing rows and their comment block are byte-identical to `main`)
  - `sources/registry/finetuned_chat.yaml` — 3 → 1 row
  - `docs/sweeps/2026-09-07-tail-batch.md` — regenerated, 3,275 lines / 565 KB
  - Nothing else. No product, score, category-roster, taxonomy, `org_handles.yaml`, `model_families.yaml` or test file, and neither generated artifact. `git diff --name-only main..HEAD` is those six paths.

- **67 rows across five categories** (`base_pretrained` 25, `safeguards` 18, `dataset_processing_tools` 15,
  `compilers` 8, `finetuned_chat` 1). Checked across the whole registry, not just the new rows: all
  100 rows in `sources/registry/*.yaml` carry an identifier that is not a URL, and no identifier
  field holds a URL. 24 rows carry a `homepage` as a *second* artifact on a row that already has an
  identifier.

- **Item B — acceptance is now a predicate with no free parameters.** The nine sub-floor exceptions
  round 2 accepted are parked with the floor that excluded them, and the code path that could admit
  a sub-floor candidate is gone rather than unused:

  ```
  accepted(signal) := in_cutoff and survived_dedup and derivable_category
                      and addressable_identifier and no_recorded_park_reason
  ```

  The only judgment-shaped clause is the last, and it can only remove a candidate. Round 2 built the
  accept set as a hand-written list and checked parks against it; round 3 derives the accept set as
  the complement of the parks, so a row cannot exist without an observation behind every field.

- **Item B, replenishment — a second retrieval pass at the same cutoffs.** 16 new queries (12 GitHub,
  4 Hugging Face), predeclared in `passB_manifest.json` — query, declared category, floor, sort and
  depth — written and saved *before* the first request went out, then fetched live today. GitHub:
  identical cutoff to pass A (≥ 1,000 stars, push since 2025-09-07, top 30 by stars). Hugging Face:
  the 1,000,000-download floor pass A used for `filter=`-shaped ranked listings.

- **Item A — every field is derived.** `slug` = `slugify(declared name)`, with
  `slugify(owner)-slugify(name)` as the declared collision fallback and a park if that still
  collides; `display_name` = the declared name verbatim; `org` = the organization
  `sources/org_handles.yaml` already declares for that `(platform, owner login)`, else
  `slugify(owner login)`; `type` from the signal's source; `category` from the query's declared
  category or the declared-metadata rule. 36 of the 43 rows carried over from round 2 changed a
  slug, a display_name or an org as a result. The three hand-made product-family folds are un-folded.

- **commands run**
  - `python3 …/scratchpad/fetchB.py` and `fetchB_hf.py` — the pass-B pulls, saved to `raw/ghB.json` (360 items) and `raw/hfB.json` (400 items), all fetched 2026-09-07.
  - `PYTHONPATH=. python3 …/scratchpad/reconcile3.py` — one pass over all 2,579 (query, item) occurrences from both passes; asserts both balance equations and writes `rows3.json` / `emit3.json`.
  - `python3 …/scratchpad/emit3.py` and `summary3.py` — registry files and the sheet are both generated from `emit3.json`, so they cannot disagree. `compilers.yaml` is regenerated from `git show main:` each run, so the appended block never stacks.
  - `uv run python -m build.validate` → **0 error(s), 2 warning(s)** (both pre-existing `model_families` overlap notices, byte-identical on `main`).
  - `uv run python -m build.check_corpus_diff --base main` → exit 0, empty sheet: no stage move, no gap move, no product-count or tier change, no untouched-product row rewritten.
  - `uv run pytest -q` → **6 failed, 1827 passed, 1 skipped in 657s**; `pytest tests/test_identity_eval.py -q` → **6 failed, 131 passed**. Live handle coverage `github` 211/328, `huggingface` 51/81, `homepage_domain` 7/50 against pinned 211/299, 51/62, 6/27 — denominators grew, numerators held or rose.
  - `git worktree add … main` + `pytest tests/test_identity_eval.py` at `9df82a12` → **137 passed** (worktree removed afterwards), so the failures are the batch's and not pre-existing.

## Why

- **decisions**
  - **Checkpoint-level identity is the honest grain for a machine batch.** A mechanical rule can read
    `roberta-base` off an API response; it cannot know that RoBERTa is one product with two sizes.
    So the batch proposes `roberta-base` and `roberta-large` as two rows and says, in the sheet and
    in the registry header, that collapsing them is the editorial act a promoter performs — where a
    `model_families.yaml` pattern or a `version_in_identity` declaration records the decision. That
    is exactly where ADR-004 point 4 puts it, and it is the cost of the rule rather than an argument
    against it. Round 2's counter-argument (that `validate` would reject the collapsed slug) was
    about *head* products: the version-token check reads category rosters, so it never touches a
    registry row. I checked that this round instead of asserting it.
  - **`org` is the declared handle or the slugified login, never a name I chose.** 55 of the 67 rows
    now name an organization with no `sources/organizations/` record (45 distinct slugs), most of
    them reading like a login (`answerdotai`, `bespokelabsai`, `facebookai`,
    `data-prep-kit`). That is uglier than round 2 and it is the point: the pretty version was
    invented. The repair is declared handles in `sources/org_handles.yaml` on `main`, which the sheet
    escalates and this workflow cannot open.
  - **Removing the exceptions had to cost rows, and the replenishment goes wider, not lower.** Nine
    parks took the batch to 46 rows, under the job's 50-row floor. The workflow's stated remedy for
    a volume problem is a disclosed retrieval bound, so I added queries at the *same* floors rather
    than dropping a floor to re-admit the nine. Nothing in this batch is accepted below a cutoff.
  - **One manifest floor was corrected upward, and the correction is disclosed.** The manifest first
    gave the four pass-B Hugging Face queries the 5,000-download floor pass A used for `search=`
    keyword queries. All four are `filter=`-shaped ranked listings — pass A's own convention gives
    that shape 1,000,000 — so the batch applies 1,000,000. The correction only ever tightens: it
    excludes signals the manifest had admitted and admits nothing the manifest excluded. At 5,000 the
    pass would have proposed 247 rows, mostly research checkpoints (`julien-c/dummy-unknown`,
    `facebook/esm2_t6_8M_UR50D`, dozens of language-specific BERT fine-tunes) that no mechanical rule
    separates from a vendor's base model. I recorded this rather than quietly picking the floor that
    produced a nice number.
  - **M1 gained a clause that can only park.** A model declaring `conversational` alongside
    `image-text-to-text` is an OCR or vision-language model, and round 2's M1 would have routed it
    into `finetuned_chat`. M1 now also requires the declared `pipeline_tag` to be `text-generation`
    or absent. Two candidates move to the parked side because of it (`datalab-to/chandra-ocr-2`,
    `llava-hf/llava-1.5-7b-hf`); none is admitted by it.
  - **One round-2 fold survives, on a predicate rather than a reading.** `LiquidAI/LFM2.5-2.6B-GGUF`
    folds onto `LiquidAI/LFM2.5-2.6B`: same owner login, name equal to the other signal's name plus
    `-GGUF`, and the repo declares `gguf` in its own tags; the representative is the shorter name.
    The seven mirror folds the reviewer settled in round 2 are untouched.
  - **The 66 new pass-B parks name the boundary the roster already draws.** Offensive-security and
    security-operations tooling that *uses* an LLM is parked; a safeguard applied *to* an AI system's
    inputs, outputs or actions is emitted. A document tool whose declared description is a human
    workflow (scan, archive, manage, search, view, edit, translate) is parked; one that names data
    prepared for a model is emitted. Both lines are the sweep's reading of an existing roster, both
    only ever park, and both are escalated in the sheet for a person to confirm or move.

- **alternatives rejected**
  - *Keeping the identity pushback.* The reviewer's example was decisive: no observation produces
    `answer-ai` from `answerdotai`. Once that is conceded, `roberta-base` + `roberta-large` →
    `roberta` falls with it, because both are the same act at a different grain.
  - *Parking the nine and shipping 46 rows.* Under the job's 50-row floor, and the reviewer named
    replenishment as the remedy.
  - *Lowering the GitHub star floor to 800 to re-admit the exceptions mechanically.* It would have
    worked and it would have been a floor chosen by looking at which rows I wanted back. A new query
    set at an unchanged floor is the same volume of work without that defect.
  - *Emitting the 247 rows the 5,000-download floor produced.* Over the job's 100-row ceiling, and
    most of them are research checkpoints the map's `base_pretrained` roster is not about. Choosing
    among them by hand is precisely the recognition the reviewer rejected.
  - *Dropping the 24 declared homepages to keep `test_a_stale_fixture_tail_row_…` green.* That test's
    arithmetic assumes a 27-row truth set. Suppressing observed evidence to satisfy it is worse than
    reporting it; `discover-candidates` says to write every artifact there is evidence for.
  - *Editing `tests/` or `sources/org_handles.yaml` to make the suite green.* Out of the diff scope
    the job sets and, for the fixtures, forbidden by ADR-004 point 3.
  - *Subagents.* None used. The judgment-bearing work was ~340 individual triage calls that had to
    apply one boundary line consistently; splitting that across agents is how the line drifts.

## Risks / unsure about

- **A sixth test now fails, and its repair is not a fixture edit.**
  `test_a_stale_fixture_tail_row_fails_the_floor_with_the_publish_lag_note` mutates one `homepage`
  row of a truth set it assumes is ~27 rows and asserts the 0.98 precision floor catches it. With 50
  tail homepage rows, one bad row is a 2.0% error, precision lands exactly on the floor, and the
  mutation no longer trips it. Nothing regressed — the floor gained headroom — but the fix is a
  one-line change in `tests/`, which this batch's diff scope forbids. ADR-004 point 3 names fixtures
  and coverage baselines, not this, so I am not claiming it covers this one. It is reported in the
  sheet as a distinct item and escalated. **This is the item most likely to be read as blocking.**
- **`finetuned_chat` is down to one row.** Two of round 2's three were sub-floor exceptions and the
  third is a vision-language model M1 no longer routes. Under a mechanical rule the category simply
  did not produce candidates this week; padding it would be the thing the reviewer stopped.
- **`base_pretrained` is now 25 rows, most of them encoder checkpoints** (`bert-base-cased`,
  `distilroberta-base`, `deberta-v3-large`, `Bio_ClinicalBERT`, `bert-large-portuguese-cased`). Every
  one clears a 1,000,000-download floor and M2 routes it, but a person may well judge that this
  roster is about vendor-marketed base models and reject a third of them. That rejection is the
  editorial step the tail exists to feed, and it is cheap; the alternative was me making the call.
- **The M2 table maps `translation` → `base_pretrained`**, which is how `t5-small` and `t5-base` land
  there. It is inherited unchanged from round 2's published rule and was not flagged, but it is the
  weakest entry in the table: a translation-only model would land in `base_pretrained` too.
- **`meta-llama/synthetic-data-kit` declares a PyPI project page as its homepage**, so the row
  carries `homepage: https://pypi.org/project/synthetic-data-kit/` rather than
  `pypi: synthetic-data-kit`. Turning the one into the other is a URL parse of a declared string —
  mechanical enough, but not a derivation the published table licenses, and adding a rule for it
  mid-round is the thing this round is trying to stop. Worth a rule in `docs/reference/` later.
- **`pymupdf`'s declared homepage carries UTM parameters** (`?utm_source=github&…`). It is the
  repository's declared homepage verbatim, and stripping query parameters would be a transformation
  the derivation table does not license. It looks like a mistake and is not one.
- **24 rows carry a homepage, which is what moved the `homepage_domain` denominator from 27 to 50.**
  That is the single cause of both the coverage-baseline drift and the sixth test failure.
- **The boundary lines in the pass-B parks are mine.** 66 candidates are parked on them. They can
  only hold a real product back for a week, and the sheet names each line explicitly so a reviewer
  can disagree cheaply, but a reviewer who disagrees with a line disagrees with ~20 rows at once.
- **`docs/sweeps/` is still a directory this batch invented**, unchanged since round 1 and never
  raised. A maintainer who wants the sheet elsewhere pays one file move plus six header edits.
- **The sheet is 565 KB / 3,275 lines**, and the three provenance tables are 2,830 of them. Nobody
  reads that end to end; it exists to be grepped and to make the reconciliation recomputable.
- **The warehouse discovery pool was still unreachable** (no `OSO_API_KEY`), so multi-signal
  consolidation still rests on self-dedup. Settled as non-blocking in round 1.
- **Pass-A raw pulls were reused, pass-B pulls are fresh.** Every fetch date in the sheet is honest
  (both passes are 2026-09-07), but the pass-A signals were not re-checked against the live APIs
  today.

## Response to round N-1

|---|---|---|---|---|
| `docs/sweeps/2026-09-07-tail-batch.md:342` — identity fields are LLM-authored assignments, not observations; ADR-004 point 4 covers every tail field | ✅ | | | Round 2's pushback is **withdrawn**; the reviewer's example settles it. Every field on every one of the 67 rows is now produced by one published derivation: `slug` = `slugify(declared name)` (fallback `slugify(owner)-slugify(name)`, then park — `pytorch/xla` is the only row that takes the fallback, because head product `xla` holds the slug); `display_name` = the declared name verbatim; `org` = the organization `sources/org_handles.yaml` already declares for that `(platform, owner login)`, else `slugify(owner login)`. 14 of the 43 carried-over rows had a hand-made organization mapping and all 14 are gone — `answer-ai` → `answerdotai`, `bespoke-labs` → `bespokelabsai`, `google` → `google-bert`, `meta` → `facebookai`, `openai` → `openai-community`, `hugging-face` → `distilbert`, `pytorch-foundation` → `pytorch` and seven more; `data-prep-kit` → `ibm` went with its row. The three product-family folds are **un-folded**: `FacebookAI/roberta-large`, `google-t5/t5-base` and `openai-community/gpt2-large` are each resolved on their own and emitted as `roberta-large`, `t5-base`, `gpt2-large`. 36 of the 43 rows carried over from round 2 changed a field as a result. The word `transcription` appears zero times in the new sheet (`grep -c` → 0); every cell of every per-row derivation table now reads `slugify(declared name)`, `declared name, verbatim`, `slugify(declared owner login)` or `declared handle … (sources/org_handles.yaml)`. Round 2's counter-argument is also retracted on the facts: `build/validate.py`'s version-token check builds `model_products` from category rosters, so it never applies to a registry row — verified, and `uv run python -m build.validate` prints `0 error(s)` with `roberta-base`, `t5-small` and `gpt2-large` in the file. The promotion-time consequence (a promoter must collapse these to a product line and record it in `model_families.yaml` or `version_in_identity`) is escalation 1 in the sheet, which is where ADR-004 point 4 says it belongs. |
| `docs/sweeps/2026-09-07-tail-batch.md:184` — eight sub-floor candidates accepted because the sweep "recognized" them; recognition is not a reproducible predicate | ✅ | | | All nine sub-floor accepts are parked with the floor that excluded them and listed by name in the sheet: `data-prep-kit/data-prep-kit` (958 stars), `MinishLab/semhash` (963), `nolabs-ai/deepfabric` (885), `magpie-align/magpie` (881, push 2025-03-17), `datadreamer-dev/DataDreamer` (push 2025-02-02), `BatsResearch/bonito` (829), `LiquidAI/LFM2.5-2.6B` (120,565 downloads), `EleutherAI/gpt-j-6b` (261,799), `LGAI-EXAONE/EXAONE-3.5-7.8B-Instruct` (373,508). The Magpie sentence the review quoted is gone with the row. Acceptance is now the predicate `in_cutoff and survived_dedup and derivable_category and addressable_identifier and no_recorded_park_reason`, derived rather than listed, so no code path can admit a sub-floor candidate; the words "recognized", "widely used" and "notable" appear nowhere in the acceptance path. Replenishment is a second documented retrieval pass at the **same** cutoffs — 16 queries predeclared in a manifest (query, declared category, floor, sort, depth) before the first request, all fetched 2026-09-07, published in full in the sheet — not a lowered floor. The one manifest change is disclosed and tightens: the four pass-B Hugging Face listings move from 5,000 to the 1,000,000 floor pass A gives the same query shape, which excludes signals the manifest had admitted and admits none it excluded. Result: 2,579 raw = 1,039 duplicates + 1,540 unique; 1,540 unique = 67 accepted + 1,473 parked (both equations asserted by the reconciliation script, which raises on either imbalance). All 1,473 parks and all 1,039 duplicates are listed individually, and 288 of the parks carry a reason of their own. |

ccerv1:
## What I did
- files touched:
  - `docs/sweeps/2026-09-07-tail-batch.md` — the only file changed this round. Committed as `ad6c7f6f` on `loop/gap-map-candidates-2026-09-07-a1`.
  - No registry file changed. `sources/registry/{base_pretrained,compilers,dataset_processing_tools,finetuned_chat,safeguards}.yaml` are byte-identical to round 3; the 67 emitted rows are untouched.
  - Scratch analysis scripts (not in the repo): `/tmp/claude-1000/-workspace-Obsidian/931860aa-e5e5-41f0-bfc6-eef224455303/scratchpad/{extract,audit,audit2,build_park,mkpark,splice,verify}.py`.
- commands run:
  - `uv run --no-sync python -m build.validate` → `0 error(s), 2 warning(s)` (both warnings are the pre-existing `model_families` pattern-overlap notices, byte-identical on `main`).
  - `uv run --no-sync python -m build.check_corpus_diff --base main` → empty review sheet, `stage moves: none`, `untouched-product row changes: none`.
  - `uv run --no-sync pytest tests/test_identity_eval.py -q` → `6 failed, 131 passed`, the same six round 3 documented and escalated under ADR-004 point 3. Nothing changed there: the registry rows that cause them are untouched.
  - `git diff --stat main` → 6 files, all under `sources/registry/**` plus this sheet.
  - A programmatic recount of all four tables in the sheet (`verify.py`), read back off the committed file rather than carried forward.

## Why
- decisions:
  - **The blocking item was a symptom of a class, so I audited the class.** The reviewer flagged two `llm` folds and named 18. I extracted all 1039 duplicate rows and tested every fold against the repo's own files: `sources/products/*.yaml` declared artifacts, `sources/registry/*.yaml` rows, `sources/resolution_ledger.yaml`, `sources/model_families.yaml` and `sources/org_handles.yaml`. 300 folds — not 18 — rested on nothing declared.
  - **The rule I wrote down is the one already in the repo.** `discover-candidates.md` says, of the identity digest, that "an item whose only evidence is a folded name match is `parked`, not shown as a reviewable item". `docs/reference/identity.md` says a shared homepage domain "must never establish equivalence between two candidates or suppress a second one from being proposed". The new **What a fold has to rest on** section applies both to the sweep's own dedup and enumerates the five evidence kinds a kept fold rests on, with a row count per kind that reconciles against the remaining table (463 + 88 + 7 + 175 + 1 + 5 = 739).
  - **A fold that a *kept* fold rests on:** `release or SKU of P` survives only where `model_families.yaml` declares the family `P-*` **and** the signal's owner login is a handle `org_handles.yaml` declares for an organization owning P's declared artifacts. 175 rows pass (`Qwen/Qwen3-32B`, `meta-llama/Llama-2-7b`, `zai-org/GLM-4.5`); 286 fail.
  - **Withdrawn, not rejected.** Each of the 300 becomes a unique candidate carrying a park reason, which the acceptance predicate reads through `no_recorded_park_reason` — its one judgment-shaped clause, and the one that can only remove. So this is a dedup rule, not a legitimacy rule: it cannot admit a row and cannot reject a product, which is why `accepted` stays at 67 and no registry file moved.
  - **Five withdrawal classes, stated once, cited per row:** N1 third-party derivative inside a declared family (181), N2 first-party with no declared family bridge (37), N3 name token only (68), N4 vendor domain plus path (2), N5 a stated reading of what a repo *is* (12). Each of the 300 rows carries its class, its signal, source URL, fetch date, the queries that returned it, the fold that was withdrawn and why.
  - **Counts recomputed from the tables, not adjusted.** 2579 = 739 + 1840; 1840 = 67 + 1773; parked = 288 reasoned + 1185 cutoff + 300 withdrawn. I recounted all four tables off the committed file to confirm rather than asserting the arithmetic.
  - **Two escalations added** (items 11 and 12) so the parks are actionable on `main`: seven first-party families with no `<product>-*` bridge declared (`granite`, `granite-guardian`, `smollm`, `minicpm`, `pythia`, `zephyr`, `apertus` — 37 signals), and 181 third-party derivatives that need a `base_model` read or a person's ruling.
- alternatives rejected:
  - **Fixing only the 18 `llm` folds.** It would have satisfied the literal item and left `pytorch-playground` folded onto `pytorch`, `Triton-guide` onto `triton`, `cursor-security-rules` onto `cursor` and 279 more — the same defect, one round later.
  - **Supplying the missing evidence instead of parking.** For 37 of the 300 the honest repair is a `model_families.yaml` family bridge, and for the 2 domain folds an `org_handles.yaml` `homepage_domain` handle. Both files are outside a `tail-batch` diff (`discover-candidates`: "Files this changes … Nothing else"), and writing them here would trade one out-of-scope act for another. Parked with the bridge named, escalated for `main`.
  - **Asserting an HF `base_model` field to keep the 181 derivative folds.** This pass never read that field. Recording an unread observation to preserve arithmetic is the error being corrected, in a new place.
  - **Keeping the seven mirror/derivative folds because earlier rounds reasoned about them individually.** They rested on sentences ("mirror under a second owner", "Core ML conversion", "plugin surface"), not declarations. Leaving them would have made the new evidence table false on its own terms.
  - **Re-running the sweep to replenish `accepted` back up after the un-fold.** Nothing was rejected — the 300 are held, not dropped — and a new retrieval pass would change the batch under review instead of correcting it.

## Risks / unsure about
- The 181 N1 rows are very probably not new products — third-party GGUFs, abliterations and merges of first-party checkpoints. Parking them counts them as unique candidates, which slightly *over*states `unique_candidates` in the direction opposite to the blocker. I took that side deliberately: the reviewer's failure mode was understating unique candidates through folds nothing supports, and the correct fold here needs an observation (`base_model`) this pass does not have.
- Three of those 181 (`ibm-granite/granite-guardian-3.3-8b-GGUF`, `ibm-granite/granite-guardian-4.1-8b-GGUF`, `janhq/Jan-v3-4B-base-instruct-gguf`) match the shape of the declared format-redistribution predicate, whose third condition is the repo's own `tags`. Not read for them, so not folded; flagged in escalation 12.
- The parked total is now 1773 across three tables. Every row is individually listed with provenance, but the sheet is 3391 lines. ADR-004's reviewability concern is met by the class tables (six evidence kinds, five withdrawal classes) rather than by row-by-row reading.
- Escalation renumbering shifted the old items 11 → 13. Nothing else in the sheet cross-references escalations by number, but a reader holding the round-3 sheet will see the shift.
- `pytest -q` in full was not re-run this round (657s, and the registry rows it grades are untouched); `tests/test_identity_eval.py` was, and returns the same six failures. The sheet now says exactly that rather than implying a full re-run.

## Response to round N-1
|---|---|---|---|---|
| `docs/sweeps/2026-09-07-tail-batch.md:2207` — `mlabonne/llm-datasets` and `poloclub/llm-landscape` classified as releases or SKUs of Datasette's `llm` with no declared artifact or identity mapping; 18 such `llm` folds; audit them, supply evidenced mappings or park with reasons, recompute all five totals. | ✅ | | | Audited the whole fold class, not just `llm`: 300 folds rested on nothing declared (286 name matches incl. all 18 `llm`, 2 vendor-domain matches, 12 stated readings). All 300 un-folded into a new **Parked — a withdrawn fold** table, each with class, signal, source URL, fetch date, returning queries, the withdrawn fold and the reason. New **What a fold has to rest on** section states the kept-fold evidence rule and reconciles per kind (463 repeats + 88 declared artifacts + 7 ledger + 175 first-party declared-family releases + 1 declared-tag redistribution + 5 hardware retries = 739). Three further false folds found and withdrawn: `meta-llama/Prompt-Guard-86M` → `llama-prompt-guard` (which declares Prompt Guard **2**, `meta-llama/Llama-Prompt-Guard-2-86M`), and two `nvidia/NVIDIA-Nemotron-3*` checkpoints → `nemotron` (which declares neither). All five totals recomputed from the tables: `raw` 2579 (unchanged), `duplicates` 1039 → 739, `unique` 1540 → 1840, `accepted` 67 (unchanged), `parked` 1473 → 1773; 739 + 1840 = 2579 ✓, 67 + 1773 = 1840 ✓, recounted off the committed file. Zero `release or SKU of llm` rows remain in the duplicate table. `build.validate` → 0 error(s); `check_corpus_diff --base main` → empty sheet, no stage or gap moves; diff is `sources/registry/**` plus this sheet only. |

PR #514, opened 2026-09-07
The org recall failure told readers to look for a bug that isn't there

ccerv1:
## Summary

`org` had **no `FLOOR_NOTES` entry at all**, so its only failure message was `the resolver missed a pair a declared handle already bridges` — and the module docstring reinforced it, calling a miss *"a bug rather than a curation gap."* For the commonest cause of this failure, that reading is wrong and sends the reader hunting a resolver defect that does not exist.

Part of `org`'s truth is tail-derived. Truth is the repo; the edges are the warehouse. A registry batch adds org pairs the moment it merges, and none can be recovered until the weekly publish lands — so recall drops to `(n-k)/n` with no resolver defect involved. Exactly the publish-lag case `membership_non_scoring` already documents, in a relation whose message denies it.

## Isolated, not inferred

Same warehouse, two repo states:

| repo state | org recall | truth | n_tail | n_emitted |
|---|---|---|---|---|
| pre-#511 (`9df82a12`) | **1.000** | 446/446 | 5 | 3089 |
| post-#511 (`7b4bb5b9`) | **0.980** | 446/455 | 14 | 3089 |

`n_emitted` identical, truth up 9, recovery up 0. `#511` added 9 tail-derived org pairs the warehouse has not published. Lag, not regression.

## What this adds

The note, plus the test a reader can actually apply: **unchanged `n_emitted` against grown `n_truth` is a lag** and clears at the next publish; **a genuine miss shows `n_emitted` moving while recovery does not.** The docstring now names the caveat instead of asserting the opposite.

## Not changed — flagged for a human

The `>= 0.99` invariant, and what counts as org truth. Making this lag-proof means either scoping truth to what has been published, or exempting unpublished tail pairs. Either is a rubric change.

Worth deciding, because an "invariant" a routine merge can trip is not one — and the same mechanism has `membership_non_scoring` recall at `0.529` right now. Both clear at the next weekly publish, so nothing is broken; the gates are just uninformative for the rest of this week.

## Test plan

- [x] `uv run pytest tests/test_identity_eval.py` — 137 passed
- [x] Live run confirms the note prints under the real failure (`--from-warehouse --floors`)
- [x] Lag hypothesis isolated against the pre-#511 tree on the same warehouse
- [ ] CI green

### DIFF
--- a/build/identity_eval.py
+++ b/build/identity_eval.py
@@ -155,7 +155,11 @@ recall read 1.000 and mean nothing:
   computed over exactly the pairs a declared handle could bridge, so recoverability is defined
   by the same route the graph emits on. That makes it a REGRESSION INVARIANT, not a coverage
   target: a miss is the resolver failing to use evidence it already has, which is a bug rather
-  than a curation gap. `RECALL_INVARIANTS` pins it at >= 0.99 for `org`, it still exits 1 on
+  than a curation gap. **One caveat the invariant does not currently model:** part of `org`'s
+  truth is tail-derived, and truth is the repo while the edges are the warehouse -- so a registry
+  batch depresses this recall the moment it merges and until the weekly publish lands, with no
+  resolver defect involved. Until that is modelled, read a failure here against `n_emitted`:
+  unchanged while `n_truth` grew means lag, not regression. `RECALL_INVARIANTS` pins it at >= 0.99 for `org`, it still exits 1 on
   failure under `--floors`, and the table labels that row `recall invariant` rather than
   `recall floor` so nobody reads a 1.000 as coverage.
 - **coverage = have we given it enough evidence.** That is `org_handle_coverage`, per route:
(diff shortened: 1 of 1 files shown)


ccerv1:
Closing: the diagnosis in this PR is wrong.

A Codex review flagged the discriminator as unsound — `unchanged n_emitted with grown n_truth` cannot separate "the inputs never arrived" from "the inputs arrived and the resolver emitted nothing for them." That was correct, and checking properly reversed the conclusion.

The module's own `membership_non_scoring` note already states the sound test — compare `sources/registry/*.yaml` against the published `registry.tail_products` — and I substituted a weaker inference instead of running it. Running it:

```
new #511 slugs in published registry.tail_products:
  agent-governance-toolkit, agentic-radar, agentic-security  -> PRESENT
```

**The publish has landed.** So the 9 new org truth pairs are not lag-blocked; the warehouse has the products and the org pairs are not being emitted for them. That is either a genuine resolver miss — which is what the original failure message said all along — or identity edge tables that have not been rebuilt since `tail_products` was published (`n_emitted` held at 3089 across both repo states).

What still stands from the investigation, and is worth keeping:

- Same warehouse, two repo states: org recall 1.000 (446/446) -> 0.980 (446/455), `n_tail` 5 -> 14, `n_emitted` 3089 both. The cause is repo-side truth growth; that much is isolated.
- `org` genuinely has no `FLOOR_NOTES` entry, so its failure prints no guidance at all. Still worth adding — but the note has to describe the real cause.
- Published `tail_products` returned duplicate slugs and holds 156 rows against 100 in `sources/registry/*.yaml`. Unexplained, and possibly relevant.

A replacement PR should wait until the 9 missed pairs are identified.

PR #519, opened 2026-09-08
Retire the refresh-data workflow, whose fetchers ADR-003 externalized

### LINKED ISSUE #509: sentinel: refresh-data failed
Scheduled run of `refresh-data` failed at 2026-09-07T10:13:40Z.

Run: https://github.com/currentai-org/os-ai-map/actions/runs/34110332795

Read the log, decide whether it is drift, staleness, or a broken upstream, and close this when the next run is green. If the gate is advisory and the finding is real, file the finding as its own issue and link it here.

ccerv1:
Finishes what ADR-003 started. The weekly `refresh-data` workflow ran two fetchers,
`warehouse/models/catalog/model_repos.py` and `model_benchmarks.py`. #406 deleted both when it
externalized the `catalog.*` tables, and left the workflow calling them. It has failed every
Monday since on `[Errno 2] No such file or directory`; the most recent is run 34216673337.

An externalized table has no repo producer by design — `warehouse/audits/externalization.json`
records `catalog.model_repos` and `catalog.model_benchmarks` as `frozen-without-producer`, retained
live under platform ownership. So there is nothing to repoint the fetchers at, and every remaining
step in the workflow existed to serve them. Closes #509.

## What changed

| File | |
|---|---|
| `.github/workflows/refresh-data.yml` | Deleted (215 lines). Its credentials preflight went with it — it was caller-specific and gated nothing else. |
| `.github/workflows/report-failure.yml` | `refresh-data` dropped from the sentinel's watch list. A sentinel on a deleted workflow can never fire, so it silently stops being a gate. |
| `docs/operations/refresh-data.md` | Reframed as the loading half of the runbook, stating plainly that the catalog fetchers are gone and why. |
| `docs/operations/deploy-models.md` | Cross-reference updated. |
| `AGENTS.md` | Same dangling reference, found during the run rather than listed up front. |
| `CHANGELOG.md` | Recorded under Unreleased. |

No changes under `sources/` or `warehouse/models/`. This is retirement and docs only.

## Verification

- `uv run python -m build.validate` → 0 errors, 2 warnings (both pre-existing `model_families`
  pattern-overlap warnings, unrelated to this change)
- `uv run pytest -q` → 1834 passed
- Every `uv run python <path>.py` invocation across all remaining workflows resolves to a file
  that exists, so the bug class behind #509 is gone from the repo rather than from this one
  workflow

## Two questions the review settled

- **Should the credentials preflight survive?** No. It was written for this workflow's two
  fetchers and has no other caller. Keeping it would mean keeping a shell with nothing in it.
- **Does deleting this orphan the `data-refresh/YYYY-MM-DD` branch convention?** No consumer of
  those branches was found.

---

Built and reviewed by the `refresh-data-retirement` build-review loop, run `2026-09-08-a1`, one
round, verdict `pass`. Reviewer note carried forward: it accepted the builder's test figures
because its sandbox blocked independent reruns. Those figures were rerun separately and match.


ccerv1:
## What I did
- files touched:
  - `.github/workflows/refresh-data.yml` — **deleted** (215 lines).
  - `.github/workflows/report-failure.yml` — dropped `refresh-data` from the `workflow_run.workflows:` list; the other nine stay.
  - `docs/operations/refresh-data.md` — rewritten. Retitled "Load external data into the warehouse (maintainer)". Opens by stating plainly that there are no catalog fetchers in this repo any more, names ADR-003, `warehouse/audits/externalization.json` (`frozen-without-producer`), #406 as the deletion and #509 as the workflow retirement, and says the deployed tables are frozen under platform ownership. Keeps steps 2–4 (static-model load, dependent UDMs, re-publish), which are the live half two other places still link to.
  - `docs/operations/deploy-models.md:261` — the cross-reference now describes what the file actually contains ("loading a fetched CSV into the warehouse as a static model"). It resolves; the file still exists.
  - `AGENTS.md:339` — same fix to the maintainer-operations line, which had said "run fetchers and reload static models".
  - `CHANGELOG.md` — one line under a new `### Removed` in `[Unreleased]`, per the repo's own "a new or removed ... workflow" rule.
- commands run:
  - `uv run python -m build.validate` → `0 error(s), 2 warning(s)` (both warnings are the pre-existing `model_families` pattern-overlap notices, unrelated).
  - `uv run pytest -q` → `1834 passed in 689.75s`. Run in the foreground and waited on.
  - `git diff --stat HEAD` → 6 files, +22 / −223, all under `.github/workflows/`, `docs/`, `AGENTS.md`, `CHANGELOG.md`. Nothing under `sources/` or `warehouse/models/`.
  - Committed `efb81d77` on `loop/refresh-data-retirement-2026-09-08-a1`.

## Why
- decisions:
  - **Deleted the workflow rather than trimming it.** Verified the two producers are actually absent: `warehouse/models/catalog/` does not exist in the tree at all. `warehouse/audits/externalization.json` carries `catalog.model_repos` and `catalog.model_benchmarks`, both `disposition: frozen-without-producer`, with `archived_source_sha256` for the two `.py` files. Read all 215 lines. Every step either serves those two fetchers or serves the step that publishes their output: the ~128-line preflight validates `GH_FETCH_TOKEN`/`HF_TOKEN`/`OSO_API_KEY` and gates only the fetch it precedes; `Run fetchers` invokes two missing files; `Open refresh PR` commits `warehouse/data/`, which holds exactly one file, `observations/product_adoption_baseline.parquet`, that no step here writes. No step does live work, so nothing was carried over.
  - **Kept `docs/operations/refresh-data.md` instead of deleting it.** Two live references want its second half: `skills/add-data-source/SKILL.md:46` points a contributor at it for the maintainer load step, and `AGENTS.md` lists it under maintainer operations. Deleting it would have broken the skill reference to fix the `deploy-models.md` one. Rewriting satisfies the success criterion either way ("says plainly that the catalog fetchers were externalized") and keeps both pointers resolving.
  - Fixed the two cross-reference *descriptions*, not just the paths — both said the file was about running fetchers, which is the exact claim the retirement falsifies.
- alternatives rejected:
  - Reducing the workflow to a preflight-only `workflow_dispatch`: the preflight exists to protect a 3.5h fetch that no longer happens, and `GH_FETCH_TOKEN` is only used to publish the fetch result.
  - Deleting `docs/operations/refresh-data.md`: see above.
  - Touching `docs/architecture/data-architecture.md` and ADR-002/003, which mention the deleted fetchers. Those are historical records of decisions, not instructions; editing them is out of scope and would rewrite the receipt.
  - Editing `skills/add-data-source/SKILL.md`, which describes `warehouse/models/<dataset>/<table>.py` as a general *route* for a future fetcher, not a claim that the deleted ones exist.

## Risks / unsure about
- Step 5 of the plan (the other four weekly workflows "ran independently on 2026-09-07") is a claim about GitHub run history I could not verify: the clone has no remote and no network check was made. What I did verify is the static half — `parity` (Mon 06:00), `artifacts` (Mon 07:00), `freshness` (Mon 08:00) and `refetch` (Thu 07:00) each still carry their own cron, are byte-unchanged in the diff, and none of them referenced `refresh-data`. Nothing couples them to it.
- `warehouse/data/observations/product_adoption_baseline.parquet` now has no workflow that commits changes to `warehouse/data/`. That was already true in effect (no fetcher in this workflow wrote it), but if some other process assumed the refresh PR would sweep up incidental changes under that directory, it no longer will. I found no such caller.
- The CHANGELOG entry links #509 as an issue (`/issues/509`). If #509 is in fact a PR the link still redirects, but the wording says workflow-retirement tracking, which matches the goal statement.

## Response to round N-1
|---|---|---|---|---|
| _(round 1 — no prior review)_ | — | — | — | — |

Issue #517, opened 2026-09-08
The PLATFORM MIRROR banner sits on six governed assets


ccerv1:
Seven files under `warehouse/models/signal_*/` open with:

```
────── PLATFORM MIRROR (read-only) ──────
The platform is the source of truth; nothing deploys from this copy,
and editing it here changes nothing.
```

Six of them are **governed assets in `warehouse/assets.yaml`**:

| File | `id` | `role` |
|---|---|---|
| `signal_packages/downloads.sql` | `signal_packages.downloads` | `repo-computation` |
| `signal_packages/downloads_daily.py` | `signal_packages.downloads_daily` | `repo-computation` |
| `signal_packages/product_adoption.sql` | `signal_packages.product_adoption` | `repo-computation` |
| `signal_github/repo_state.py` | `signal_github.repo_state` | `compatibility-shim` |
| `signal_github/product_adoption.sql` | `signal_github.product_adoption` | `compatibility-shim` |
| `signal_huggingface/hub_state.py` | `signal_huggingface.hub_state` | `compatibility-shim` |

All carry `population: gap_map`. ADR-003 reserves the mirror for a category-3 dependency and says
plainly that a mirror "is provenance, not ownership, and confers no governance" — so a governed
asset with `role: repo-computation` should not be wearing one. `signal_packages` in particular is
repo computation by its own declaration, and the banner tells a reader the opposite.

## Why it matters beyond tidiness

The banner is the first thing in the file and it is the only ownership signal visible without
opening two manifests. It currently says "you cannot change this" on code the repository owns. That
is not a theoretical risk: it produced a wrong reading of the board on 2026-09-08, in which #314 and
#358 were both classified as blocked-on-platform when most of their work is ordinary repo work.

An agent or contributor who trusts the banner will decline work it can do, and the gates will not
catch that because nothing compares the banner to the manifests.

## What would fix it

Decide what the banner means and make it derivable rather than hand-written:

1. If these are genuinely repo-owned, the banner is wrong and should go, or be replaced with one
   that names the `role`.
2. If they are genuinely platform-authored, the `assets.yaml` entries are wrong and they belong in
   `dependencies.yaml`.
3. Either way, a gate should assert the banner and the manifests agree, the same way
   `dependency_mirror_provenance_violations` already asserts the mirror block matches the bytes.

Case 1 looks right for `signal_packages` and wants a decision for the three compatibility shims.

ccerv1:
[...]

PR #520, opened 2026-09-08
Stop the repository describing a frozen table as though it were live

ccerv1:
# archived-table-doc-guard — pass

**Goal.** Stop the repository describing a frozen table as though it were live. Closed PR #341 carried a test

## Success criteria
- [ ] a test exists that fails when a doc presents an externalized table as current, and its list of tables is derived from `warehouse/audits/externalization.json` rather than hand-maintained, so a table added to the receipt is covered without editing the test
- [ ] the test passes, which means every mention it finds has either been marked or been fixed
- [ ] `docs/reference/queries.md` no longer hands a reader a copy-paste query against a frozen table without telling them the data does not advance
- [ ] `docs/architecture/data-architecture.md` no longer instructs anyone to classify and migrate `currentai.catalog.stack_map`, which ADR-003 already did
- [ ] `uv run pytest -q` passes
- [ ] no change under `sources/` or `warehouse/models/`; docs and tests only

## Outcome
`pass` after 1 round(s). Branch `loop/archived-table-doc-guard-2026-09-08-a1`.

## Final verdict
```json
{
 "verdict": "pass",
 "blocking": [],
 "non_blocking": [],
 "settled_now": [
  "No P0 blockers found; the documentation fixes communicate actual staleness.",
  "Receipt additions should gate immediately; no grace mechanism is needed.",
  "Fence-level warnings are acceptable at P0, but do not fail closed when readers copy partial fences.",
  "This guard would not generally catch #509 or #517: deleted producers and contradictory governed-asset banners require other checks.",
  "Validation relies on the builder\u2019s reported passing runs; independent reruns were blocked by the read-only environment."
 ],
 "escalation": null
}
```

## Settled ground this run added (9)
- ADR-003 and the externalization receipt as the authority for what is frozen.
- That #341 is closed and correctly so; only its test is being kept.
- Docs and tests only. `sources/` and `warehouse/models/` are out of scope.
- That a frozen table's query examples should stay rather than be deleted — they still execute, and
- No P0 blockers found; the documentation fixes communicate actual staleness.
- Receipt additions should gate immediately; no grace mechanism is needed.
- Fence-level warnings are acceptable at P0, but do not fail closed when readers copy partial fences.
- This guard would not generally catch #509 or #517: deleted producers and contradictory governed-asset banners require other checks.
- Validation relies on the builder’s reported passing runs; independent reruns were blocked by the read-only environment.

Later runs of this job inherit `/workspace/GitHub/personal/agent-loops/jobs/settled/archived-table-doc-guard.json`; prune it there.

### DIFF
--- a/docs/architecture/data-architecture.md
+++ b/docs/architecture/data-architecture.md
@@ -410,7 +410,7 @@ The following do not belong in `catalog` long term:
 - repo-derived bridges presented as external inventory;
 - frozen historical tables without an explicit historical label.
 
-`currentai.catalog.stack_map` must be classified and migrated. Its curated identity and membership fields belong in `registry`; any compatibility output required by `scores.stack_contributors` should become an explicitly named compatibility view with a documented deprecation plan.
+ADR-003 settled `currentai.catalog.stack_map`, which this section previously said must still be classified and migrated. It was externalized: removed from this repo's inventory, its producer archived, and the deployed table left frozen under platform ownership at its last publish. Its only reader, `scores.stack_contributors`, was externalized with it, so the compatibility view this section once called for has nothing left to serve. The curated identity and membership fields it carried live in `registry`. Nothing in this repo migrates it now, and a doc that asks someone to is asking them to redo ADR-003.
 
 ### 4.3 `currentai.observations`
 
--- a/docs/reference/queries.md
+++ b/docs/reference/queries.md
@@ -23,27 +23,40 @@ down that read them are historical.
 
 `oso.*` tables are public and can be queried with any valid key.
 
-## Starting points
+## Worked examples (externalized datasets)
 
-**For notebook queries:** use `scores.repos_summary` (a pre-computed snapshot):
+Every example below reads a dataset ADR-003 externalized. They still run. What they return
+stopped advancing at that table's last publish, so a result set that looks current is not,
+and nothing in the query itself says so. Each example carries the warning as a SQL comment,
+which is the only part of this page that survives a copy into your client.
+
+For a table this repo still maintains, start at `registry`, `observations` or `evaluation`.
+
+**Repo-level snapshot,** `scores.repos_summary`:
 ```sql
+-- FROZEN (ADR-003): scores.repos_summary has no producer in this repo and stopped
+-- advancing at its last publish. Results are historical.
 SELECT * FROM currentai.scores.repos_summary WHERE country = 'France' ORDER BY stars DESC
 ```
 
-**For project-level data:** use `scores.project_summary`:
+**Project-level data,** `scores.project_summary`:
 ```sql
+-- FROZEN (ADR-003): historical, no longer refreshed.
 SELECT * FROM currentai.scores.project_summary ORDER BY total_stars DESC LIMIT 20
 ```
 
-**For time-series:** use `metrics.daily` (long format):
+**Time-series,** `metrics.daily` (long format). The series ends at the freeze date, so do not
+read its last point as today:
 ```sql
+-- FROZEN (ADR-003): the series stops at the last publish, it does not run to today.
 SELECT day, value FROM currentai.metrics.daily
 WHERE repo = 'pytorch/pytorch' AND metric = 'stars'
 ORDER BY day
 ```
 
-**For raw events:** use `events.github_events`:
+**Raw events,** `events.github_events`:
 ```sql
+-- FROZEN (ADR-003): no events after the last publish, so counts are a closed window.
 SELECT event_type, COUNT(*) FROM currentai.events.github_events
 WHERE repo = 'pytorch/pytorch' GROUP BY event_type
 ```
@@ -58,7 +71,13 @@ WHERE repo = 'pytorch/pytorch' GROUP BY event_type
 
 ## Join patterns
 
+The `entities`, `metrics` and `scores` joins below all read externalized tables. They are
+kept because the join shapes are still the right ones, not because the data is current.
+
 ```sql
+-- FROZEN (ADR-003): every entities/metrics/scores table in this block is frozen at its
+-- last publish. The join shapes hold; the rows do not advance.
+
 -- Repo → project
 SELECT r.repo, r.project_slug, p.display_name, p.location
 FROM currentai.entities.repos r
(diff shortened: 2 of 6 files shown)


ccerv1:
## What I did
- files touched:
  - `tests/test_externalized_table_docs.py` (new) — the guard, lifted from closed PR #341 and rebuilt
    on the externalization receipt. Table list comes from `build.assets.still_externalized()`, which
    reads `warehouse/audits/externalization.json` and subtracts the `reclaims` events. 27 tables today
    (28 assets minus the reclaimed `signal_goodailist.repo_catalog`); one parametrized case per table,
    plus a guard-on-the-guard test asserting the list is receipt-derived and reclaim-aware, plus a
    self-test that the marker logic actually rejects an unmarked mention in both copy units.
  - `docs/reference/queries.md` — "Starting points" renamed "Worked examples (externalized datasets)"
    with a lead paragraph saying the examples still run and the data stopped advancing; every SQL fence
    now opens with a `-- FROZEN (ADR-003)` comment naming what is stale about that particular result
    (a closed event window, a series that ends at the freeze date, a snapshot that is not current).
    Two dedupe/coverage bullets marked in prose.
  - `docs/architecture/data-architecture.md` — the `currentai.catalog.stack_map` "must be classified and
    migrated" sentence replaced with what ADR-003 actually did to it.
  - `docs/reference/where-scores-live.md` — `catalog.osai_gap_map` row's State cell now leads with
    **Externalized and frozen (ADR-003)**, matching its sibling `catalog.stack_map` row.
  - `docs/sweeps/2026-09-07-tail-batch.md`, `docs/workflows/discover-candidates.md` — the two
    `currentai.entities.repos` mentions marked; the workflow one also tells a sweeper that a miss means
    a stale pool rather than an empty world, which is the actual operational consequence.
  - nothing under `sources/` or `warehouse/models/`; `git status` shows five docs and one new test.
- commands run:
  - `gh pr diff 341` (in the real repo — the branch is deleted, the diff is the only copy)
  - `uv run pytest -q tests/test_externalized_table_docs.py` — 10 failing tables / 14 unmarked mentions
    before the doc fixes, 29 passed after
  - `uv run pytest -q tests/test_externalized_table_docs.py tests/test_docs_integrity.py` — 44 passed
  - `uv run python -m build.validate` — **0 errors, 2 warnings** (both pre-existing `model_families`
    pattern-overlap warnings, untouched by this change)
  - `uv run pytest -q` — full suite, **1869 passed in 719s**

## Why
- decisions:
  - **The unit of the rule is the unit of copying.** This is the granularity call the job asked me to
    make and defend. A mention inside a fenced code block is satisfied by a marker anywhere in that
    fence *at or above* the mention's line. A mention in prose is satisfied only by a marker on the same
    physical line. The reasoning: the harm here is a reader pasting a query that returns plausible stale
    rows, and what a reader pastes is a fence, so a marker inside the fence is the only one that travels
    with the query into their SQL client. Prose is read rather than copied and the eye lands on a line,
    so the line is the unit there.
  - **Rejected: #341's strict same-physical-line rule.** It cannot be satisfied inside SQL without
    wedging `/* frozen */` between `FROM` and the table name. A warning that mangles the query is a
    warning someone deletes.
  - **Rejected: nearest-heading or n-lines-of-context.** `queries.md` already had a correct
    externalization preamble fifteen lines above the first example, and the examples were still
    misleading — a reader arriving from the table of contents never sees it. A rule that a distant
    paragraph can satisfy is a rule any distant paragraph can fool, and this file was the proof.
  - **A marker below the mention does not count**, and there is a test for it. A warning printed after
    the thing it warns about is not a warning.
  - **Reclaimed tables are out of scope.** `still_externalized()` already subtracts the `reclaims`
    events, so `signal_goodailist.repo_catalog` — reclaimed 2026-09-04 as an identity-graph dependency —
    is excluded. Describing it as current is correct now, and the guard-on-the-guard test asserts the
    subtraction rather than trusting it.
  - **`where-scores-live.md:293` is not a false positive.** Its section heading says "do not read these
    as current", but the row's State cell — the column a reader consults to decide whether to query the
    thing — said only "another organisation's taxonomy and scale". That is a reason not to *compare*
    with it, not a statement that it is frozen; both are true and only one was written down. Marked.
  - **No exclusion list for dated sweep logs.** `docs/sweeps/2026-09-07-tail-batch.md` is a historical
    record, so an exclusion was tempting, but a hand-maintained exclusion list is exactly the
    hand-maintained thing the success criteria push against. Adding a parenthetical fact to a log entry
    does not falsify the log.
  - Marker vocabulary is #341's six plus the receipt's own three (`externaliz`, `no producer`,
    `out of scope`), as the plan specified.
- alternatives rejected:
  - Deleting the frozen examples from `queries.md` outright. The join shapes are still the right shapes
    and the tables still answer queries; the defect was silence about staleness, not the examples.
  - Matching on the receipt's bare `dataset.table` ids as well as `currentai.<id>`. It would widen the
    net past the qualified names docs actually use and start flagging prose like "the `scores` dataset".
  - Requiring a marker per distinct table inside a fence. The join-patterns fence touches six tables;
    six comments would bury the queries. One header comment plus per-sub-query notes reads better and
    the at-or-above ordering rule keeps it honest.

## Risks / unsure about
- The fence rule is deliberately more permissive than per-line, so a single `-- FROZEN` at the top of a
  long fence covers everything below it. A reader copying only the third sub-query out of a 30-line
  fence could miss it. I mitigated in the doc (each sub-query in the join block carries its own note)
  rather than in the rule, on the grounds that a rule forcing a comment on every line would get worked
  around. A reviewer could reasonably want the rule tightened to per-sub-query instead.
- Marker matching is substring and case-insensitive, so a sentence containing "frozen" for an unrelated
  reason on the same line as a frozen-table mention would pass. I judged the false-negative cheap
  relative to the false-positive noise of anything cleverer.
- The full-suite run (1869 passed) was launched before a last cosmetic pass over `queries.md` headings
  and bold labels. I re-ran the three tests that read docs (`test_externalized_table_docs`,
  `test_docs_integrity`, `test_render`) and `build.validate` afterwards, all green, but the 1869 figure
  is from the commit's content minus those prose-only edits.
- The job's headline figure was 18 unmarked mentions across 11 tables measured 2026-09-08; my scan finds
  14 across 10 under this rule and this doc surface. The gap is most likely the reclaimed
  `signal_goodailist.repo_catalog` (one mention in `current-state-dag.md`, correctly excluded here) plus
  marker-vocabulary or glob differences in whoever measured. Flagging rather than reverse-engineering
  the difference: every mention the receipt-derived rule can see is now marked or fixed.

## Response to round N-1
|---|---|---|---|---|
| n/a — round 1 | | | | |

PR #524, opened 2026-09-09
Add an end_of_life field, and declare perspective-api ending

ccerv1:
# end-of-life-field — pass

**Goal.** Give the schema a way to say a product is live but ending, and use it on `perspective-api`. Closes

## Success criteria
- [ ] `docs/schemas/product.schema.json` accepts an optional `end_of_life` date. The schema sets `additionalProperties: false`, so this is a deliberate addition rather than a free-form key
- [ ] `sources/products/perspective-api.yaml` declares it, with the date the service stops and a source for that date
- [ ] `uv run python -m build.validate` reports 0 errors with the field present
- [ ] the record a reader sees says the product is ending — in the serialized payload, not only in the YAML, since the payload is what the front end renders
- [ ] `docs/reference/identity.md` states what the field means and what it does and does not change, so the next product that ends is handled the same way
- [ ] `uv run pytest -q` passes, with a test that an `end_of_life` product still satisfies the roster and organization gates
- [ ] `sources/`, `docs/` and `build/` only; nothing under `warehouse/`

## Outcome
`pass` after 2 round(s). Branch `loop/end-of-life-field-2026-09-08-a1`.

## Final verdict
```json
{
 "verdict": "pass",
 "blocking": [],
 "non_blocking": [],
 "settled_now": [
  "Both round-one blockers are resolved: AGENTS.md is absent from the full diff, and post-expiry curation policy is explicitly deferred.",
  "The data-side change satisfies the agreed scope. Rendering end_of_life.date and its source in aipotluck.org remains held.",
  "Full-suite verification remains builder-reported: 1881 passed, 1 skipped."
 ],
 "escalation": null
}
```

## Settled ground this run added (11)
- That the field is the chosen shape. The withdrawals subsystem is preserved on #522 and is not
- That the product, its score record and the Jigsaw organization all stay. Keeping them is the
- That leaving it scored until December was considered and rejected, because it defers the same
- `sources/`, `docs/` and `build/` only.
- The sourced object and serialized date/source satisfy the data-side requirement; generated payload files should remain uncommitted.
- Declaring the field leaves scores, counts, stages, and gaps unchanged, including for fully open products.
- The aipotluck.org counterpart must render end_of_life.date and its source; that consumer work remains held.
- Independent validation returned 0 errors and 2 warnings. Test execution was blocked by the read-only environment; the full-suite result remains builder-reported.
- Both round-one blockers are resolved: AGENTS.md is absent from the full diff, and post-expiry curation policy is explicitly deferred.
- The data-side change satisfies the agreed scope. Rendering end_of_life.date and its source in aipotluck.org remains held.
- Full-suite verification remains builder-reported: 1881 passed, 1 skipped.

Later runs of this job inherit `/workspace/GitHub/personal/agent-loops/jobs/settled/end-of-life-field.json`; prune it there.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -146,8 +146,10 @@ registry, plus the single `sources/taxonomy.yaml` manifest:
   business logic. Open artifacts are declared as typed top-level arrays of `{url: ...}`
   objects: `github`, `npm`, `pypi`, `crates`, `go`, `huggingface_model`,
   `huggingface_dataset`, `arxiv`. Only keys with entries are included; `product.schema.json`
-  is the authoritative list. Four further optional keys are not artifacts and are documented
-  in the schema: `aliases`, `lineage`, `version_in_identity`, `artifact_exceptions`. Optional
+  is the authoritative list. Five further optional keys are not artifacts and are documented
+  in the schema: `aliases`, `lineage`, `version_in_identity`, `artifact_exceptions`,
+  `end_of_life` (a declared sunset date for a product that still works; it moves no score and
+  no category stage -- docs/reference/identity.md). Optional
   `comments` is a free-text string for provenance and scoring notes (version, license, last
   release date).
 - **scores**: one file per product (same slug) with `openness`, `adoption`, `capability`.
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -146,10 +146,8 @@ registry, plus the single `sources/taxonomy.yaml` manifest:
   business logic. Open artifacts are declared as typed top-level arrays of `{url: ...}`
   objects: `github`, `npm`, `pypi`, `crates`, `go`, `huggingface_model`,
   `huggingface_dataset`, `arxiv`. Only keys with entries are included; `product.schema.json`
-  is the authoritative list. Five further optional keys are not artifacts and are documented
-  in the schema: `aliases`, `lineage`, `version_in_identity`, `artifact_exceptions`,
-  `end_of_life` (a declared sunset date for a product that still works; it moves no score and
-  no category stage -- docs/reference/identity.md). Optional
+  is the authoritative list. Four further optional keys are not artifacts and are documented
+  in the schema: `aliases`, `lineage`, `version_in_identity`, `artifact_exceptions`. Optional
   `comments` is a free-text string for provenance and scoring notes (version, license, last
   release date).
 - **scores**: one file per product (same slug) with `openness`, `adoption`, `capability`.
--- a/build/check_payload.py
+++ b/build/check_payload.py
@@ -88,6 +88,30 @@ def _require_dict(payload: dict, key: str) -> dict:
     return value
 
 
+def _check_end_of_life(slug: str, eol: object) -> None:
+    """Gate the shape of a declared end-of-life record.
+
+    The field says a product still works and stops on a stated day, and the payload is where a
+    reader meets that claim. A row carrying `end_of_life` that the app cannot read -- a bare
+    string where the object belongs, a date in some other notation, a source that is not a
+    link -- would render as nothing at all, which is the one outcome worse than never declaring
+    it: the map would look like it had checked and found no sunset. So an unreadable record
+    fails here rather than being skipped.
+    """
+    if not isinstance(eol, dict):
+        raise PayloadError(
+            f"{slug!r} has an end_of_life that is not an object (got {type(eol).__name__})"
+        )
+    date = eol.get("date")
+    if not is_iso_date(date):
+        raise PayloadError(f"{slug!r} has an end_of_life date that is not ISO 8601: {date!r}")
+    source = eol.get("source")
+    if not isinstance(source, str) or not source.startswith(("http://", "https://")):
+        raise PayloadError(
+            f"{slug!r} declares an end_of_life with no usable source URL: {source!r}"
+        )
+
+
 def check(payload: dict) -> None:
     if not isinstance(payload, dict):
         raise PayloadError(f"payload is not an object (got {type(payload).__name__})")
@@ -138,6 +162,8 @@ def check(payload: dict) -> None:
                 or fresh.get("basis") not in _FRESHNESS_BASES):
             raise PayloadError(f"{slug!r} has no usable freshness record")
         _check_freshness_caveats(slug, fresh)
+        if "end_of_life" in row:
+            _check_end_of_life(slug, row["end_of_life"])
 
     dates = {row["freshness"]["date"] for row in rows}
     if len(dates) <= 1:
--- a/build/serialize.py
+++ b/build/serialize.py
@@ -18,7 +18,7 @@ ROOT = Path(__file__).resolve().parents[1]
 
 PRODUCT_KEY_ORDER = ["slug", "product", "org_slug", "org", "type", "description",
                      "openness", "adoption", "capability", "overall_score", "tier",
-                     "maturity", "mature",
+                     "maturity", "mature", "end_of_life",
                      "freshness", "version_note", "lineage"]
 
 # --- Gap analysis (category-level stage + gaps) -----------------------------
@@ -320,6 +320,20 @@ def _row(slug: str, prod: dict, org_slug: str, org_name: str, score: dict,
     row["mature"] = (
         m is not None and m >= _MATURE_MIN and openness["bucket"] == "open"
     )
+    # A declared end-of-life date: the product still works and stops on that day. Emitted as
+    # the date and the announcement link, so a consumer can say the product is ending and point
+    # at who said so; the `shows` quote stays in sources/ because the payload is not the audit
+    # trail. It is deliberately inert in every number above it -- the openness, adoption and
+    # capability blocks, `overall_score`, `tier`, `mature` and the category stage all read the
+    # same as they did before the field was declared. A date that moved a score would move a
+    # category's stage with no evidence change and no curator acting; where a sunset has really
+    # moved usage, that belongs in the adoption LEVEL with a source. Nothing here compares the
+    # date to today either: the payload carries what was declared, and whether that day has
+    # passed is the reader's arithmetic, not a value that silently changes overnight.
+    # See docs/reference/identity.md.
+    eol = prod.get("end_of_life")
+    if eol:
+        row["end_of_life"] = {"date": eol["date"], "source": eol["source"]}
     # Bridge: the source field is now `comments` (a string), but the payload key
     # the notebook consumes is still `version_note`. Same value, renamed at rest.
     if prod.get("comments"):
--- a/docs/reference/identity.md
+++ b/docs/reference/identity.md
@@ -124,13 +124,15 @@ that stops in three months, and for that case the argument is:
 - The reader is better served by the fact than by an arithmetic gesture. "This category's leading
   open option stops in December" is a sentence the map can print. A silently lowered stage is not.
 
-**Nothing happens when the date passes.** No automatic retirement, no change in the payload, no
-gate that turns red overnight; an ending product does not become a different kind of record by
-being left alone. The date passing is a curation trigger. On or after it, a curator re-reads the
-source and either keeps the record with a past date, which is honest and still linkable, or
-retires the slug through the alias mechanism if something has actually replaced it. If a sweep is
-ever wanted for dates that have gone by, it belongs with the verification queue, alongside the
-other work that a curator picks up, and not in a build gate.
+**Nothing happens when the date passes.** Nothing in the build compares `end_of_life.date` to
+today's date. `build/serialize.py` copies the declared date onto the product row, the gates read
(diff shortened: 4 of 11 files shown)


ccerv1:
## What I did

Commit `0a5014e7` on `loop/end-of-life-field-2026-09-08-a1`, message
`feat(schema): add end_of_life, and declare perspective-api ending`. 11 files, +328 / -4.

- files touched:
  - `docs/schemas/product.schema.json` — new optional `end_of_life` object, slotted after
    `comments`. Required `date` (ISO, pattern-checked), `source` (http(s)), `shows` (the quote);
    optional `accessed`. `additionalProperties: false` inside it too, so the block cannot grow a
    free-form key either. Descriptions written to the file's house register (what the field is,
    what reads it, what it does not do).
  - `sources/products/perspective-api.yaml` — declares it: `2026-12-31`, `https://perspectiveapi.com/`,
    the sunset sentence quoted from the page, accessed `2026-08-13`. The quote and the accessed
    date are the ones already carried by the score file's verified source for that URL
    (`content_sha256: c40c721c…`), so nothing new was asserted about a page I did not fetch.
  - `build/serialize.py` — `end_of_life` added to `PRODUCT_KEY_ORDER`; `_row` emits
    `{date, source}` when the product declares one. The `shows` quote is deliberately not emitted.
  - `build/check_payload.py` — `_check_end_of_life` gates the shape of the emitted record
    (object, ISO date, http(s) source) and fails rather than skipping what it cannot read.
  - `docs/reference/identity.md` — new normative section "A product that is ending", between the
    alias rules and the multi-release combine rules, plus the field in the Related list.
  - `docs/workflows/update-product.md` — a fourth route in the classify-first flowchart
    ("Announced an end date, still works") and its section, so the next curator hits it before
    reaching for retirement.
  - `AGENTS.md` — the "four further optional keys" sentence was falsified by this change; now
    five, with a one-clause statement of what the field does not move.
  - `tests/test_validate.py` — the criterion test: an `end_of_life` product still satisfies the
    roster and organization gates (`validate_sources(...) == []`), plus a missing `source` is a
    schema error, plus a bare date (string and `datetime.date`) is refused.
  - `tests/test_serialize.py` — the field reaches the payload with its source and without the
    quote; a product without it carries no key; declaring it moves no score, no tier, no
    `mature`, no stage and no gaps (asserted on the whole row minus the new key, so a future
    penalty anywhere on the row fails here); a date already in the past serializes identically,
    which pins "declared, not derived from today".
  - `tests/test_check_payload.py` — the four gate branches.
  - `tests/test_signal_routing.py` — `end_of_life` added to `METADATA_KEYS` with a reason. This
    was a real failure, not a formality: the inverted check fails any new product key by default,
    and the reason it belongs in metadata is that the day a service stops is announced on a page
    and never emitted by a registry, so there is no source it could ever route to.

- commands run:
  - `uv run python -m build.validate` → `0 error(s), 2 warning(s)` (both warnings pre-existing,
    the `model_families` pattern overlaps).
  - `uv run pytest -q` → `1881 passed, 1 skipped in 862.26s`, after the `test_signal_routing`
    fix. The run before it caught that one failure.
  - `uv run python -m build.serialize` then `uv run python -m build.check_payload` →
    `616 products, 348 organizations, ok`; payload diff was the four lines of the new block on
    `perspective-api` and nothing else. `uv run python -m build.check_retirement` →
    `0 retired, all routed`.
  - Inertness check, run before writing the rule down:
    `_stage_and_gaps(rows, weights)` on the real `safeguards` rows returns
    `{'num': 3, 'name': 'Viable Alternatives', 'gaps': ['adoption']}` both with and without the
    `perspective-api` row; the product's blended maturity is 3.5 (adoption 4, capability 3, even
    weights), so it does not set `mature_anywhere` either.

## Why

- decisions:
  - **The field is an object, not a bare date.** The success criterion asks for both a date and a
    source for it, and the product record has nowhere else to put a citation (`comments` is prose
    the front end renders, not an audit trail). Requiring `shows` follows the score-source shape:
    a page saying support ends and a page saying the service is switched off are different dates,
    and a paraphrase loses the difference. A test pins that `end_of_life: 2026-12-31` — the
    shortcut a curator reaches for — is refused, in both its string and `datetime.date` spellings.
  - **The rule: it changes nothing that is computed.** Identity is untouched (slug, both rosters,
    the organization, the scores), and no number moves — openness, adoption, capability,
    `overall_score`, `tier`, `mature`, the category's stage and its gaps. Three reasons that do
    not depend on this first case: a stage is a count of what exists now, so excluding an ending
    product would drop a category a rung on a calendar tick with no evidence change and no human
    act, and this repo's discipline is that promotion is a human act — demotion must not be
    cheaper. Where a sunset has genuinely moved usage, that belongs in the adoption level with a
    source a reader can audit; a schema-applied penalty would double-count with an honest
    re-score and be invisible in the numbers it changed. And the reader is better served by the
    printed fact ("this category's leading open option stops in December") than by a silently
    lowered stage.
  - **Nothing fires when the date passes.** No auto-retirement, no payload change, no gate that
    turns red overnight. The date passing is a curation trigger: re-read the source, then keep the
    record with a past date or retire the slug if something actually replaced it. Documented in
    identity.md, and pinned by the test that a past date serializes like any other.
  - **Verified the inertness claim rather than trusting it** (numbers above), and wrote the
    section so the rule stands on the general argument with the `perspective-api` result as a
    corroborating footnote, not as the reason.
  - **`build/notebook_data.json` is not in the commit.** AGENTS.md line 277 and four workflow docs
    say it is bot-owned, preview only, and that CI blocks hand edits. I regenerated it to verify
    the payload and reverted it. The payload path is covered by the serializer test and by
    `check_payload` on a freshly built payload, both green.
  - **No `Co-Authored-By` / session trailer on the commit**, per the standing repo rule (no AI
    tells in commits or PRs for os-ai-map). Flagging it since it differs from the generic
    harness attribution instruction.

- alternatives rejected:
  - A `status: live | sunsetting | withdrawn` enum (#262's third open question). An enum needs a
    withdrawn state, and withdrawn is retirement, which the repo already implements by alias — so
    it would put two retirement mechanisms in the schema at once. It also has no room for the
    date, which is the fact a reader wants. The two adjacent cases #262 names
    (`google-coral-dev-board`, `gpt4all`) are "wound down" and "unmaintained", which are
    inferences from evidence rather than announced dates; they need their own field or none, and
    forcing them into this one would make the date optional and the field mushy.
  - Emitting a derived `ended: true` when the date is past. It would let the front end pick
    wording without re-deriving, but it makes a committed payload change on a day boundary with
    no source change, which is exactly the "declared rather than inferred" line the disclosure
    flag already draws in `serialize.py`.
  - A gate (or a `check_retirement` line) that fires on a past `end_of_life`. It turns an
    unattended CI run red for a fact nobody changed. Named the right home in the docs instead —
    the verification queue, where a curator picks work up.
  - Adding the key to `build/details_payload.py`'s `PRODUCT_KEYS`. That payload is the marimo
    Details modal, under a hard 8 MB cap that has already been breached once, and `render.py`
    would not render the key anyway. Carrying a field nothing reads is the defect that module
    exists to prevent.

## Risks / unsure about

- The payload is not committed, so grepping `build/notebook_data.json` on this branch finds no
  `end_of_life`. If the reviewer reads criterion 4 as "the committed payload must contain it", the
  fix is one `uv run python -m build.serialize` — but it contradicts an explicit repo rule and CI
  is documented to block hand edits, so I did not.
- The schema shape is a judgment call against the criterion's wording ("an optional `end_of_life`
  date"). It accepts an optional end-of-life date; the date sits under `end_of_life.date` so the
  required source has somewhere to live. If the reviewer wants a scalar, the source has to move
  into prose and stops being machine-readable.
- The general rule (no number moves) is now documented and tested, but it has never been exercised
  by a case where it costs something. The first fully open, mature, ending product is where it
  gets argued about; #262's re-ruling comment says the same and calls it worth deciding early.
- `aipotluck.org` does not read the key, so today a reader still meets the sunset only through the
  description prose. Named in the commit body and in identity.md; out of scope here, and it is a
  front-end issue on the monorepo, not on os-ai-map.
- `docs/sweeps/` and the skills directory were left alone. `skills/update-product/SKILL.md` is a
  thin wrapper over the workflow doc I did edit, so it should still route correctly, but I did not
  re-read every skill for a place this ought to be mentioned.

## Response to round N-1

|---|---|---|---|---|
| n/a — round 1, no prior review | | | | |

ccerv1:
## What I did
- files touched:
  - `AGENTS.md` — reverted to the `HEAD~1` version. The round-1 commit had widened the
    "four further optional keys" sentence to five and named `end_of_life`; that whole hunk is gone.
  - `docs/reference/identity.md` — rewrote the "Nothing happens when the date passes" paragraph.
    It now states only the implementation fact (nothing in the build compares `end_of_life.date`
    to today, so a record with a past date serializes, validates and scores as before) and defers
    the curation policy to its own issue, with an explicit line that deferral is not permission to
    leave a past date sitting there.
  - `docs/workflows/update-product.md` — the same deferral, in the one sentence of the how-to that
    had repeated the policy ("re-read the source then and decide whether to keep the record or
    retire the slug" → "no gate compares the date to today", policy left open).
  - No other file changed from round 1. Schema, `perspective-api.yaml`, `serialize.py`,
    `check_payload.py` and the tests stand as reviewed.
- commands run:
  - `git checkout HEAD~1 -- AGENTS.md`
  - `uv run python -m build.validate` → `0 error(s), 2 warning(s)` (both pre-existing
    `model_families` pattern-overlap warnings, unrelated)
  - `uv run pytest -q` → `1881 passed, 1 skipped in 1084.00s`
  - `grep -n "date.today\|datetime.now\|utcnow" build/serialize.py build/check_payload.py
    build/validate.py` → three hits, none of them `end_of_life`: `serialize.py:487` stamps the
    payload's `generated` field, `validate.py:580` bounds `accessed` / `last_verified` on score
    blocks. This is the evidence behind the retained implementation sentence.

## Why
- decisions:
  - Both items are fixed, neither pushed back. Item 1 is a plain scope breach and the reviewer is
    right that the tests were authorized and `AGENTS.md` was not. Item 2 is the sharper call: the
    round-1 paragraph said a curator "either keeps the record with a past date … or retires the
    slug", which reads as a settled policy that an expired product may stay scored. That decision
    was not mine to make in the change that added the field.
  - What I kept is the part that is a fact about the code rather than a rule about people: no gate
    compares the date to today. A reader needs that, because otherwise the obvious question after
    reading the field's definition is whether the build will do something on 2027-01-01, and the
    answer is verifiable rather than a matter of policy.
  - I added one clause the reviewer did not ask for: "do not read this section as permission to
    leave one there indefinitely." A bare "not settled here" next to "nothing in the build changes"
    is easy to read as a default of indefinite drift, which is exactly the outcome the review was
    guarding against. Flagging it as an addition rather than burying it.
  - The `end_of_life` block is still an object with a required `source` and `shows`. That was not
    challenged in round 1 and I have not revisited it.
- alternatives rejected:
  - Rewriting `AGENTS.md`'s sentence more narrowly instead of reverting it. Nothing in the repo
    gates on that list — `validate.py:643` only mentions `AGENTS.md` in a comment — so a partial
    keep would have bought an out-of-scope edit nothing needs.
  - Deleting the paragraph outright. That leaves the 2027-01-01 question unanswered in the one
    document a reader consults, and invites the next person to answer it by guessing at the code.
  - Citing an issue number for the deferred policy. There is no such issue and this clone has no
    remote, so a number would have been invented. The text says the policy belongs on its own
    issue and stops there; filing it is a follow-up, named in the handoff below.

## Risks / unsure about
- The deferred policy is now a real loose end rather than a documented one. Someone has to file it.
  It is not urgent — the first case, `perspective-api`, expires 2026-12-31, so there is a quarter
  of runway — but it should be filed before then, not after.
- Rendering remains out of scope and unbuilt. `aipotluck.org` does not read `end_of_life` yet, so
  a reader of the live map still meets the sunset only through the prose fields. Named in round 1,
  still true, still needs its own issue on the front-end repo.
- The claim that no arithmetic consults the calendar rests on a grep of `build/` plus reading the
  two hits. It is a negative over one directory. `warehouse/` is out of scope for this change and
  I did not check whether anything there date-compares, so the sentence is scoped to the build.

## Response to round 1
|---|---|---|---|---|
| `AGENTS.md:149` — outside the settled `sources/`, `docs/`, `build/` scope | yes | | | `git checkout HEAD~1 -- AGENTS.md`; the file is byte-identical to its pre-change state, and `git diff HEAD~1 --stat -- AGENTS.md` is empty. Suite still passes, confirming nothing gated on the sentence. |
| `docs/reference/identity.md:127` — post-expiry curation policy decides the held 2027-01-01 question | yes | | | The paragraph now asserts only that nothing in the build compares `end_of_life.date` to today (verified: the only `date.today()` calls are `serialize.py:487` for the payload `generated` stamp and `validate.py:580` for score `accessed`/`last_verified` bounds). The "keep it or retire it" policy is gone from both `identity.md` and `update-product.md` and is marked as needing its own issue. |

PR #528, opened 2026-09-09
Address a source entry by position, so a duplicated URL cannot be written to the wrong citation

### LINKED ISSUE #527: reverify stamps an axis whose dimension was never re-read, because a duplicate URL sends the write to the wrong source entry
Found while working #521. The `reverify` workflow has **one run in its whole history and it failed** (run [34236753717](https://github.com/currentai-org/os-ai-map/actions/runs/34236753717)), so the deterministic leg of the rolling re-verification has never completed once. The invariant gate is what stopped it, correctly:

```
invariant  a confirmation with no supporting evidence  [FAIL]
  ! compar-ia:openness: records 'source' (as 'source') but only established by a source
    last read 2026-06-22; last_verified 2026-09-08 is not supported for that dimension
```

## Root cause

`compar-ia`'s openness axis cites the same two URLs twice each — legal under the schema, and there are **55 (product, axis) pairs across the corpus that do this, 140 entries in total**:

| establishes | accessed | digest | url |
|---|---|---|---|
| license | 2026-06-22 | no | `raw.../develop/LICENSE` |
| license, source | 2026-06-22 | no | `github.com/betagouv/ComparIA` |
| core-gated | 2026-08-11 | yes | `raw.../develop/README.md` |
| license | 2026-08-13 | yes | `raw.../develop/LICENSE` |
| source, core-gated | 2026-08-13 | yes | `raw.../develop/README.md` |

`components.set_source` resolves a URL with `next(i for i, e in enumerate(entries) if e.get("url") == url)` — **the first occurrence**. So when `reverify` fetches a URL once and writes the result back by URL, the write lands on whichever entry happens to come first, not the entry whose confirmation it actually computed.

Reproduced on a scratch copy at today's date:

```
stamped: ['openness'] drifted: 0
last_verified -> 2026-09-09
  est=['license']            accessed=2026-09-09 digest=y  LICENSE      <- was 06-22, undigested
  est=['license','source']   accessed=2026-06-22 digest=N  github.com/...
  est=['core-gated']         accessed=2026-09-09 digest=y  README.md
  est=['license']            accessed=2026-08-13 digest=y  LICENSE      <- the entry actually re-fetched
  est=['source','core-gated'] accessed=2026-08-13 digest=y README.md    <- the entry establishing `source`
```

Two defects, and the second is the one that fails the gate:

1. **Write misattribution.** The first `LICENSE` entry had no `http_status` and no `content_sha256` — it had never been fetched. It now carries today's date *and a digest*, while the entry that was actually re-fetched keeps its old date. A fetch record was attached to an entry the pass did not fetch as that entry.
2. **Citation identity is lost before the write.** `reverify.py:250` holds confirmations in `confirmed: dict[str, dict]` — **keyed by URL**. Two entries citing one URL collapse to a single key, and `apply` hands that one record to `set_source`, which lands it on the first occurrence. So the entry establishing `source` never gets its date bumped, `check_verification.invariant` finds no establishing source fresher than the claimed date (`accessed >= last_verified`, `check_verification.py:182-200`), and the axis is dated on evidence it doesn't have.

   ~~`reverify` confirms a dimension when *any* digested establishing source confirms.~~ **Corrected:** it requires **all** planned digested establishing sources to confirm — every drift or transient sets `ok = False` and the axis stamps only `if ok and confirmed`. There is no ANY-versus-freshness hole; the defect is entirely the identity collapse at writeback.

This is the same shape as the parity problems already on record: two pieces of code mirroring one rule by hand and drifting.

## What the gate did right

Nothing bad reached `main` — the run failed before the commit step, and every subsequent checkout is clean. The cost is that the deterministic leg has never produced a PR, and the agent leg (#526) has been carrying the whole freshness chain alone.

## Fix options, needs a call

- **(a) Make ambiguity loud.** `set_source` raises when a URL matches more than one entry, and `reverify` addresses entries by index rather than URL. Smallest change, turns a silent misattribution into a failure, and leaves the 140 duplicate entries alone.
- **(b) Address by identity.** Resolve an entry by `(url, establishes)` rather than `url`, in both `set_source` and `reverify`.
- **(c) De-duplicate the data.** Merge each duplicated pair into one entry carrying the union of `establishes`. Removes the hazard at the source, but it is a 140-entry corpus edit and it re-digests those entries.

(a) plus (b) look right to me — make the write path refuse to guess, and give it a way to say which entry it means. (c) is a separate curation question.

Blocks #521, which cannot go green until this is fixed.

ccerv1:
Closes #527.

## Why

The `reverify` workflow has **one run in its entire history and it failed**, so the deterministic leg of the rolling re-verification has never completed once and the agent leg has been carrying the freshness chain alone. The invariant gate is what stopped it, correctly:

```
! compar-ia:openness: records 'source' (as 'source') but only established by a source
  last read 2026-06-22; last_verified 2026-09-08 is not supported for that dimension
```

An axis may cite one page twice under different `establishes` lists. That is legal, and **55 (product, axis) pairs do it**. Citation identity was lost twice on the way to the write:

1. `reverify` held confirmations in `confirmed: dict[str, dict]` **keyed by URL**, collapsing two entries into one record.
2. `components.set_source` resolved a URL with `next(i for i, e in enumerate(entries) if e["url"] == url)` — **the first match**.

So a pass that confirmed the second entry stamped the first. The confirmed entry kept its stale date, the axis was dated anyway, and the gate rejected the whole run.

## What changed

An ordinal identifies an entry where a URL cannot, so the ordinal is carried end to end — planning, confirmation storage, and writeback, which is the part that has to hold or the hole stays open:

- `plan_product` yields `(position, src)`
- `confirmed` is keyed by position
- `apply` passes `index=` to `set_source`
- the fetch cache stays keyed by URL, because fetching one page once is right; only the write-back needs to know which citation it belongs to
- `set_source` **raises on an ambiguous URL** rather than guessing, and rejects an `index`/`url` pair that disagrees, so an ordinal computed against a stale parse cannot land a fetch record on an unrelated citation

`_source_span` moves to the same selector, which fixes a second defect it was hiding: it matched with `url in block`, so a URL that is a **prefix** of another resolved to the wrong entry. Real shape, hit on openpipe earlier today — a record cited `.../faq.md` and the live page is `.../faq`.

## What this deliberately does not do

- **No corpus de-duplication.** Merging the 140 entries means reconciling `shows`, digests and dates — a curation question — and one entry per URL would not stop the next duplicate anyway.
- **No change to the confirmation rule.** Every drift or transient already sets `ok = False`, so there was never an any-versus-freshness disagreement with the gate, only the identity collapse. #527's original text said otherwise and is corrected.

`skills/clean-score-notes` already documented this hazard and prescribed addressing by index. The code now enforces it instead of asking the author to remember.

## Test plan

- New unit test reproduces the production error string verbatim and passes
- New test asserts both entries citing one URL are re-dated
- New test asserts a stamped axis satisfies the invariant gate — the assertion the 2026-09-08 run failed
- New test pins the prefix-URL case
- On a copy of the real corpus: `compar-ia` re-verifies, both README entries take the new date, the never-fetched LICENSE entry stays untouched instead of gaining a digest, and `invariant` returns **0 problems corpus-wide**
- `build.validate` 0 errors; all four `check_verification` gates OK
- `uv run pytest -q`: **1888 passed**

Reviewed once by Codex before implementation; it corrected two things in the original analysis, both folded in above.


ccerv1:
## Codex review, three passes

Reviewed before implementation (on the diagnosis) and twice on the diff. The second pass found one blocking issue, now fixed in 5c6a5f8f; the third passed clean.

**The blocking finding, quoted:**

> `build/reverify.py:300-310` — The stale-parse safeguard compares against a URL recovered from the fresh file, not the confirmed entry. If a citation changes or entries reorder between confirmation and `apply()`, the check passes for the new occupant. Reproduced: a confirmation fetched for `/OLD` stamps `/NEW` with `/OLD`'s digest.

Correct, and the worst shape of wrong: `apply` read `entries[position]["url"]` from a fresh parse and passed it alongside `position` into that same parse, so the index/url agreement check was comparing `x == x`. It could never fire, and the comment above it advertised a protection that did not exist. The URL is now captured at the moment of confirmation in a frozen `Confirmation(url, fetched)`, so the check compares what was confirmed against what is on disk. `test_apply_refuses_a_confirmation_whose_citation_changed_underneath_it` replaces the citation at the confirmed position between `reverify_product` and `apply` — it wrote silently before, and now raises with the file byte-identical.

**Third pass, `settled_now`:**

> All three shown confirmation paths capture the source URL in `Confirmation`. `apply` passes `confirmation.url` directly; the fresh-parse tautology is gone.
>
> A guard exception during any entry update occurs before the sole `write_text` call, so earlier in-memory edits cannot partially land on disk.

Two non-blocking notes, both addressed. It could not verify from the diff whether the shared fetch cache is immutable — checked by hand: `_shows_confirms`, `_spdx_confirms` and `_read_body` only read from the record, and `fetch()` returns a fresh dict per URL, so no mutation is possible. And it suggested the refusal test compare full file text rather than assert nothing was stamped, which is the stronger claim since an earlier entry is already edited in memory when the guard fires — done in 1cd307cf.

**One finding deliberately left out of scope.** The last entry's span runs to the axis end, so an axis with a field after `sources` raises. Measured: 24 axes, **zero on openness**, which is the only axis `reverify` writes by default. Pre-existing, caught by the reparse guard rather than corrupting, and unrelated — filed separately.

Verification: `pytest` 1889 passed, `validate` 0 errors, all four `check_verification` gates OK, and on a copy of the real corpus `compar-ia` re-verifies with `invariant` returning 0 problems corpus-wide.

Issue #319, opened 2026-08-17
[scoring] Score capability for evaluation datasets


ccerv1:
## The axis already exists here — it is applied to 4 products out of 27

`benchmark_eval_data` is internally inconsistent about whether an evaluation set can be capability-scored:

| Product | Capability | Recorded basis |
|---|---|---|
| LiveBench | 5 | 6 domains, objective ground-truth scoring, monthly question rotation for contamination resistance |
| GAIA | 5 | 466 questions, 3 levels, multi-step tool use and live browsing, one checkable answer |
| Humanity's Last Exam | 5 | 2,500 expert questions, 100+ subjects, multimodal, private held-out set |
| MT-Bench | 4 | 80 multi-turn questions, LLM-as-judge, "now small and partly saturated vs harder peers" |
| HumanEval | null | "A dataset is not 'capable'. Worth noting it is now largely saturated/contaminated as a discriminator, but per recipe capability is n/a here" |
| MMLU | null | "Dataset, not capability-scored. Saturation/contamination (top models 88-94%, superseded by MMLU-Pro for differentiation) is a quality caveat noted in version_note" |

Those two positions cannot both be right. The four scored benchmarks are banded on **discriminative power** — contamination resistance, objective versus judge-based scoring, breadth, difficulty. The abstentions deny that axis exists while recording saturation in prose and routing it to `version_note`. The rubric is not missing; it is unevenly applied, and it skips exactly the products it would mark down.

## Proposed

Write the rubric down — capability for an evaluation set is **evaluative power**: does it still separate frontier models? The dimensions are already visible in the four existing notes:

- contamination resistance (live rotation, private held-out set)
- objective/checkable scoring versus LLM-as-judge
- domain breadth
- item count and difficulty
- headroom — MMLU at 88-94% top-model accuracy is not unscoreable, it is a **2**

Then score the remaining 23.

## What it changes

**The category stays at Stage 3.** The saturated classics (adoption 4, capability null, blending to 4.0 on adoption alone) drop to about 3.2 once capability lands, but LiveBench is `open` with capability 5 and adoption 3, blending to 3.8 at the category's 0.6/0.4 weights — above the 3.5 band. Nothing on the published map moves.

[...]

PR #543, opened 2026-09-11
Score capability for five benchmark_eval_data products, and write the rubric they were banded against

ccerv1:
Calibration slice for issue #319. Five products scored, not the full 28, so the rubric application can be checked before the remaining 23 land in one tranche.

## The rubric, written down

`benchmark_eval_data` had no capability recipe. The category now declares one in its `scoring_recipe`, with `bands: null` and hand-authored rungs on `embeddings_retrieval`'s precedent, because the four dimensions do not combine by counting - GAIA is a 5 on a single domain and MT-Bench a 4 with judge scoring and no contamination control, so an attribute count reproduces neither.

The question is the one the issue names: does this benchmark still separate frontier models? Headroom leads, and contamination control, checkable scoring and distinctness say whether it will last.

- **5** - still separates frontier models and the design holds that: answers held out or the item set replaced over time, and an answer checkable without a judge.
- **4** - a real discriminator on a dimension nothing cheaper reaches, with one structural weakness: a judge, a small item set, or partial saturation.
- **3** - harder than the classic it replaced and still separating, but static and published in full.
- **2** - plateaued; a reporting convention rather than a discriminator.
- **1** - saturated, or shown not to measure what it claims.

The rungs reproduce all four existing bands: livebench, gaia and humanitys-last-exam at 5, mt-bench at 4. Three rules are recorded with them - saturation is a fact about the benchmark rather than a caveat about it, openness does not enter the band, and gating is not contamination control on its own (livecodebench is gated over an unstated license, math under a DMCA takedown).

## The five

**terminal-bench - 5**, basis `feature_matrix`.
Read the Terminal-Bench 2.0 paper (arXiv 2601.11868) and the repository README. The paper reports 89 tasks, each with its own environment, a human-written solution and comprehensive tests, and says frontier models and agents score under 65%. The README shows the per-task test script and oracle solution, the leaderboard pinned to a named dataset version, and the standing pointer at Terminal-Bench 2.0. Real headroom, graded by the task's own tests with no judge, and the suite is replaced version by version rather than frozen, which is how LiveBench holds this rung. The live leaderboard at tbench.ai was deliberately not cited: it renders its rows in the browser and its digest changed between two fetches minutes apart.

**swe-bench-verified - 4**, basis `feature_matrix`.
Read the HF dataset card (500 human-validated instances, "Evaluation is performed by unit test verification using post-PR behavior as the reference solution"), the SWE-bench paper, and the SWE-bench repository README. A real discriminator on a dimension nothing cheaper reaches, which is why frontier labs report against it. One rung below the top because the set has been frozen since August 2024 and the gold patch and tests ship with every instance, so its headroom is being spent rather than protected.

**mmlu-pro - 3**, basis `feature_matrix`.
Read the MMLU-Pro paper. Ten options instead of four, reasoning-focused items, noisy questions removed, a measured 16-33% accuracy drop against MMLU and prompt sensitivity down from 4-5% to 2%. Measurably more discriminating than what it replaced, and the claim is a measurement. It stays off the top two rungs because questions and answers ship together, nothing rotates and nothing is held back.

**mmlu - 2**, basis `feature_matrix`.
Read the original MMLU paper for the design and the MMLU-Pro paper for the plateau: "as models continue to improve, their performance on these benchmarks has begun to plateau, making it increasingly difficult to discern differences in model capabilities." The issue's own worked example. It survives as a reporting convention, which is the level-2 rung rather than an abstention.

**gsm8k - 1**, basis `feature_matrix`.
Read the GSM8K paper and GSM-Symbolic (arXiv 2410.05229), which reports that "the performance of all models declines when only the numerical values in the question are altered" and that one irrelevant added clause costs up to 65%. A score that moves on a rewrite preserving the problem is not a reading of mathematical reasoning, so GSM8K separates frontier models on neither difficulty nor reliability.

## The contamination instrument (added on review)

The Hugging Face API reports how many models declare a corpus as a training dataset, which is a countable, re-derivable input to the contamination-resistance dimension this recipe leads with:

```
https://huggingface.co/api/models?filter=dataset:<owner>/<name>&limit=1000&full=false
```

Readings across the category, taken 2026-09-11 and reproduced independently before recording:

| product | count | capability |
|---|---:|---:|
| livebench | 0 | 5 |
| terminal-bench | 0 | 5 |
| gaia | 5 | 5 |
| humanitys-last-exam | 11 | 5 |
| mt-bench | 1 | 4 |
| swe-bench-verified | 13 | 4 |
| humaneval | 29 | null |
| mmlu-pro | 30 | 3 |
| mmlu | 327 | 2 |
| gsm8k | 834 | 1 |

It reproduces all four pre-existing anchors and rank-orders the five new scores. It is declared under `contamination_control` in the recipe and added as corroborating evidence on each of the five records. **No score changed** — they stay 5/4/3/2/1.

**The signal is asymmetric, and that is the load-bearing part.** A high count is strong evidence that a benchmark has been trained on. A low count establishes nothing, because a corpus can reach a pretraining mix without any model tagging it on the Hub. `humaneval` is the proof: 29 models declare it and it is known-saturated on every other reading. So the instrument may move a band **down** and may never move one **up**, and it never overrides a read of the primary sources. Same rule as a transient fetch not being evidence of absence.

It is also one input to one of four dimensions rather than a rung of its own. `mt-bench` sits at 1 with a capability of 4, marked down for LLM-as-judge scoring and an 80-question set — facts about its design, not about contamination.

**The API, never the rendered page.** The two disagree: 834 against 1,109 for `openai/gsm8k`. The API figure is recorded because it is the one that re-fetches to a stable digest; the page is browser-rendered, the same class of source already avoided for tbench.ai. The discrepancy is written into the gsm8k record's `shows`.

## Published numbers move, and the issue predicted otherwise

The issue says the category stays at Stage 3 and nothing on the published map moves. It was already at **Stage 4**, and this change takes it to **Stage 3**.

`gsm8k` was the only fully-open product clearing the 4.5 maturity bar, and it cleared it on adoption 5 with a null capability. At the category's 0.6/0.4 weights it now blends to 3.4.

| | before | after |
|---|---|---|
| stage | 4, Competitive Open Ecosystem | 3, Viable Alternatives |
| gaps | `resiliency` | none |
| best fully-open | gsm8k, 5.0 | humaneval, 4.0 |

The `adoption` gap the issue expects does not fire either: the best fully-open product is now an adoption-only record at 4.0, which clears the adoption cutoff, so no driver gap fires. That may change once the remaining 23 land and the adoption-only records get capability scores of their own.

`notebook_data.json` and the notebooks are left to the regenerate bot.

## Open questions for review

- **`basis`.** All five use `feature_matrix`, matching the four anchors. An argument exists for `benchmark` here, since the reading behind the band is a published frontier figure on the product itself. Consistency with the anchors won; say if it should be the other way.
- **No peer edges.** Nothing records `relative_to`/`relation`. The rungs place the bands, and adding edges into a rubric under review seemed premature.
- **Rung distribution.** 9 of 32 scored, 4/2/1/1/1 across 5/4/3/2/1. The top rung is 44% of what is scored, which is not yet a distribution to judge - the 23 still unscored are mostly the saturated classics this rubric puts on 1 to 3.

## Gates

```
build.validate            0 error(s), 2 warnings (pre-existing model_families patterns)
build.check_verification  all four gates OK
build.check_capability    OK (12 weak roots pre-existing, report-only, none in this category)
build.check_recipe        0 failure(s)
build.check_rubric        benchmark_eval_data 30/30 reproduced, 2 deferred
build.check_freshness     all 20 categories within 45d
build.check_refetch --strict, each of the five: [OK] no fabricated digests
pytest -q                 1888 passed, 1 skipped (re-run after the instrument commit)
```

Every capability source on all five products, the training-tag counts included, re-fetched and reproduced its digest exactly. The drifts `check_refetch` reports on these products are on pre-existing openness and adoption sources (Hugging Face pages), not on anything added here.

Closes nothing yet - #319 stays open for the remaining 23.

### DIFF
--- a/sources/categories/benchmark_eval_data.yaml
+++ b/sources/categories/benchmark_eval_data.yaml
@@ -91,6 +91,78 @@ scoring_recipe:
         rung that tests a tier, and the recorded 4 stands rather than being replaced by a
         number nobody derived. Extending the dataset ladder with a use-bounded tier is the fix
         and it is a rubric change.
+  capability:
+    basis: feature_matrix
+    question: Does this benchmark still separate frontier models?
+    dimensions:
+      headroom:
+        question: How much room is left between the best reported score and the ceiling?
+        evidence: A published frontier figure on this benchmark, or a successor's measured
+          comparison against it.
+      contamination_control:
+        question: Does the design defend its headroom, by holding answers back or by replacing
+          items over time?
+        evidence: A held-out or private split, a documented rotation or version cadence, or a
+          canary string on the published artifact.
+      checkable_scoring:
+        question: Is an answer graded against ground truth or executed tests rather than by a
+          judge model?
+        evidence: The scoring procedure as the project documents it.
+      distinct_dimension:
+        question: Does it measure something no cheaper instrument already covers?
+        evidence: What the project states it tests, read against the other benchmarks on this
+          roster.
+    anchor:
+      primary: the benchmark's own reported frontier ceiling
+      machine_signal: none - requires research. `semanticscholar` routes adoption here and
+        corroborates capability without settling it.
+    bands: null
+    bands_status: >-
+      Hand-authored against the rungs below, on embeddings_retrieval's precedent. The four
+      dimensions do not combine by counting: GAIA is a 5 on one domain and MT-Bench a 4 with
+      judge scoring and no contamination control, so an attribute count reproduces neither.
+      Headroom leads and the other three say whether it will last. The layer-2 scorer must not
+      write this axis.
+    note: >-
+      Written on 2026-09-11 under issue #319, which ruled that a benchmark's difficulty and
+      contamination resistance IS a capability and that the category could not hold both
+      positions at once - four products banded on exactly that, and 28 abstaining at the time
+      of the ruling on the grounds that a dataset is not capable. The rungs below are read off
+      the four bands that already existed rather than invented beside them.
+
+      5 - still separates frontier models AND the design defends that: answers held out or
+      the item set replaced over time, and an answer is checkable without a judge. livebench
+      (monthly rotation, judge-free, top models under 70%), gaia (300 answers private, one
+      checkable answer, 15% against 92% for humans) and humanitys-last-exam (private holdout)
+      hold this rung.
+      4 - a real discriminator on a dimension nothing cheaper reaches, with one structural
+      weakness: judge-based scoring, a small item set, or partial saturation. mt-bench is the
+      anchor - 80 multi-turn questions and an LLM judge, still the instrument for a dimension
+      multiple choice cannot reach.
+      3 - harder than the classic it replaced and still separating, but static and published
+      in full, so its headroom is being spent rather than defended.
+      2 - plateaued. Frontier scores cluster near the ceiling, its own successors say so, and
+      it survives as a reporting convention rather than as a discriminator.
+      1 - saturated, or shown not to measure what it claims: frontier accuracy is at the
+      ceiling, or published work shows the score moves on rewrites that preserve the problem.
+
+      Three rules settle the calls the rungs alone leave open, and each one came out of
+      reading the four existing bands rather than out of scoring a product.
+      (a) Saturation is a fact about the benchmark, not a caveat about it. Recording "now
+      largely saturated" in prose while abstaining on the axis is the inconsistency #319
+      names; a saturated benchmark is a 1 or a 2 and the note says why.
+      (b) Openness does not enter the band. gaia and humanitys-last-exam are gated
+      specifically to hold answers back, and that is what earns them the top rung on this
+      axis. The openness axis records the cost separately.
+      (c) Gating is not contamination control on its own. livecodebench is gated over an
+      unstated license and math under a DMCA takedown; neither withholds answers to defend a
+      score.
+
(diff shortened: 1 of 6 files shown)


ccerv1:
Reviewed by Codex before closing, and the verdict is that closing unmerged is right on the evidence alone — independent of the direction change in #548. Five blocking defects: the contamination instrument's asymmetry rule holds only for declared use and cannot speak to the evaluation split; three of the five scores break that rule by treating a low count as evidence; the count query caps at limit=1000 so array length is not a total; gsm8k at 1 rests on a robustness finding that does not establish saturation, with no contemporary frontier accuracy cited; and the recipe is fitted to the existing scores rather than defining reproducible boundaries.

Full findings and the salvage guidance are on #546, so the work stays findable. Nothing here is wasted — the research dimensions and the cited evidence are worth keeping, with narrower wording.

ccerv1:
Closing unmerged, as agreed. The direction this explored was set aside in #548, and Codex review found the slice's own evidence would not have supported merging it regardless — five blocking defects, recorded on #546 along with what is worth salvaging.

Nothing here is wasted. The research dimensions, the cited evidence and the contamination-instrument idea all survive on #546 with narrower wording; the branch stays available if anyone reopens the question.

Issue #547, opened 2026-09-11
A null capability suppresses the capability gap, so a category can report a stage with no gaps


ccerv1:
Found while evaluating #319. Independent of how that was resolved.

`_maturity_score` in `build/serialize.py` treats a null capability as "grade this product on adoption alone":

```python
capability = (row.get("capability") or {}).get("score")
if capability is None:
    return round(float(adoption), 2)
```

That is deliberate and documented. But it has a second effect that is not documented and is easy to miss.

## The two effects

1. **It inflates maturity.** A product with adoption 5 and a null capability scores 5.0, the same as one that earned 5 on both axes.
2. **It suppresses the capability gap.** At stages 1 to 3 the engine names the drivers holding the best fully-open product back. A capability gap fires when that product's capability is below 4. A **null is not below 4** — it is not comparable at all — so no capability gap fires.

Put those together and a category whose best fully-open product has a null capability can report a stage with an **empty gap set**: a reader sees a category sitting at Stage 3 and is told nothing about what is missing.

## It is reachable today

Simulated on `benchmark_eval_data` during the #319 calibration: with gsm8k scored and the rest of the category still abstaining, the category lands at Stage 3 with `gaps: []`. The best fully-open product was then humaneval at maturity 4.0 with a null capability, so nothing could fire.

That particular path is closed now that #319 resolved toward abstention, but the shape is general: any category whose best fully-open product abstains on capability has the same hole.

## Also worth reconciling

`_stage_and_gaps` excludes a product with **no adoption** from the stage computation entirely — "we can't judge what we can't measure, so they neither advance nor depress the category's stage" — and then measures anyway when **capability** is missing. Two missing axes, two different rules, and only one of them is argued for in a comment.

There is a real argument for the asymmetry, since maturity is anchored on adoption. There is also a real argument that a product graded on one axis should not count toward a stage that claims to be about both. Currently only 28 of 664 products with an adoption level are graded this way, and exactly two of those are fully open with adoption at or above 4.5, so the blast radius of changing it is small and measurable.

## Suggested fix

Either make a null capability exclude the product from the stage computation, as a missing adoption already does, or keep the current grading and emit a distinct gap or flag when the best fully-open product's capability is unscored — so that a category is never silently gapless. The first is more consistent; the second is smaller.

ccerv1:
## Correction and measurement

[...]

Issue #546, opened 2026-09-11
[methodology] What capability means for a dataset, and why the evaluative-power reading was rejected


ccerv1:
Raised out of #319, which is closed. This records the open question so the argument is not lost, and so the next person to notice the same inconsistency finds the reasoning rather than re-deriving it.

## What was settled

`benchmark_eval_data` was internally inconsistent: four benchmarks carried a capability score and the rest abstained. #319 proposed closing that by scoring the remaining twenty-three on **evaluative power** — does this benchmark still separate frontier models. A five-product calibration slice was built and then rejected, and the four scored records now abstain instead.

The reason it was rejected is worth stating precisely, because it is not "the scores were wrong":

**Capability read as discriminative power decays with use.** The more a benchmark is adopted, the more models train on it, the less it separates them. That makes capability **anti-correlated with adoption** — and `maturity` blends the two at equal weight. A composite of two anti-correlated axes carries little meaning, and it guarantees every successful benchmark trends toward 1 over time. Stage numbers would then move on a calendar rather than on evidence.

The calibration slice demonstrated it concretely: gsm8k scored 1 on that reading, which dropped `benchmark_eval_data` from Stage 4 to Stage 3, and the drop was driven by the single product whose capability had been null.

## What is still open

**1. Is there a reading of capability that works for a dataset?** "Evaluative power" fails for the reason above. A reading based on the *artifact* rather than its current discriminative state — breadth, item count, construction rigour, scoring mode, whether answers are held out — does not decay and would not fight adoption. It was not evaluated. If someone wants to reopen this, that is the version to argue for, not the one #319 proposed.

**2. The abstention has a side effect worth knowing.** A null capability does two things, and the second is easy to miss: it grades the product on adoption alone, *and* it suppresses the capability gap, because a null cannot be "below 4". A category whose best fully-open product has a null capability can therefore report a stage with no gaps at all. Filed separately.

**3. A contamination instrument now exists and is unused.** The Hugging Face API returns the number of models trained on a dataset:

    https://huggingface.co/api/models?filter=dataset:<owner>/<name>&limit=1000&full=false

Measured across this category it rank-orders exactly as intuition about saturation does — livebench 0, gaia 5, hle 11, mt-bench 1, swe-bench-verified 13, humaneval 29, mmlu-pro 30, mmlu 327, **gsm8k 834**. Note the API figure differs from the rendered page, which reports 1,109 for gsm8k; the API figure is the re-derivable one and the page should not be cited.

[...]

PR #548, opened 2026-09-11
Abstain on capability for the four scored evaluation datasets

### LINKED ISSUE #319: [scoring] Score capability for evaluation datasets

ccerv1:
Closes #319, resolved in the opposite direction to the one it proposed.

## Why

#319 was right that `benchmark_eval_data` was internally inconsistent: four benchmarks carried a capability score and twenty-eight abstained, and both positions cannot be right. It proposed closing the gap by scoring the remaining twenty-three on **evaluative power** — does this benchmark still separate frontier models. A five-product calibration slice (#543) was built against that reading and then rejected.

**Capability read as discriminative power decays with use.** The more a benchmark is adopted, the more models train on it, the less it separates them. That makes capability anti-correlated with adoption — and `maturity` blends the two at equal weight, so the composite means very little and every successful benchmark trends toward 1 over time. That is not a correction to apply once. It is a treadmill that would re-score the category on a calendar and move published stage numbers with it.

The calibration slice showed it concretely: gsm8k scored **1**, which dropped the category from Stage 4 to Stage 3 on the strength of the single product whose capability had been null.

The abstaining records already had this right. They recorded saturation and contamination as a quality caveat in prose and left the axis unscored. `build/serialize.py` says the same in its own comment: capability may be legitimately null for some product types, which are graded on adoption alone.

## What changed

`livebench`, `gaia`, `humanitys-last-exam` and `mt-bench` now abstain, matching the shape `humaneval` and `mmlu` already use: `score: null`, `basis: n/a`, `value: null`.

Each note carries forward what the record previously asserted, as the quality caveat it always was:

- LiveBench's six domains, judge-free auto-scoring and monthly rotation
- GAIA's 466 tool-use questions with one checkable answer
- Humanity's Last Exam's 2,500 expert questions and private held-out set
- MT-Bench's 80 multi-turn questions and LLM-as-judge scoring

**Every source is kept.** Nothing that was read is lost, and the arXiv and lastexam.ai citations still carry their digests.

## No published number moves

`benchmark_eval_data` stays **Stage 4, Competitive Open Ecosystem, gaps `[resiliency]`** — re-derived through `_stage_and_gaps` before and after.

## Test plan

- `build.check_verification` — invariant / digests / producible-pairs / placeholder-shows all OK
- `build.check_capability` — OK
- `build.check_rubric` — `benchmark_eval_data` 30/30 reproduced, 2 deferred
- `build.check_recipe` — 0 failures
- `uv run pytest -q` — 1895 passed, 1 skipped, **1 failed**: `test_publish_evaluation.py::test_plan_validates_offline_and_writes_nothing`. It passes in isolation, and a sibling test in the same file failed once earlier today under different conditions. Reported rather than explained away — CI runs the suite split (`-n auto -m "not serial"`, then `-m serial`) and is the arbiter.

## Follow-ups filed

- **#546** records the methodology argument, including the artifact-based reading of capability that was *not* evaluated and is the version to argue if anyone reopens this, plus the Hugging Face contamination instrument that is now measured and has nowhere to live.
- **#547** the null-capability hole: a null both grades on adoption alone and suppresses the capability gap.

#543 stays closed and unmerged as the rejected calibration.

### DIFF
--- a/sources/scores/gaia.yaml
+++ b/sources/scores/gaia.yaml
@@ -64,13 +64,17 @@ capability:
   basis: n/a
   value: null
   confidence: high
-  note: 'A dataset is not ''capable'', so this axis is left unscored and the product is graded on adoption
-    alone. What the record otherwise recorded: GAIA is 466 questions across three levels requiring multi-step
-    tool use, live browsing and multimodal files, each with a single checkable answer. That is what makes
-    it the canonical end-to-end agentic-assistant benchmark, distinct from knowledge and code benchmarks,
-    and contamination-resistant. That is a quality caveat about the instrument rather than a capability
-    score: read as discriminative power, capability would decay as the field trains on a benchmark, which
-    makes it anti-correlated with adoption and unfit to blend with it. See #319.'
+  note: 'A dataset is not ''capable'' in the sense this axis measures elsewhere, so it is left unscored
+    and the product is graded on adoption alone. What the record previously asserted, kept as the quality
+    caveat it always was: GAIA is the canonical end-to-end agentic-assistant benchmark, 466 questions across
+    three levels requiring multi-step tool use, live browsing and multimodal files, each with a single checkable
+    answer. It is distinct from knowledge and code benchmarks, and contamination-resistant. This records
+    the absence of an agreed rubric, not a claim that an evaluation set has no measurable quality. #319
+    proposed reading the axis as discriminative power, whether a benchmark still separates frontier models,
+    and that reading was set aside on two grounds: it decays as the field trains on a benchmark, so it works
+    against adoption in the maturity blend, and the evidence assembled for it did not establish the bands
+    it assigned. A reading built on the artifact instead - coverage, grading validity, refresh cadence -
+    would not decay and has not been worked out. Until one exists this axis abstains. See #546.'
   last_verified: '2026-08-13'
   sources:
   - url: https://arxiv.org/abs/2311.12983
--- a/sources/scores/gaia.yaml
+++ b/sources/scores/gaia.yaml
@@ -60,12 +60,17 @@ adoption:
     http_status: 200
     content_sha256: 0bcfa21e0181419fa7c6c33c382a5e50ac59e8d343ab40a9ad4557a33377c4f0
 capability:
-  score: 5
-  basis: feature_matrix
-  value: 466 questions, 3 levels; multi-step tool use, live browsing, multi-modal files with one checkable answer
+  score: null
+  basis: n/a
+  value: null
   confidence: high
-  note: 'The canonical end-to-end agentic-assistant benchmark: distinct from knowledge and code benchmarks,
-    and contamination-resistant.'
+  note: 'A dataset is not ''capable'', so this axis is left unscored and the product is graded on adoption
+    alone. What the record otherwise recorded: GAIA is 466 questions across three levels requiring multi-step
+    tool use, live browsing and multimodal files, each with a single checkable answer. That is what makes
+    it the canonical end-to-end agentic-assistant benchmark, distinct from knowledge and code benchmarks,
+    and contamination-resistant. That is a quality caveat about the instrument rather than a capability
+    score: read as discriminative power, capability would decay as the field trains on a benchmark, which
+    makes it anti-correlated with adoption and unfit to blend with it. See #319.'
   last_verified: '2026-08-13'
   sources:
   - url: https://arxiv.org/abs/2311.12983
--- a/sources/scores/humanitys-last-exam.yaml
+++ b/sources/scores/humanitys-last-exam.yaml
@@ -61,12 +61,17 @@ capability:
   basis: n/a
   value: null
   confidence: high
-  note: 'A dataset is not ''capable'', so this axis is left unscored and the product is graded on adoption
-    alone. What the record otherwise recorded: 2,500 expert questions across more than 100 subjects, multimodal,
-    with a private held-out set, built as a successor to a saturated MMLU and MATH. GPQA is the nearest
-    narrower peer on difficulty. That is a quality caveat about the instrument rather than a capability
-    score: read as discriminative power, capability would decay as the field trains on a benchmark, which
-    makes it anti-correlated with adoption and unfit to blend with it. See #319.'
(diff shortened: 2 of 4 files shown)


ccerv1:
## Codex review: `fail`, two blocking findings, both fixed

### 1. "No published number moves" was too strong — corrected

The claim held at category level and not at product level. Verified against the corpus:

| product | maturity before | after |
|---|---|---|
| livebench | 3.8 | **3.0** |
| gaia | 3.8 | **3.0** |
| humanitys-last-exam | 3.8 | **3.0** |
| mt-bench | 2.8 | **2.0** |

`overall_score` is a published field, so four product scores fall. What is unchanged is the **category**: `benchmark_eval_data` stays Stage 4, Competitive Open Ecosystem, gaps `[resiliency]`, and no product crosses a tier threshold or the maturity bar in either direction. Nulling capability raises maturity only where capability sits below adoption, and all four here had capability above it, so all four move down.

The PR body's claim is corrected accordingly.

### 2. The notes asserted a methodological claim as settled — rewritten in 0961aeda

> The repeated rationale treats decay and anti-correlation with adoption as necessary consequences. Neither follows: capability could measure documented evaluation functions — coverage, reproducible scoring, multi-turn or tool-use evaluation, and question refresh — which need not decay as models improve. **LiveBench's retained monthly-rotation claim directly challenges the static-benchmark premise.**

That last point is sharp and correct. The corpus contained its own counterexample and the note was arguing past it. The notes now justify the abstention as **the absence of an agreed rubric**, name the two grounds on which the discriminative-power reading was set aside, and say an artifact-based reading has not been worked out. LiveBench's note names its rotation as the counterexample it is.

Also restored, having claimed to preserve the substance and then not quite: LiveBench's "among the most influential contamination-resistant benchmarks" and Humanity's Last Exam's "far harder and broader" comparison.

### Non-blocking, confirmed rather than actioned

- No `relative_to` anywhere in the corpus points at these four, so nothing dangles.
- `check_rubric` reads openness only; no repo warehouse model consumes capability (`product_evidence` calls it unbridged).
- The generated map notebook does consume capability for display, filtering and sorting, so **presentation changes** even though stage and gaps do not. Worth knowing before the next publish.

### A correction to the analysis that led here

The category declares **`weights: {adopt: 0.6, cap: 0.4}` as a top-level key**, which is what `serialize` reads via `cat.get("weights")`. I had looked under `scoring_recipe`, found nothing, and worked from the 0.5/0.5 fallback. The conclusion is unaffected — gsm8k needed capability ≥ 4 to hold Stage 4 under either weighting — but figures I quoted from the fallback were wrong, and 12 of 20 categories declare non-even weights.

Issue #536, opened 2026-09-10
[methodology] Zenodo is measurable and unread, and 17 scientific models fall through to stars


ccerv1:
The adoption axis reads a model's usage from Hugging Face. That is the right default and it covers the corpus as it stands, but it is a **channel** assumption wearing a **type** label, and a population of models that distributes elsewhere lands on the wrong instrument rather than on a wider error bar.

This is an instrument-coverage question, not a request to change any category's `weights`. It does not block #533.

## Where the assumption is written

Download bands are declared per product type, and the model type's `signal_preference` names the channel outright — `usage_volume`, "Hugging Face downloads, summed across the family's shipped SKUs". Nothing downstream asks where a given model is actually distributed from; the type decides.

Worth noting because it makes the fix cheap: **the model and software download bands are numerically identical** (`<10K` / `10K-100K` / `100K-1M` / `1M-10M` / `>10M`). The gap is not a miscalibrated scale. It is that one count gets read and another does not.

## What the channel actually looks like off the Hub

From the #533 sweep, 20 of 35 scientific-model candidates have no Hugging Face presence at all. Chasing where their weights really live:

**PyPI, via a package that fetches weights on first use — measurable today, and large.**

| Package | Downloads / month | Its model's HF count |
|---|---|---|
| `mace-torch` | 113,484 | none — no Hub repo |
| `fairchem-core` | 86,690 | 83 (`facebook/UMA`, gated) |
| `boltz` | 22,091 | none |
| `orb-models` | 12,043 | none |
| `climax` | 1,156 | none |

MACE reads as level 1 on the model instrument and level 3 on its actual install path. UMA reads level 1 against 83 gated Hub downloads while the library that loads it draws 86,690 a month. Same products, two levels apart, decided by which channel the instrument happens to look at.

**Zenodo — measurable, not currently read.** The API exposes `stats.downloads` and `stats.unique_downloads` per record. This is where a lot of academic weight releases and their DOIs sit.

**Cloud storage, institutional servers, Google Drive, Earth Engine — not measurable at all.** GraphCast's weights are on GCS, scGPT's on Drive. There is no count to read and no prospect of one, so these genuinely fall back to a proxy.

**GitHub releases are not the channel, despite the intuition.** I checked six of these repositories — MACE, OpenFold, RFdiffusion, ClimaX, Boltz, Orb — and across 65 releases they publish **zero** release assets between them. Weights are never attached there. Worth writing down so nobody builds that fetcher.

[...]

Issue #446, opened 2026-09-01
[methodology] No adoption instrument for Homebrew, app stores, or container pulls


ccerv1:
153 products route to stars, and the under-coverage detector shows the cost concentrated in products that ship as images and desktop builds: gvisor, ollama, searxng, promptfoo, perplexica, n8n, mistral-rs, uzu. n8n is the case docs/reference/adoption.md uses to establish the doctrine — 246M cumulative Docker pulls against a 393k/month npm figure.

There is no instrument for a Homebrew formula, an app-store listing, or a container image pull count. #163 already covers GitHub release-asset downloads and #305 covers Hugging Face Spaces; this issue is the remaining three channels.

Why it matters beyond the individual bands: any candidate ranking built on evidence of real use will systematically under-rate everything distributed outside PyPI, HF, npm and crates. The 2026-08-31 census had to record channel evidence by hand for exactly this population, and five of six #433 products banded on stars because no instrument existed for how they ship, not because they are obscure.

Bridge-building order of value, from the corpus: container pulls (largest measured discrepancies), Homebrew (has a public analytics API), app stores (hardest, likely stays reported_traction).

Issue #434, opened 2026-08-31
[category-proposal] compilers holds three populations, and one cannot score above 3 — but a split is the wrong fix


ccerv1:
`compilers` currently holds 26 products, and its own description admits three different kinds of
thing: software that "lowers, converts, quantizes, **or otherwise optimizes** trained models **and
tensor programs**". Classifying the roster against that sentence splits it cleanly, with nothing
left over:

| population | n | what it does | examples |
|---|---:|---|---|
| compiler / converter | 9 | lowers or converts a model graph to a target | `apache-tvm`, `iree`, `xla`, `tensorrt`, `mlc-llm`, `coremltools`, `hummingbird`, `tf2onnx`, `onnx-simplifier` |
| quantization / optimizer | 7 | transforms a trained model's weights | `nvidia-model-optimizer`, `intel-neural-compressor`, `torchao`, `bitsandbytes`, `llm-compressor`, `gptqmodel`, `aimet` |
| kernel library | 10 | compute primitives an engine calls at runtime | `onednn`, `arm-compute-library`, `cutlass`, `flash-attention`, `flashinfer`, `deepgemm`, `liger-kernel`, `sageattention`, `thunderkittens`, `composable-kernel` |

## The measurable problem

The three populations do not score alike, and the pattern is not about openness:

| population | n | capability mean | capability range | openness |
|---|---:|---:|---|---|
| compiler / converter | 9 | 3.67 | 2–5 | 2, 5 |
| quantization / optimizer | 7 | 3.43 | 3–4 | all 5 |
| kernel library | 10 | **2.60** | **2–3** | all 5 |

**No kernel library scores above 3.** Not one of the ten. That is hard to read as a fact about the
software: `flash-attention` is the reference implementation of IO-aware exact attention and
underpins essentially every modern transformer, and `cutlass` is the template layer a large share of
CUDA GEMM work is built on. They score low because they are being measured on a ladder built for
compilers, where breadth of target hardware and depth of the lowering pipeline are the
discriminators — axes on which a focused kernel library has nothing to say.

## Why a new category is NOT the answer — corrected

An earlier version of this issue argued that the kernel layer is invisible to the maturity
computation and should be split out. That reasoning was wrong in two ways, and the correction
matters more than the original observation.

**1. The layer looks uniformly open because the incumbent set is incomplete — not because the
missing half is closed.** Every kernel library in the corpus is openness 5, which an earlier draft
[...]

Issue #305, opened 2026-08-16
Hugging Face Spaces has no adoption instrument after its mis-declared artifact was removed


ccerv1:
`spaces` declared `gradio-app/gradio` as its GitHub artifact. That declaration was removed on 2026-08-14 because it was wrong in three ways at once:

1. Gradio is an **app SDK that runs anywhere** — Hugging Face Spaces is the hosting product, and the SDK is not it.
2. It is already the declared artifact of `gradio`, a **separate product on this map**, so one repository was being counted for two products.
3. Banding Spaces on Gradio's stars would have measured the popularity of one product and attributed it to another.

## Current state

The adoption axis is confirmed and dated, but Spaces publishes **no countable channel of its own** — no package registry, no download endpoint, no public usage figure. Nothing replaced the removed artifact, deliberately: repointing without evidence would repeat the original error.

## What would settle it

Any first-party countable signal for the hosting product itself. Candidates worth checking:

- a Hub API endpoint that returns a Spaces count or per-Space usage
- Hugging Face's own published platform statistics, if any carry Spaces separately
- a declared `signal_routing.yaml` route with `hand_authored: true` and a stated reason, which is how `reported_traction` and `active_users` already handle sources that have no machine endpoint

Until then the axis rests on a reported-traction reading rather than a measurement, and it says so.

Recorded during the deployment pass (PR #300).

PR #549, opened 2026-09-11
Name the axis the open ecosystem has never topped, so no category reports a stage with no gaps

ccerv1:
`compilers` reported **Stage 3 with no gaps at all**. A reader saw a category below the leading bar and was told nothing about what was missing.

## Why it happened

Not a bug — a dead zone in the constants. `_CAPABLE_MIN` and `_ADOPTED_MIN` are 4, `_MATURE_MIN` is 4.5, so a product must reach **5 on an axis** to be mature, and anything at exactly 4 and 4 clears both cutoffs while missing the bar.

**32 fully-open products across 13 categories sit at exactly 4/4**, so the zone is structural. It produces silence only where such a product is its category's *best*, which today is compilers alone — its best fully-open option is `coremltools` at 4/4.

The per-product drivers ask what holds the best **blended** product back. Where that product clears both cutoffs the honest answer is "nothing, on its own axes", and asserting a shortfall would be the knowingly false label the original code was right to refuse. The code comment said as much, and named the trigger for revisiting: *"when measurement gaps like this become common enough to name, introduce a gap for them deliberately."*

## What changed

Ask the **category-level** question instead: which axis has the open ecosystem never topped?

Compilers has three fully-open products at capability 5 — `apache-tvm`, `iree`, `xla` — and across 44 products **none at adoption 5**. The capability exists in the open and has not been adopted. That is `adoption`, and it is a fact about the category rather than an inference about one product.

## Two deliberate boundaries

- **An unmeasured axis yields no gap.** If no fully-open product records the axis at all it is unmeasured rather than deficient, so a category graded on adoption alone is never told it has a capability shortfall.
- **Both axes topped, never in the same product, still reports nothing.** The parts exist and nobody has assembled them. That is a real state this vocabulary cannot yet name, and faking it would be worse than the silence.

## This reverses a documented decision

`test_stage_1_to_3_carries_no_driver_gap_when_axes_clear_but_blend_misses_bar` pinned the old silence deliberately, using `benchmark_eval_data`'s shape: adoption 4, capability null. It is **rewritten rather than deleted**, and the rename says what changed.

The old reasoning answered the per-product question and was right on its own terms. The category-level question has an honest answer for that same shape: graded on adoption alone, the category needs an adoption 5 to reach maturity and has never had one. `adoption` is what is missing, and naming it beats the silence.

That also closes the live half of #547.

## Corpus effect

`compilers` gains `adoption`. **Every other category is unchanged** — verified by re-deriving `_stage_and_gaps` for all 20 with their declared weights.

## A caveat worth recording

This gap is substantially **instrument-limited**, not purely an ecosystem fact. `xla` ships inside the frameworks that embed it, so no download channel measures it; it is on `stars_fallback`, which caps at 3, and can therefore **never** reach maturity regardless of real use. `apache-tvm`'s own record says TVM is "predominantly built from source rather than installed from the wheel, so the package figure is a floor on real use rather than a measure of it."

The label is the most honest one available and silence is worse, but it reads as an ecosystem verdict when part of it is a measurement limit. Same thesis as #163, #446, #305 and #536. #434 is also relevant: if compilers holds three populations, this gap is partly an artifact of the mix.

## Test plan

- Four new tests: no category is silent at Stages 1-3; the fallback names the axis never topped; both fire when neither is topped; the fallback does not displace a real driver gap
- `build.goldens --check` — `tests/goldens/corpus.json` is current
- `uv run pytest -q` — **1901 passed**, no failures

### DIFF
--- a/build/serialize.py
+++ b/build/serialize.py
@@ -41,6 +41,7 @@ _ADOPTED_MIN = 4           # raw adoption below which an open option is "not ado
 # 5. Scored across all openness buckets -- the tier
 # describes the product -- while `mature` gates the same 4.5 bar on the fully-open bucket,
 # because only fully-open products advance a category's stage.
+_AXIS_TOP = 5              # a raw axis value the open ecosystem has "topped"
 _LEADING_MIN = _MATURE_MIN     # 4.5
 _STRONG_MIN = 4.0
 _STAGE_NAMES = {0: "Void", 1: "Open Experiments", 2: "Emerging Alternatives",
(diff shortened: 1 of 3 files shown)


ccerv1:
## Codex review: `fail`, two blocking. Both fixed in 9e5a57a7.

### 1. A new claim under an old label — and the comment above it said so

> Below-cutoff performance of the best product and absence of an ecosystem-wide maximum are different claims... **the unchanged surrounding code comment also directly contradicts the fallback.**

Correct, and the contradiction was mine: the comment saying *"asserting an adoption shortfall where adoption clears its cutoff would be a knowingly false label"* sat directly above code that now does exactly that. `docs/reference/gap-analysis.md` documented the old contract in full, unamended, and even prescribed the alternative — *"add a gap for that state deliberately rather than reusing an inaccurate one."*

**Resolved by widening the contract rather than minting a seventh gap type.** A driver gap now means its axis is short **for the category**, true two ways: the best fully-open product is below the cutoff, or no fully-open product has reached the top of that axis. Both readings answer the question a reader is actually asking — which axis is holding this category back — and a new chip buys precision at the cost of a vocabulary nobody asked to learn.

The widening is recorded in `gap-analysis.md` as a deliberate revision of #318's contract, with the two states that still carry no driver gap spelled out, and the code comment now carries the new position and says what changed.

### 2. A real defect in the population

> `open_rows` excludes products with missing adoption... incorrectly receives a capability fallback, despite having recorded capability 5.

Right. A product with no adoption is rightly excluded from the stage arithmetic, since maturity is anchored on adoption — but its recorded capability is still a fact about what the open ecosystem has reached. The fallback now reads every fully-open product. Latent today (no fully-open corpus product lacks an adoption band) and wrong regardless. Test added.

### Non-blocking, all actioned

- **32 products across 12 categories, not 13.** My count was wrong; corrected in the docstring, the doc and the PR body.
- **Split-axis test added.** Worth noting the first draft of it was wrong: a product topping capability at 5 with adoption 2 is *below* the adoption cutoff, so the per-product driver fires and the fallback never runs. It needs a 4/4 best product with the two axis-toppers beneath it.
- **`test_a_category_is_never_silent_at_stages_one_to_three` renamed** — its universal claim contradicted the split-axis exception this same PR documents.

### Re-verified

Corpus effect unchanged: `compilers` gains `adoption`, every other category identical, checked against each category's declared **top-level** weights. `tests/test_serialize.py` 65 passed, `validate` 0 errors. Full suite running.

PR #550, opened 2026-09-11
Stop storing reproduction counts the repo already computes

### LINKED ISSUE #545: derived_from attestations have drifted in 15 of 20 categories, and nothing compares them to the live recompute
Every category's `scoring_recipe.derived_from` block carries a dated attestation: the method, how many scores reproduced, how many were deferred, the date it was verified, and the checker that did it. `build/check_rubric.py` recomputes the same figures live on every run.

**Nothing compares the two.** So the declared numbers drift silently as products are added, moved or re-scored, and 15 of 20 categories have.

| category | declared | live | verified_on |
|---|---|---|---|
| `inference_code` | 10/20 | 39/39 | 2026-07-30 |
| `ui_api` | 30/43 | 49/49 | 2026-07-30 |
| `evaluation_code` | 16/26 | 32/32 | 2026-07-30 |
| `finetuning_code` | 22/27 | 32/32 | 2026-07-30 |
| `base_pretrained` | 26/26 | 32/32 | 2026-07-28 |
| `ml_frameworks` | 20/22 | 25/25 | 2026-07-30 |
| `finetuned_chat` | 39/39 | 41/41 | 2026-07-29 |
| `benchmark_eval_data` | 25/27 (2 def) | 30/30 (2 def) | 2026-08-12 |
| `training_synthetic_datasets` | 37/38 (1 def) | 37/37 (1 def) | 2026-08-11 |
| `dataset_processing_tools` | 20/21 (0 def) | 20/20 (1 def) | 2026-08-12 |
| `edge_hardware` | 19/20 (1 def) | 19/19 (1 def) | 2026-08-12 |
| `storage` | 39/39 | 40/40 | 2026-09-02 |
| `orchestration_agents` | 51/51 | 52/52 | 2026-08-12 |
| `telemetry_observability` | 33/33 | 32/32 | 2026-09-02 |
| `embeddings_retrieval` | *(absent)* | 35/35 | 2026-09-11 |

Five are in sync.

## Why it matters, and which direction it fails

The drift understates, which is the bad direction. `inference_code` declares **10 of 20 reproduced**, so a reader checking whether the rubric is trustworthy sees half the scores failing to reproduce. Live, all 39 reproduce. The attestation makes the corpus look considerably less sound than it is, and it is the artifact a reader reaches for precisely when they are auditing that question.

`embeddings_retrieval` shows the other shape: the category landed on 2026-09-11 with `verified_on` set and the counts never filled in.

## The fix has two halves, and only one is mechanical

1. **The counts are derivable.** `check_rubric` already computes reproduced, total and deferred per category. Rewriting the declared numbers from its output is safe and needs no judgement.

2. **`verified_on` must not be bumped mechanically.** It attests that a person verified the method against the hand-authored scores on that date. Deriving it from a run would assert a verification that did not happen, which is the same rule `docs/guides/freshness.md` sets for `last_verified` — a date records when something was confirmed and is never computed from a side effect.

So: recompute the counts, leave every `verified_on` alone, and note in the block that the counts are maintained against the checker while the date records the last human pass.

## The durable part

Add a gate comparing each declared `derived_from` count to what `check_rubric` computes, failing on drift. Without it this recurs the next time a product changes category — which is exactly how `orchestration_agents` and `telemetry_observability` drifted this week, in opposite directions, from a single move in #544.

Found while reviewing #544, which touched both of those categories and deliberately left the stale attestations alone rather than rewriting a dated claim without a re-run.

ccerv1:
Closes #545, by **removing** the numbers rather than correcting them.

## What was wrong

`scoring_recipe.derived_from` carried `scores_reproduced`, `scores_total` and `scores_deferred` beside the `verified_on` date. `build/check_rubric.py` recomputes all three on every run, so the stored copies were redundant the day they were written. Nothing compared them, and **15 of 20 categories had drifted**:

| category | declared | live |
|---|---|---|
| `inference_code` | 10/20 | 39/39 |
| `ui_api` | 30/43 | 49/49 |
| `evaluation_code` | 16/26 | 32/32 |
| `embeddings_retrieval` | *(none at all)* | 35/35 |

They understated, which is the damaging direction: that block is what a reader consults when asking whether the rubric can be trusted, and it made the corpus look far less sound than it is. **680 of 680 scores reproduce today, across all 20 categories.**

## Why correcting them was the wrong fix

The first version of this change corrected the counts in place and added a gate. It was worse than the drift, and it briefly shipped on this branch.

A count recorded beside `verified_on` reads as *"this many reproduced when a person last checked"*. Writing today's `39/39` next to a `2026-07-30` date asserts a verification of 39 products that never happened — nineteen of them did not exist then. And the drift was carrying real information: **how much of a category has been added since anyone last verified the method by hand.** Overwriting it destroyed that signal and replaced it with a false claim.

## What this does instead

- The three count keys are **gone from all 20 categories**.
- `build/check_recipe.py` fails any category that reintroduces one.
- `derived_from` keeps only what nothing can compute: the method, the date a person checked it, and which checker they used.
- Run `build.check_rubric` for the live figures; it prints them every run.

## Two documents taught the old behaviour

- `docs/workflows/promote-category.md` told the editor to update the counts on every promotion — **the mechanism that generated the drift.** It now says not to, and says why.
- `docs/schemas/category.schema.json` described `derived_from` as ungated, which is no longer true.

## The general rule

Any number the repo states about itself should be computed and compared, never typed — and there should be **as few such statements as the work allows**. A number that is not stated cannot drift, and a gate on a redundant number is still a maintenance surface.

This is one instance of a recurring class. The same day produced three others: the PLATFORM MIRROR banner contradicting `assets.yaml` (#517), an issue instructing an edit to a file retired a month earlier (#303), and `category.schema.json` still stating "522 head slugs across 18 published rosters" against a live 685 across 20. The August foundation review put it the same way — the logic and arithmetic hold; the self-description drifts.

Where a number must appear in prose, date it. "522 ... as of 2026-08-18" ages visibly. A structured field that looks live does not.

## Test plan

- Five new tests, including one asserting the whole corpus stores none of the keys
- `build.check_recipe` — 0 failures
- `build.check_rubric` — clean, all 20 categories reproducing
- `uv run pytest -q` — **1901 passed, 1 skipped** (the corpus golden, which goes stale on any `sources/` change and is rewritten by `regenerate.yml`, per its own notice)


ccerv1:
## Codex review: `fail`, one blocking. Fixed.

### The removal missed the authoring path

> The authoring instructions still require the forbidden data. **Step 9 explicitly says to record counts in the category's `derived_from` before sharing a ladder**; both shared-rubric comments reinforce that instruction.

Correct, and it would have been self-defeating: an editor following `skills/build-rubric/SKILL.md` would reintroduce the fields and fail the gate this branch adds. I had updated `promote-category.md` and the schema and stopped there. Three more places fixed — the skill's step 9, and the header comments in `sources/rubrics/software.yaml` and `sources/rubrics/dataset.yaml`. Each now says to set `verified_on` to the date checked and read the counts off a run.

### A claim of mine that was too strong

> Count drift alone does not measure additions since verification: **removals, roster transfers, score corrections and ladder changes also affect it.** If verification coverage matters operationally, record a checked revision or explicit reviewed roster rather than infer it from aggregate drift.

Right. I argued the drift was carrying signal about how much of a category postdates the last human pass, and an aggregate count cannot carry that. The docstring now says so, states the honest cost of deleting rather than correcting — the historical result is no longer readable from the file, only from git history — and points at the mechanism that would actually record verification coverage.

### Confirmed rather than actioned

- **No executable consumer** of the three fields anywhere in the repo; the warehouse serializer exports selected recipe fields, not this attestation. External consumers outside the checkout remain unverified.
- **19 files, not 20.** `embeddings_retrieval` already carried none of the keys — it declared a `verified_on` with nothing beside it, which is what made it the example in the issue.
- **`finetuned_chat.scores_corrected` is deliberately retained.** It records historical editing activity rather than a live reproduction count, so it is not in the forbidden set.
- `check_recipe`'s default runner skips recipes with resolution errors or no variants before reaching the new check. The whole-corpus test added here catches forbidden fields in those categories regardless.

PR #498, opened 2026-09-05
identity: adopt 51 first-party Hugging Face handles (#483)

ccerv1:
## What

Two-pass review of the 69 Hugging Face handle proposals in #483. A handle asserts that an organization **publishes under** an account, so I adopt the ones where the account is the org's own official account and hold the rest — dataset hosting, collaboration, benchmark maintenance, or a mirror is not ownership. **Adds 51 handle rows** to `sources/org_handles.yaml`; nothing else changes.

**Coverage:** Hugging Face org coverage rises **2/62 → 51/62** with org-resolution **precision unchanged at 1.000**. The eleven still-uncovered orgs are the held cases below, plus `princeton-nlp-openai` — intentionally uncovered because the shared `princeton-nlp` account belongs to the canonical `princeton-nlp` org (see the shared-account note under Adopted). `github` (211/299) and `homepage` (6/27) coverage unchanged. The 52→51 step after the Princeton correction is a data-quality improvement, not a regression: the metric should never turn a shared account into a second org's ownership claim.

## Adopted (51)

The obvious first-party namespaces — `Qwen`, `Anthropic`, `deepseek-ai`, `google`, `meta-llama`, `microsoft`, `nvidia`, `openai`, `mistralai`, `CohereLabs`, `ByteDance-Seed`, `ibm-granite`, `MiniMaxAI`, `moonshotai`, `openbmb`, `PleIAs`, `swiss-ai`, `thinkingmachines`, `01-ai`, `ai21labs`, `LLM360`, `bigcode`, `HuggingFaceFW/H4/TB`, `google-research-datasets`, `ministere-culture` — plus the benchmark/lab accounts that are unambiguously their org's own: `livebench`, `livecodebench`, `TIGER-Lab`, `lmsys`, `nuprl`, `NousResearch`, `open-thoughts`, `open-web-math`, `oscar-corpus`, `OpenAssistant`, `openguardrails`, `uonlp`, `teknium`, `TinyLlama`, `mlfoundations`, `XiaomiMiMo`, `zentropi-ai`, `RWKV`.

Included despite a name that doesn't string-match the org, because the account is unmistakably first-party (the same case as `Qwen ← alibaba-cloud`): `tiiuae ← technology-innovation-institute`, `togethercomputer ← together-ai`, `internlm ← shanghai-ai-laboratory`, `zai-org ← zhipu-z-ai`, `BSC-LT ← barcelona-supercomputing-center`.

[...]

PR #553, opened 2026-09-11
Guard the identity coverage pin in both directions

### LINKED ISSUE #493: Guard the identity coverage baseline against silent lowering
\`tests/fixtures/identity_coverage_baseline.json\` pins per-route handle coverage (github, huggingface, homepage_domain) as a ratchet for \`build/identity_eval.py\`. \`--write-coverage-baseline\` can rewrite it in either direction, and nothing compares the committed values to the previous ones.

Add a check, in the eval or as a test, that compares the baseline to the merge-base copy (\`git show origin/main:tests/fixtures/identity_coverage_baseline.json\`) and fails if any route's ratio went down unless the file carries a \`lowered_because\` key explaining why. Report explicitly when the base copy is unreadable rather than passing vacuously. The merge-base idiom already exists in \`build/assets.py\` and \`build/observation_snapshot.py\`.

Not urgent: the file is codeowned and a downward re-pin is preceded by a red run that printed the drop.

### LINKED ISSUE #506: Coverage ratchet goes slack when coverage rises and the pin is not moved
`tests/fixtures/identity_coverage_baseline.json` is a ratchet: `build/identity_eval.py` fails
when a route's live coverage falls **below** the pinned ratio. Nothing fails when coverage rises
above it, so a gain that forgets to re-pin leaves the gate permanently loose.

## How it showed up

#498 adopted 51 first-party Hugging Face handles and touched only `sources/org_handles.yaml`.
The baseline stayed at `2/61` while actual coverage became `51/62`. Every run after that passed —
correctly, by the rule as written — against a floor 79 points below reality. The gate would have
accepted the deletion of 49 handles without a word.

Caught by hand on 2026-09-06, the day before `identity-eval`'s first scheduled run. #505 re-pins
the data; this issue is the mechanism.

## Why the existing guards miss it

- The ratchet itself is one-directional by design.
- #493 proposes guarding against a silent **lowering** of the baseline. That is the opposite
  case and would not have fired here.
- The sentinel (#459) only opens an issue on a **red** run. This failure is green.

So the slack is invisible from every direction we currently look.

## What would fix it

A check that fails when live coverage exceeds the pinned baseline by more than a small margin,
with a message saying to re-pin. That converts "someone remembered" into "the build insisted".
Two decisions to make when implementing:

- **The margin.** Zero would fail on every single-handle gain, which is noise. Some tolerance —
  a few percentage points, or an absolute count — keeps it actionable.
- **Where it runs.** Scoring runs already compute coverage, so this is cheap to add there; the
  question is whether a PR that legitimately adds handles should fail until it re-pins in the
  same PR. Failing is probably right: re-pinning belongs in the commit that earns it, which is
  exactly what #498 did not do.

Worth pairing with #493, since both are about the baseline being trustworthy rather than merely
present, and a single implementation could hold both directions.

ccerv1:
## Summary

`tests/fixtures/identity_coverage_baseline.json` only ever failed when live coverage fell below the pin. This closes both holes #493 and #506 describe with one comparison over one file.

- **Lowering (#493).** A test reads the baseline as of the merge base with `origin/main` (`git show <sha>:tests/fixtures/identity_coverage_baseline.json`) and fails when any route's pinned ratio went down, or a pinned route vanished, unless the file carries a `lowered_because` that is new or changed in the same PR. An inherited note excuses nothing, and a fresh note with no route lowered is reported as stale. A pin rewritten to `0/0` counts as a lowering (a pin is a written claim of ratio 0, unlike a live route with no orgs). A pin that vanishes is never excusable: an unpinned route is skipped by both live gates, so an explained deletion would switch the ratchet off for that route and every later PR would inherit the silence. The committed file must pin every route (an existing test), the weekly scoring run exits 2 on a baseline file that pins fewer than every route, `{}` included (only a missing file is the lenient no-ratchet case), and the remedy is to pin to `0/0` with a note. When the merge-base copy cannot be read, the test prints git's own words and the path it tried: a hard fail under `GITHUB_ACTIONS`, a visible skip elsewhere.
- **Slack (#506).** Live coverage more than `COVERAGE_MARGIN_POINTS` (5 percentage points) above the pin fails, in both the pytest gate and the weekly `identity_eval` run, with a message telling the author to re-pin with `--write-coverage-baseline` in the same PR. Five points is about a dozen GitHub handles or three homepage ones, so a single-handle PR passes and a #498-sized gap (79 points) cannot.
- **`lowered_because`** is the only non-route key `load_coverage_baseline` accepts; it must be a non-empty string. `--write-coverage-baseline` drops any prior note and says so; `--lowered-because` writes a new one.
- **The re-pin.** The new gate fired on its own PR: live coverage was already above the pin on two routes. Re-pinned in its own commit, after the code commit, so the history reads as "gate found it, fix follows".

| route | pinned before | live now |
|---|---|---|
| github | 211/328 | 237/350 |
| huggingface | 51/81 | 72/103 |
| homepage_domain | 7/50 | 7/50 |

Closes #493. Closes #506.

## Test plan

- `uv run pytest -q` passes on the branch
- `uv run python -m build.validate` reports 0 errors
- New tests cover: lowering without a note fails, with a fresh note passes, inherited note is inert, stale note fails, vanished route is a lowering, `0/0` pin is a lowering, margin boundary at exactly +5.0 passes and above fails, unreadable merge base prints the reason
- Codex reviewed round 1 and found one hole (the `0/0` pin); round 2 fixed it. The loop's reviewer step could not read round 2 out of the build clone, so a second Codex pass was run by hand on the PR: it confirmed the `0/0` fix and found the vanished-pin case, closed in the third and fourth commits on both the pytest gate and the scoring path, and re-checked. Known limit, accepted: note freshness is string inequality against the merge base, so a reworded old reason counts as fresh; the code does not judge the explanation's substance.

### DIFF
--- a/build/identity_eval.py
+++ b/build/identity_eval.py
@@ -2043,7 +2043,10 @@ def main(argv: list[str] | None = None) -> int:
     # here forever. The pytest gate requires every route in the committed file; the scoring
     # path refuses the same shape so the weekly run cannot report green on a route it never
     # checked. A missing file (no ratchet at all, an older tree) is still the lenient case.
-    unpinned = [route for route, _label in ORG_ROUTE_LABELS if route not in baseline] if baseline else []
+    unpinned = (
+        [route for route, _label in ORG_ROUTE_LABELS if route not in baseline]
+        if COVERAGE_BASELINE_PATH.exists() else []  # the FILE decides, not the dict: `{}` is a file with every pin deleted
+    )
     if unpinned:
         print(
             f"[FAIL] {_repo_path(COVERAGE_BASELINE_PATH)} pins no value for {', '.join(unpinned)}; "
--- a/build/identity_eval.py
+++ b/build/identity_eval.py
@@ -1135,11 +1135,15 @@ def coverage_lowered(
 
     A route's pin may go down only with a `lowered_because` that is new or changed in this
     PR; one inherited verbatim from the merge base explained an earlier lowering and excuses
-    nothing now. A pin that vanished (the route was in `before`, not in `current`) is a
-    lowering to nothing, and so is a pin rewritten to `0/0` (see `_pin_lowered`). A fresh
-    `lowered_because` with no route lowered is stale and is reported too: an explanation
-    that explains nothing is how a guard rots.
+    nothing now. A pin rewritten to `0/0` is a lowering (see `_pin_lowered`). A pin that
+    VANISHED (the route was in `before`, not in `current`) is not a lowering that a note can
+    excuse: an unpinned route is skipped by both live-coverage gates, so an explained deletion
+    would switch the ratchet off for that route and every later PR would inherit the silence.
+    Lower it, to `0/0` if need be, and keep the key. A fresh `lowered_because` with no route
+    lowered is stale and is reported too: an explanation that explains nothing is how a guard
+    rots.
     """
+    vanished: list[str] = []
     lowered: list[str] = []
     for route, _label in ORG_ROUTE_LABELS:
         prior = before.get(route)
@@ -1147,10 +1151,19 @@ def coverage_lowered(
             continue
         now = current.get(route)
         if now is None:
-            lowered.append(f"{route}: pinned {prior[0]}/{prior[1]} at the merge base, unpinned now")
+            vanished.append(
+                f"{route}: pinned {prior[0]}/{prior[1]} at the merge base, unpinned now -- a route "
+                f"may be lowered with a {LOWERED_BECAUSE_KEY!r}, never unpinned; pin it (0/0 if need be)"
+            )
         elif _pin_lowered(now, prior):
             lowered.append(f"{route}: pin lowered from {prior[0]}/{prior[1]} to {now[0]}/{now[1]}")
     fresh_note = lowered_because is not None and lowered_because != before_lowered_because
+    if vanished:
+        needs_note = lowered and not fresh_note
+        return vanished + lowered + (
+            [f"a pin may only go down with a {LOWERED_BECAUSE_KEY!r} written or changed in this PR"]
+            if needs_note else []
+        )
     if lowered and not fresh_note:
         return lowered + [
             f"a pin may only go down with a {LOWERED_BECAUSE_KEY!r} written or changed in this PR"
--- a/build/identity_eval.py
+++ b/build/identity_eval.py
@@ -1113,6 +1113,18 @@ def coverage_stale(
     return failures
 
 
+def _pin_lowered(now: tuple[int, int], prior: tuple[int, int]) -> bool:
+    """Is the pin `now` strictly below the pin `prior`? Cross-multiplied like
+    `_below_baseline`, but this compares two PINS, not live coverage to a pin, so a
+    denominator of 0 on either side is read as the ratio 0 it pins rather than as "no orgs
+    to cover". Without that, rewriting 211/328 to 0/0 would pass vacuously, and a re-pin
+    after every eligible org was removed could erase positive coverage unexplained.
+    """
+    now_n, now_d = now if now[1] else (0, 1)
+    prior_n, prior_d = prior if prior[1] else (0, 1)
+    return now_n * prior_d < prior_n * now_d
+
+
 def coverage_lowered(
     current: dict[str, tuple[int, int]],
     before: dict[str, tuple[int, int]],
@@ -1124,8 +1136,9 @@ def coverage_lowered(
     A route's pin may go down only with a `lowered_because` that is new or changed in this
     PR; one inherited verbatim from the merge base explained an earlier lowering and excuses
     nothing now. A pin that vanished (the route was in `before`, not in `current`) is a
-    lowering to nothing. A fresh `lowered_because` with no route lowered is stale and is
-    reported too: an explanation that explains nothing is how a guard rots.
+    lowering to nothing, and so is a pin rewritten to `0/0` (see `_pin_lowered`). A fresh
+    `lowered_because` with no route lowered is stale and is reported too: an explanation
+    that explains nothing is how a guard rots.
     """
     lowered: list[str] = []
     for route, _label in ORG_ROUTE_LABELS:
@@ -1135,7 +1148,7 @@ def coverage_lowered(
         now = current.get(route)
         if now is None:
             lowered.append(f"{route}: pinned {prior[0]}/{prior[1]} at the merge base, unpinned now")
-        elif _below_baseline(now, prior):
+        elif _pin_lowered(now, prior):
             lowered.append(f"{route}: pin lowered from {prior[0]}/{prior[1]} to {now[0]}/{now[1]}")
     fresh_note = lowered_because is not None and lowered_because != before_lowered_because
     if lowered and not fresh_note:
--- a/build/identity_eval.py
+++ b/build/identity_eval.py
@@ -2038,6 +2038,18 @@ def main(argv: list[str] | None = None) -> int:
     except CoverageBaselineInvalid as exc:
         print(f"[FAIL] {exc}")
         return 2
+    # A baseline file that pins some routes and not others is not a partial ratchet, it is a
+    # switched-off one: both live gates skip an unpinned route, so a deleted pin would pass
+    # here forever. The pytest gate requires every route in the committed file; the scoring
+    # path refuses the same shape so the weekly run cannot report green on a route it never
+    # checked. A missing file (no ratchet at all, an older tree) is still the lenient case.
+    unpinned = [route for route, _label in ORG_ROUTE_LABELS if route not in baseline] if baseline else []
+    if unpinned:
+        print(
+            f"[FAIL] {_repo_path(COVERAGE_BASELINE_PATH)} pins no value for {', '.join(unpinned)}; "
+            f"every route must be pinned (0/0 if need be) or the ratchet is off for it"
+        )
+        return 2
     print("")
     for line in coverage_lines(coverage, baseline):
         print(line)
--- a/build/identity_eval.py
+++ b/build/identity_eval.py
@@ -176,8 +176,33 @@ falls BELOW its pinned ratio. So coverage can only go up. `--write-coverage-base
 rewrites that file from the live corpus; it is a deliberate act in its own commit, never
 something a failing run does for itself, or the ratchet would ratchet nothing.
 
+A ratchet that only fails downward has two holes, and the pin is guarded on both sides:
+
+- **The pin cannot be quietly lowered** (#493). `tests/test_identity_eval.py` reads the
+  file as of `git merge-base HEAD origin/main` (`merge_base_coverage_baseline`) and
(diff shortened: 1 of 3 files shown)


ccerv1:
## What I did
- files touched:
  - `build/identity_eval.py` — contract docstring (the two holes and how each is closed); `COVERAGE_MARGIN_POINTS = 5` and `LOWERED_BECAUSE_KEY` next to `COVERAGE_BASELINE_PATH`; `_parse_coverage_baseline(text, where)` shared by `load_coverage_baseline`, new `load_lowered_because`, and new `merge_base_coverage_baseline(base, rel) -> (routes | None, note | None, where)`; `write_coverage_baseline(..., lowered_because=None)` never carries a previous note forward; `_above_baseline_by_more_than` (integer cross-multiplication, mirrors `_below_baseline`), `coverage_stale`, `coverage_lowered`; `coverage_lines` gains a `STALE BASELINE` status; `main()` gets `--lowered-because` (only with `--write-coverage-baseline`), prints when it drops a previous note, and exits 1 with a re-pin instruction when the corpus has outrun the pin.
  - `tests/test_identity_eval.py` — `test_the_live_corpus_is_within_the_margin_above_the_committed_baseline` (the #506 gate over `REAL_TRUTH`), `test_the_committed_baseline_was_not_lowered_against_the_merge_base` (the #493 gate; fails under `GITHUB_ACTIONS` / skips with reason elsewhere when the merge-base copy is unreadable, via `_lowering_check_or_explain`), plus 22 synthetic tests over `coverage_lowered`, `coverage_stale`, `merge_base_coverage_baseline` (unresolvable ref, absent file, `base="HEAD"`), the `lowered_because` parser rules, `write_coverage_baseline` dropping/writing the note, the CLI flag, and the `main()` stale exit. Pre-existing `test_main_exits_zero_at_the_baseline` now pins the live github ratio instead of 0/299 (0/299 is correctly stale under the new gate).
  - `tests/fixtures/identity_coverage_baseline.json` — re-pinned in its own commit: github 211/328 -> 237/350, huggingface 51/81 -> 72/103, homepage_domain 7/50 unchanged.
- commits on `loop/coverage-ratchet-both-directions-2026-09-11-a1`: `0d73e1f2 test(identity): guard the coverage pin in both directions`, `03a12163 chore(identity): re-pin handle coverage to the live corpus`.
- commands run:
  - `uv run python -m build.validate` -> `0 error(s), 2 warning(s)` (the two pre-existing model_families overlap warnings)
  - `uv run pytest -q -p no:cacheprovider -x` -> `1930 passed, 1 skipped in 1552.37s`; the skip is the merge-base test in this remote-less clone, reason printed: "cannot compare the committed tests/fixtures/identity_coverage_baseline.json to its merge-base copy: `git merge-base HEAD origin/main` failed: fatal: Not a valid object name origin/main. Nothing was checked for a lowered pin."
  - `uv run python -m build.identity_eval --write-coverage-baseline` (the re-pin)
  - End-to-end probe in a scratch clone with `refs/remotes/origin/main` pointed at `main` and `GITHUB_ACTIONS=1`: raised pin passes; lowered pin without note fails naming `github: pin lowered from 211/328 to 100/350`; lowered pin with fresh note passes; fresh note with nothing lowered fails as stale; deleting `origin/main` makes the test FAIL (not skip) with the git error in the message.

## Why
- decisions:
  - **Margin = 5 percentage points** (`COVERAGE_MARGIN_POINTS`). The ratchet compares ratios by cross-multiplication, so the margin is in the same unit and uses the same integer arithmetic (`100*(live_n*base_d - base_n*live_d) > M*live_d*base_d`). One handle is 0.3 points on github (350 orgs) and 2 points on homepage_domain (50 orgs), so zero fails every single-handle PR on the small route (#506's noise objection); 5 is about a dozen github handles or three homepage ones, a batch rather than a typo, and the 79-point gap #506 found clears any sane margin. Rejected an absolute org count (means different things per route) and "touched `sources/org_handles.yaml`" (coverage also moves when products or artifacts change the denominator, and it ties the gate to a filename).
  - **Where each check runs.** Upper bound (#506): both. The pytest gate over the live corpus is what binds, because `validate.yml` runs on every PR and the PR that earns the gain must re-pin; the scoring path prints the same `STALE BASELINE` / `[FAIL]` vocabulary so the weekly report agrees with CI. Trade-off: a weekly `--from-warehouse` run could go red on a pin main already carries, but the pytest gate keeps main from reaching that state except through a merge race, and a red weekly run then names the exact fix. Lowering (#493): `tests/` only. It needs git history; the weekly run checks out `main` where the merge base is HEAD and the comparison is trivially equal, and the scoring path should not shell out to git. `validate.yml` uses `fetch-depth: 0` so `origin/main` is present on PR checkouts.
  - **Unreadable merge base**: fail under `GITHUB_ACTIONS` (a CI checkout that cannot see `origin/main` is a CI bug and must not pass), `pytest.skip` with the full git error elsewhere (a clone with no remote has nothing to compare, and a skip is visible in the summary, not a pass). The message names both sides: the committed path and the `git show <sha>:<path>` spec or the git error.
  - **`lowered_because` semantics**: excuses a lowering only when it is new or changed relative to the merge-base copy. A note inherited verbatim excuses nothing (otherwise the last lowering's note becomes a standing permission, the rot the criterion warns about); a fresh note with no route lowered is reported as stale and fails. An inherited note with nothing lowered is inert and passes, so PRs after a lowering merges are not all broken. `write_coverage_baseline` drops any prior note and prints that it did; `--lowered-because` writes a new one.
  - A pin that vanished (route present at merge base, absent now) counts as lowered. A pinned denominator of 0 reads as ratio 0 for the stale check, matching `_below_baseline`.
  - Re-pin order: code commit first, re-pin second, so the history reads as "gate found it, fix follows". The first commit alone fails the live-corpus stale test on huggingface, which is the intended behavior.
- alternatives rejected: returning a bare `None` from `merge_base_coverage_baseline` like `merge_base_assets` (loses git's reason, which #493 wants printed); making `load_coverage_baseline` return a tuple (breaks every caller; a separate `load_lowered_because` keeps the route dict shape); running the lowering check in `main()` (no git history on the weekly run, and a fixture run should not depend on the checkout's refs).

## Risks / unsure about
- The merge-base test is a skip in any clone without `origin/main` (including this harness clone and forks that name their remote differently). In CI it is a hard fail. If the harness's real repo checkout names the upstream something other than `origin`, the test will skip locally forever and only the CI run binds.
- Inherited `lowered_because` is deliberately inert rather than failing; if the reviewer wants every later PR to clean it up, flip the `fresh_note and not lowered` branch to also fire when `lowered_because == before_lowered_because`.
- The full suite takes ~26 min serially here; CI uses `-n auto`. Nothing in the new tests is marked `serial`; `merge_base_coverage_baseline(base="HEAD")` reads only from git objects, so it is xdist-safe.
- Exact-margin boundary: +5.0 points passes, anything strictly above fails (tested with 220/400 vs 221/400 against 200/400).

## Response to round N-1
|---|---|---|---|---|
| n/a (round 1) | | | | |

ccerv1:
## What I did
- files touched:
  - `build/identity_eval.py` — new `_pin_lowered(now, prior)` comparator used by `coverage_lowered` in place of `_below_baseline`. It cross-multiplies like the rest of the ratchet, but reads a denominator of 0 on EITHER side as the ratio 0 it pins (`(0, 1)`), because it compares two pins, not live coverage to a pin. `coverage_lowered`'s docstring now names `0/0` as a lowering.
  - `tests/test_identity_eval.py` — `test_coverage_lowered_reads_a_zero_denominator_pin_as_zero`: 211/328 -> 0/0 without a note yields `github: pin lowered from 211/328 to 0/0` plus the "needs a fresh note" line; the same with a fresh `lowered_because` passes; 0/0 -> 0/0, 0/0 -> 0/5, 0/0 -> 3/5 are not lowerings; 0/0 -> 3/5 with a fresh note is reported as stale.
- commits on `loop/coverage-ratchet-both-directions-2026-09-11-a1`: `6b59ac4e test(identity): read a zero-denominator pin as zero when checking for a lowering` on top of round 1's `0d73e1f2` and `03a12163`.
  - `uv run pytest -q -p no:cacheprovider tests/test_identity_eval.py -k "coverage or baseline or lowered or stale"` -> `44 passed, 1 skipped in 8.83s`
  - `uv run python -m build.validate` -> `0 error(s), 2 warning(s)` (the two pre-existing model_families overlap warnings)
  - `uv run pytest -q -p no:cacheprovider -n auto` -> `1931 passed, 1 skipped in 380.90s`; the skip is the merge-base test in this remote-less clone (no `origin/main`), which prints the git error as its reason.

## Why
- decisions:
  - A separate comparator rather than a special case inside `_below_baseline`. `_below_baseline` is the pre-existing live-vs-pin ratchet whose vacuous pass on a live 0/0 is documented and intended (no orgs on the route means nothing to cover); the plan says to keep the cross-multiplication contract. The pin-vs-pin comparison has different semantics: a pin of 0/0 is a written claim of ratio 0, so it must compare as 0. Putting the difference in a named function with a docstring keeps the two readings from being confused again.
  - Kept the ratchet's exact-ratio arithmetic; `(0, 1)` substitution is the only change and only when a denominator is 0.
- alternatives rejected:
  - Failing `coverage_ratchet` / `coverage_stale` when live is 0/0 against a positive pin (the reviewer's "both live-coverage checks also pass on an empty route"). A live 0/0 means the corpus has no eligible orgs on the route; coverage is undefined there, not low, and `_below_baseline`'s docstring defends that pass on purpose. The hole the reviewer prevents is closed at the pin: re-pinning to 0/0 now needs a fresh `lowered_because`, so positive coverage cannot be erased silently. Changing the pre-existing live comparator is out of the plan's scope and would fail the weekly report on a route that has legitimately emptied.
  - Rejecting a 0-denominator pin in `load_coverage_baseline`. The file format allows it (an unpinned-yet route pins 0/0 by design, and `test_coverage_stale_passes_an_empty_live_denominator_and_fails_an_outgrown_zero_pin` relies on it), so it belongs in the comparison, not the parser.

## Risks / unsure about
- The merge-base test still skips in any clone without `origin/main` and fails only under `GITHUB_ACTIONS`; unchanged from round 1.
- The full suite was run with `-n auto` (xdist) rather than serially this round, to fit the command timeout; round 1 ran it serially with the same result shape. Nothing new is order-dependent.

## Response to round N-1
|---|---|---|---|---|
| `build/identity_eval.py:1138` — positive pin rewritten to `[0, 0]` passes `coverage_lowered` vacuously | yes (pin-to-pin comparison; the ask) | partial: the "both live-coverage checks also pass on an empty route" clause — a live 0/0 is undefined coverage, the pre-existing `_below_baseline` pass is documented and in-plan, and the erasure path is now closed at the pin | no | `_pin_lowered` at `build/identity_eval.py:1116`; `coverage_lowered({"github": (0, 0)}, {"github": (211, 328)}, None)` returns `["github: pin lowered from 211/328 to 0/0", "a pin may only go down with a 'lowered_because' written or changed in this PR"]`; regression test `test_coverage_lowered_reads_a_zero_denominator_pin_as_zero`; commit `6b59ac4e`; full suite 1931 passed, 1 skipped |

PR #554, opened 2026-09-11
Require a host on every URL field the schemas constrain

### LINKED ISSUE #525: Every URL field accepts a bare scheme with no host
Every URL field in the schemas is constrained by `pattern: ^https?://` and nothing else, so a bare
scheme with no host validates.

Measured on 2026-09-09 by setting `end_of_life.source` to the literal string `https://`:

```
uv run python -m build.validate     -> 0 error(s), 2 warning(s)
uv run python -m build.check_payload -> 616 products, 348 organizations, ok
```

It passes both gates, is copied unchanged into the payload by `build/serialize.py`, and would render
as a link a reader cannot follow.

## Where the pattern is used

| Schema | Field |
|---|---|
| `product.schema.json` | `source`, `url` |
| `organization.schema.json` | `url` |
| `score.schema.json` | `url` |

`organization.homepage` and `product.url` in one position carry no constraint at all;
`registry.schema.json` `homepage` uses `format: uri`, which is a third convention.

## Why it matters more than it looks

A citation is the part of a record a reader uses to check the claim, and these fields exist for
exactly that. A gate that accepts `https://` is not checking the thing it was written to check — it
confirms the shape of a URL without confirming there is one. The same reasoning the repo already
applies to evidence elsewhere: recording that a source exists is not the same as recording a source.

## What a fix looks like

Require a host: something like `^https?://[^/\s]+\.[^/\s]+` rather than `^https?://`, applied
consistently across the four fields, plus a decision on whether `format: uri` or the pattern is the
house convention — right now there are three.

Worth checking against the existing corpus before landing: 616 products and 348 organizations carry
these fields, and a stricter pattern may reject a URL that is unusual but real. That check is the
work here, not the regex.

## Not a blocker for anything currently open

Found during an adversarial review of #524, which introduced `end_of_life.source` using the existing
convention. #524 follows the house rule correctly; this issue is about the rule.

ccerv1:
## Summary

Every URL field in `docs/schemas/` was constrained by `^https?://` and nothing else, so the literal `https://` validated, passed `check_payload`, and shipped as a link nobody can follow (#525, measured 2026-09-09).

- **One pattern on all six URL positions**: `end_of_life.source` and artifact `url` in `product.schema.json`, `url` and `homepage` in `organization.schema.json`, `url` in `score.schema.json`, `homepage` in `registry.schema.json`. It requires an http(s) scheme, a dotted hostname of label characters, an optional port, then a path, query, fragment or end of string. `https://`, `https://?a.b`, `https://#a.b` and `https://user.name@` all fail; a dotted IPv4 host passes; `http://localhost` fails.
- **`pattern`, not `format: uri`.** `build/validate.py` calls `jsonschema.validate` with no format checker, so the `format: uri` on the registry homepage asserted a check nobody ran. Replaced.
- **Diagnostics name the field.** A schema error now reads `products/llama: schema: github[0].url: 'https://' does not match ...` rather than quoting only the value.
- **`check_payload` uses the character-identical regex** so a URL the schema rejects cannot pass the payload gate on a looser reading.
- **Corpus check ran on the branch**: 5,635 http(s) strings under `sources/` (products, organizations, scores, registry), 0 rejected. No corpus edits.

One thing a follow-up could tidy: the convention paragraph is repeated in each field's description rather than stated once and linked. Left as is to keep this PR to the rule.

Closes #525.

## Test plan

- `uv run python -m build.validate` reports 0 errors
- `uv run python -m build.check_payload` reports ok
- `uv run pytest -q` passes on the branch
- New tests in `tests/test_validate.py` assert the bare-scheme and hostless cases fail with a field-path diagnostic and a normal URL passes, per constrained schema
- Codex independently executed the six regression cases and the 5,635-string corpus walk in round 3


ccerv1:
## What I did
- files touched (commit 05f29598 on loop/url-fields-require-host-2026-09-11-a2):
  - `docs/schemas/product.schema.json` — `end_of_life.source` and `definitions.url.url` (the artifact URL every github/huggingface/etc block refs) now `pattern: ^https?://[^/\s]+\.[^/\s]+`; descriptions corrected ("only its scheme is checked" -> scheme and dotted host) and the shared convention paragraph appended.
  - `docs/schemas/organization.schema.json` — account `url` tightened to the same pattern; `homepage` gains the pattern (was free text). Description history kept, "free text with no pattern" sentence corrected with a date.
  - `docs/schemas/score.schema.json` — citation `url` tightened to the same pattern.
  - `docs/schemas/registry.schema.json` — `homepage` `format: uri` replaced with the pattern plus the convention paragraph (the only field in that file with a description; the file is otherwise compact).
  - `build/check_payload.py` — end_of_life source check mirrors the schema regex (`_URL_WITH_HOST`) instead of `startswith(("http://","https://"))`, so `https://` no longer passes the payload gate either.
  - `tests/test_validate.py` — fixture citation `https://x` -> `https://example.com/x` (it would now fail the schema); new parametrized `test_bare_scheme_url_fails_schema_and_dotted_host_passes` covering 6 positions across product (artifact url, end_of_life.source), organization (homepage, account url), score (sources[].url), registry (homepage): `https://` yields a `schema` error naming the record, `https://example.com/some/page` yields none.
- commands run (all foreground, in the clone):
  - corpus measurement script over `sources/{products,organizations,scores,registry}`: **5,635 http(s) strings, 0 rejected** by the pattern (matches the 2026-09-11 session figure).
  - `uv run python -m build.validate` -> `0 error(s), 2 warning(s)` (pre-existing model_families overlap warnings).
  - `uv run python -m build.check_payload` -> `685 products, 375 organizations, ok`.
  - `uv run pytest -q` -> `1914 passed in 1550.16s`.

## Why
- decisions:
  - Pattern `^https?://[^/\s]+\.[^/\s]+` as planned: rejects `https://`, `http://localhost`, a bare-IP-less host; corpus re-measured on the branch with zero rejections, so no loosening needed.
  - Convention written into each field's `description` (one identical paragraph, cites #525 and the measured count) rather than a separate doc: the schemas already carry their own history inline and nothing else in `docs/` describes URL fields.
  - `pattern` over a format checker: matches the existing `format: date` note in product.schema.json and keeps `build/validate.py` untouched.
  - Tightened `check_payload` too, since the goal names it as a gate the bare scheme passed; the payload check now agrees with the schema.
- alternatives rejected:
  - Adding `FormatChecker` to every `jsonschema.validate` call: larger, and registry would be the only field using `format`.
  - Leaving `check_payload` on `startswith`: it would keep passing exactly the string the issue is about.
- PR-body note: #525 mentions "the one `product.url` position with no pattern". On this branch product.schema.json has exactly two URL positions (`end_of_life.source` and `definitions.url.url`, ref'd by all eight artifact blocks) and both carried `^https?://` already; both now carry the host pattern. Nothing was left out.

## Risks / unsure about
- The pattern rejects intranet-style hostless names and bare IPv4 hosts. None exist in the corpus today; a future contributor adding one gets a schema error that names the field, which is the intended trade.
- `score.schema.json` has three `format: date` fields (lines ~135/206/267/298) that are equally unenforced; out of scope for #525, left as is.
- The JSON schema files for product/organization/score were rewritten via `json.dumps(indent=2)`; the diff is confined to the edited lines (20/7/4 lines), so formatting was preserved.

## Response to round N-1
|---|---|---|---|---|
| n/a (round 1) | | | | |

ccerv1:
## What I did
- files touched:
  - `docs/schemas/product.schema.json` (`end_of_life.source`, artifact `url` x2), `organization.schema.json` (`homepage`, `github[].url`), `score.schema.json` (`sources[].url`), `registry.schema.json` (`homepage`): pattern on all six URL positions changed from `^https?://[^/\s]+\.[^/\s]+` to `^https?://[A-Za-z0-9-]+(\.[A-Za-z0-9-]+)+(:[0-9]+)?([/?#]|$)`. The shared "URL convention (#525)" paragraph now says the host must be a dotted hostname of label characters plus optional port, names the hostless shapes it rejects (`https://?a.b`, `https://#a.b`, `https://user.name@`), and carries the re-measured count (5,635 strings, 0 rejected). The per-field sentences that said only "a dotted host" is checked now point at the convention paragraph.
  - `build/validate.py`: new `_schema_msg(e)` prefixes a schema error with `e.json_path` minus the `$.`, applied at all six `jsonschema.validate` call sites. Output is now `products/llama: schema: github[0].url: 'https://' does not match ...`.
  - `tests/test_validate.py`: `test_hostless_url_fails_schema_naming_the_field_and_real_url_passes`, parametrized over the six fields with the expected field path; five hostless inputs (`https://`, `https://?a.b`, `https://#a.b`, `https://user.name@`, `http://localhost`) must each fail with a message starting `<place>: schema: <field>: '<url>' does not match`; four real URLs (bare host, path, port+query+fragment, IPv4) must pass.
- commands run:
  - corpus walk over `sources/{products,organizations,scores,registry}/**/*.yaml` with the new pattern: 5,635 http(s) strings, 0 rejected (matches the 2026-09-11 session figure)
  - `uv run python -m build.validate` -> 0 error(s), 2 warning(s)
  - `uv run python -m build.check_payload` -> 685 products, 375 organizations, ok
  - `uv run pytest -q tests/test_validate.py -k hostless` -> 6 passed
  - `uv run pytest -q` -> PYTEST_RESULT

## Why
- decisions: host must be `[A-Za-z0-9-]` labels joined by dots, then optional `:port`, then `/`, `?`, `#` or end of string. This keeps `@`, `?` and `#` out of the host, so the reviewer's three shapes fail. Bare IPv4 still passes (dotted). `http://localhost` fails, as in round 1. The field path comes from `jsonschema`'s `json_path` rather than being hand-built per call site, so every schema error (not only URLs) now names its field.
- alternatives rejected: allowing userinfo (`user@host`): none in the corpus, and the bug is exactly dotted text that is not a host. A URL-library parse: schemas only express `pattern`, and the validator runs no custom checkers (plan step 3).

## Risks / unsure about
- Underscore hosts (`a_b.example.com`) and IPv6 literals fail the pattern. Zero in the corpus today; a contributor adding one gets a schema error naming the field and the pattern would need loosening.
- The round-2 work was found uncommitted in the clone with descriptions quoting 5,644; my re-measure (restricted to the four directories #525 names) gives 5,635, so I corrected the descriptions to 5,635 before committing.

## Response to round 1
|---|---|---|---|---|
| pattern accepts `https://?a.b`, `https://#a.b`, `https://user.name@` (product.schema.json:199 and the other five positions) | x | | | new pattern in all four schemas; `HOSTLESS_URLS` in `tests/test_validate.py` covers all three plus `https://` and `http://localhost` across six fields (30 negative checks), 6 passed |
| diagnostic does not name the field (tests/test_validate.py:936) | x | | | `build/validate.py::_schema_msg` prefixes the json path; the test asserts `startswith(f"{place}: schema: {field}: {bad!r} does not match")` per field |

ccerv1:
## What I did
- files touched:
  - Root cause of the round-2 fail: the round-2 changes were written to the clone but never committed. `git status` at the start of this round showed all six files modified against HEAD `05f29598` (the round-1 commit), so the reviewer's branch had the round-1 pattern and the round-1 diagnostic. This round verified the working-tree changes and committed them as `cd7de9d6`.
  - `docs/schemas/product.schema.json` (`end_of_life.source`, artifact `url` x2), `organization.schema.json` (`homepage`, `github[].url`), `score.schema.json` (`sources[].url`), `registry.schema.json` (`homepage`): pattern on all six URL positions is `^https?://[A-Za-z0-9-]+(\.[A-Za-z0-9-]+)+(:[0-9]+)?([/?#]|$)`; the shared "URL convention (#525)" paragraph names the rejected hostless shapes and carries the 5,635 / 0 count.
  - `build/validate.py`: `_schema_msg(e)` prefixes every schema error with `e.json_path` minus `$.`, applied at all six `jsonschema.validate` call sites. Output for the reviewer's case is `products/llama: schema: github[0].url: 'https://' does not match ...`.
  - `tests/test_validate.py`: `test_hostless_url_fails_schema_naming_the_field_and_real_url_passes`, parametrized over six fields; asserts each of five hostless inputs fails with a message starting `<place>: schema: <field>: '<url>' does not match`, and four real URLs pass.
- commands run (all in the clone, on the working tree that is now HEAD `cd7de9d6`):
  - corpus walk over `sources/{products,organizations,scores,registry}/**/*.yaml` with the committed pattern: 5,635 http(s) strings, 0 rejected
  - direct regex check: `https://`, `https://?a.b`, `https://#a.b`, `https://user.name@`, `http://localhost` all rejected; `https://example.com`, `https://1.2.3.4/x`, `https://a.b:8080/p?q#f` all accepted
  - `uv run python -m build.validate` -> 0 error(s), 2 warning(s)
  - `uv run python -m build.check_payload` -> 685 products, 375 organizations, ok
  - `uv run pytest -q tests/test_validate.py -k hostless` -> 6 passed
  - `uv run pytest -q` -> 1914 passed in 1356.54s
  - `git commit` -> `cd7de9d6 fix(schemas): require a host on every URL field`; `git status --short` clean afterwards

## Why
- decisions: no new design this round. The round-2 pattern and diagnostic were correct on disk and only missing from git; committing them is the whole fix. Kept a second commit with the same message rather than amending `05f29598`, since the harness has already moved that commit once and the reviewer's diff references it.
- alternatives rejected: amending or squashing the round-1 commit (rewrites a commit the harness and reviewer already saw).

## Risks / unsure about
- Underscore hosts and IPv6 literals fail the pattern; zero in the corpus today, a contributor adding one gets a schema error naming the field.
- Two commits with identical subjects on the branch. Harmless for the PR, slightly noisy in `git log`.

## Response to round 2
|---|---|---|---|---|
| round-two pattern fix absent from branch; all six fields still accept `https://?a.b`, `https://#a.b`, `https://user.name@` (product.schema.json:199) | x | | | the fix was uncommitted in the clone; now committed in `cd7de9d6`. `grep -n 'https?://' docs/schemas/*.json` at HEAD shows `^https?://[A-Za-z0-9-]+(\.[A-Za-z0-9-]+)+(:[0-9]+)?([/?#]\|$)` at all six positions; direct regex check rejects all three shapes plus `https://` and `http://localhost`; corpus 5,635 / 0 rejected |
| diagnostic fix absent; output names products/llama but omits `github[0].url` (build/validate.py:639) | x | | | `_schema_msg` committed in `cd7de9d6` (validate.py:29, applied at :651, :661, :669, :678, :726, :740); `tests/test_validate.py::test_hostless_url_fails_schema_naming_the_field_and_real_url_passes` asserts `startswith(f"{place}: schema: {field}: {bad!r} does not match")` for six fields x five inputs, 6 passed; full suite 1914 passed |

PR #558, opened 2026-09-12
Confirm a source on the quoted fragments inside shows, not on the sentence around them

ccerv1:
# shows-match-fragments — pass

**Goal.** The deterministic re-verification leg re-dates about one product in twenty-five. Measured on

## Success criteria
- [ ] `build/reverify.py`'s shows-match confirms when **every** quoted fragment inside the recorded `shows` occurs in the fresh body, under the same normalization the whole-sentence path uses (whitespace-collapsed, `html.unescape` tried); the existing whole-sentence match keeps working where it already fires
- [ ] a `shows` with **no** quoted fragment is unchanged in behaviour — it can still only confirm by whole-sentence match, so nothing becomes easier to confirm by having said less
- [ ] a placeholder `shows` still never confirms, by either path
- [ ] **partial fragment matches never confirm.** The measured counter-example is `atropos`, whose `shows` carries nine fragments of which two occur; it must still drift. A rule that accepted "at least one" would confirm it, and that is the rubber stamp this apparatus exists to stop
- [ ] a fragment shorter than a stated minimum length does not count toward confirmation, and the threshold is defended in the PR — a two-word quote occurring somewhere on a page proves nothing
- [ ] the before/after effect is **measured, not asserted**: run `uv run python -m build.reverify --limit 25 --pace 1.5 --dry-run --report <path>` on `main` and on the branch over the same slugs, and quote both reason counts (`reconfirmed_by_shows`, `drifted`, `transient`, `skipped`) in the PR body
- [ ] no file under `sources/` changes. `--dry-run` writes nothing; the deliverable is the matcher
- [ ] `uv run python -m build.validate` reports 0 errors and `uv run pytest -q` passes

## Outcome
`pass` after 2 round(s). Branch `loop/shows-match-fragments-2026-09-12-a1`.

## Final verdict
```json
{
 "verdict": "pass",
 "blocking": [],
 "non_blocking": [],
 "settled_now": [
  "The round 1 blocker is fixed: unbalanced quotes refuse fragment confirmation while preserving whole-sentence matching. Six isolated matcher checks passed.",
  "The reported before/after measurements and validation results satisfy the requested checks; the full diff contains no sources/ changes."
 ],
 "escalation": null
}
```

## Settled ground this run added (9)
- The three confirmation paths and their terms. This changes how one of them matches, not what
- A placeholder `shows` never confirms.
- Machine re-dating stays limited to openness.
- No score file changes in this PR.
- Short fragments must all occur, with at least one fragment meeting the 24-character floor.
- Presence is sufficient; fragment ordering need not be imposed.
- The live atropos fragment counts can differ from the earlier probe; partial-match refusal remains independently testable.
- The round 1 blocker is fixed: unbalanced quotes refuse fragment confirmation while preserving whole-sentence matching. Six isolated matcher checks passed.
- The reported before/after measurements and validation results satisfy the requested checks; the full diff contains no sources/ changes.

Later runs of this job inherit `/workspace/GitHub/personal/agent-loops/jobs/settled/shows-match-fragments.json`; prune it there.

### DIFF
--- a/build/reverify.py
+++ b/build/reverify.py
@@ -4,9 +4,11 @@ What it does: for each of the N oldest products (by the oldest of its three axis
 ties broken by slug), re-fetch every digested source that `establishes` a recorded
 dimension on the selected axes through build.fetch_source. A dimension re-confirms when
 every establishing source's fresh fetch is non-transient and either (a) the body is
-byte-identical to what was recorded, (b) the source's recorded `shows` excerpt still
-occurs in the fresh body, or (c) for a GitHub license API / repo endpoint or a Hugging
-Face model-info source, the fresh SPDX id normalizes to the same license already recorded.
+byte-identical to what was recorded, (b) the source's recorded `shows` still checks out
+against the fresh body — the whole sentence occurs, or every fragment quoted inside it
+does and at least one of those is long enough to discriminate — or (c) for a GitHub
+license API / repo endpoint or a Hugging Face model-info source, the fresh SPDX id
+normalizes to the same license already recorded.
 Whichever path confirms it, the source takes the new `accessed`, `http_status` and
 `content_sha256`, and once every dimension an axis records has confirmed, that axis's
 `last_verified` is stamped to today. Any drift, any transient result, or any dimension
@@ -26,6 +28,12 @@ by default (--axes openness). Ruled again 2026-09-03 (#445 follow-up): shows-mat
 SPDX comparison are also acceptable confirmations, since byte identity is the wrong test
 for evidence pages that legitimately re-render on every load. See
 docs/reference/evidence-and-freshness.md.
+
+Shows-match tests the quoted fragments as well as the whole sentence, because `shows` is
+written as a curator's sentence *about* the page with the verbatim material quoted inside
+it, and such a sentence never occurs on the page it describes. Measured 2026-09-12 over
+the 25 oldest products: 72 sources drifted, 0 were skipped, 0 carried a placeholder
+`shows`, and whole-sentence match confirmed 4 of them. See `_fragments_confirm`.
 """
 from __future__ import annotations
 
@@ -214,15 +222,90 @@ def _read_body(fetched: dict) -> str | None:
     return html.unescape(raw.decode("utf-8", errors="replace"))
 
 
+# The corpus quotes with both straight and curly double quotes, sometimes in one
+# sentence, so the delimiters are folded together before the pair scan.
+_SMART_DOUBLE_QUOTES = "\u201c\u201d\u201e\u201f\u2033"
+
+# A quoted fragment shorter than this does not count toward a confirmation. Chosen off
+# the corpus rather than by taste: of the 1,714 fragments quoted inside a `shows` today,
+# the 563 below 24 characters are almost entirely JSON key names and license ids —
+# `license`, `name`, `spdx_id`, `MIT`, `Apache-2.0`, `open_issues_count` — and one- or
+# two-word English like `gated`, `open source`, `Enterprise`. Those occur on any page of
+# the kind being cited, so finding one discriminates nothing. 426 of the 563 are one or
+# two words, and at 24 characters only 5 of the 1,151 surviving fragments are that short
+# (they are file paths, `surfsense_backend/app/proprietary/LICENSE` and the like). At a
+# floor of 20 characters, 93 two-word fragments still qualify; at 24, two do. A two-word
+# quote found somewhere on a page proves nothing, and 24 is where the corpus stops
+# offering them.
+MIN_FRAGMENT_CHARS = 24
+
+
+def _shows_fragments(shows: str) -> list[str]:
+    """The verbatim material quoted inside a curator's `shows` sentence.
+
+    A plain double-quote pair scan: split the whitespace-normalized sentence on `"` and
+    the odd-indexed segments are the insides of the pairs. An unbalanced trailing quote
+    opens a pair that never closes, and that dangling segment is dropped rather than
+    guessed at. Nested quotes are not parsed as nesting — the scan just pairs off
+    quotes in order, which is what a curator writing `"license":{"spdxId":"MIT"}` inside
+    a sentence actually means.
+
+    Fragments are matched with `in` and are never compiled, so a regex metacharacter, a
+    bracket or a bare URL inside one is only ever text.
+    """
+    text = _normalize_text(str(shows))
+    for ch in _SMART_DOUBLE_QUOTES:
+        text = text.replace(ch, '"')
+    parts = text.split('"')
+    pairs = (len(parts) - 1) // 2
+    return [f for f in (p.strip() for p in parts[1::2][:pairs]) if f]
+
+
+def _fragments_confirm(shows: str, body: str) -> bool:
+    """Every fragment quoted inside `shows` still occurs in `body`, and at least one of
+    them is long enough to mean something. `body` is already normalized.
+
+    Both halves carry weight, and dropping either one breaks it in a different
+    direction. **Every** is what stops a partial match being read as a confirmation: a
+    `shows` that quotes nine things and finds two of them on the page has not been
+    confirmed; seven of the things it says the page carries are not on the page. **At least one long fragment** is
+    what stops the opposite failure, where a `shows` whose only quoted material is
+    `"license"` and `"MIT"` would confirm against any GitHub API response ever served.
+    A short fragment still has to be there; it just cannot be the thing that earns the
+    date. A `shows` that quotes nothing at all has no long fragment and so never reaches
+    this path — saying less must not make a source easier to confirm.
+    """
+    fragments = _shows_fragments(shows)
+    if not any(len(f) >= MIN_FRAGMENT_CHARS for f in fragments):
+        return False
+    return all(f in body for f in fragments)
+
+
 def _shows_confirms(src: dict, fetched: dict) -> bool:
-    """The source's recorded `shows` excerpt still occurs in the fresh body."""
+    """The source's recorded `shows` still checks out against the fresh body.
+
+    Two ways, under one normalization (whitespace collapsed, the body read through
+    `html.unescape`). The whole sentence may occur verbatim, which is the original test
+    and is unchanged. Otherwise the fragments quoted inside it must all occur.
+
+    The second path exists because `shows` is almost never an excerpt. It is a curator's
+    sentence *about* the page with the verbatim material quoted inside it — "repo page
+    carries the label "Public archive" and the banner "This repository was archived..."" —
+    and a sentence like that does not appear on the page it describes, so testing the
+    whole of it tests the curator's prose rather than the evidence. Measured 2026-09-12
+    over the 25 oldest products through `build.fetch_source`: 72 sources drifted, none
+    skipped, none carrying a placeholder `shows`, and the whole-sentence path confirmed 4.
+    """
     shows = src.get("shows")
     if not shows or _is_placeholder_shows(shows):
         return False
     body = _read_body(fetched)
     if body is None:
         return False
-    return _normalize_text(str(shows)) in _normalize_text(body)
+    body = _normalize_text(body)
+    if _normalize_text(str(shows)) in body:
+        return True
+    return _fragments_confirm(str(shows), body)
 
 
 def _spdx_confirms(url: str, fetched: dict, recorded_license: str) -> bool:
--- a/build/reverify.py
+++ b/build/reverify.py
@@ -268,12 +268,12 @@ def _fragments_confirm(shows: str, body: str) -> bool:
     Both halves carry weight, and dropping either one breaks it in a different
     direction. **Every** is what stops a partial match being read as a confirmation: a
     `shows` that quotes nine things and finds two of them on the page has not been
-    confirmed; seven of the things it says the page carries are not on the page. **At least one long fragment** is
-    what stops the opposite failure, where a `shows` whose only quoted material is
-    `"license"` and `"MIT"` would confirm against any GitHub API response ever served.
-    A short fragment still has to be there; it just cannot be the thing that earns the
-    date. A `shows` that quotes nothing at all has no long fragment and so never reaches
-    this path — saying less must not make a source easier to confirm.
+    confirmed; seven of the things it says the page carries are not on the page.
+    **At least one long fragment** stops the opposite failure, where a `shows` whose
+    only quoted material is `"license"` and `"MIT"` would confirm against any GitHub API
+    response ever served. A short fragment still has to be there; it just cannot be the
+    thing that earns the date. A `shows` that quotes nothing at all has no long fragment,
+    so it never reaches this path: saying less must not make a source easier to confirm.
     """
     fragments = _shows_fragments(shows)
     if not any(len(f) >= MIN_FRAGMENT_CHARS for f in fragments):
--- a/build/reverify.py
+++ b/build/reverify.py
@@ -240,15 +240,23 @@ _SMART_DOUBLE_QUOTES = "\u201c\u201d\u201e\u201f\u2033"
 MIN_FRAGMENT_CHARS = 24
 
 
-def _shows_fragments(shows: str) -> list[str]:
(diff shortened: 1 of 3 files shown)


ccerv1:
## What I did

Made `reverify.py`'s shows-match test the quoted fragments inside a `shows` as well as the
whole sentence. Three commits on `loop/shows-match-fragments-2026-09-12-a1`, nothing amended or rewritten:
`4af751f8` (the matcher, its tests, the doc rule), `01ab8e25` (a paragraph break the first
commit ran together in the doc) and `45f7bce5` (a docstring rewrap).

- files touched:
  - `build/reverify.py` — added `_SMART_DOUBLE_QUOTES`, `MIN_FRAGMENT_CHARS = 24`,
    `_shows_fragments()` and `_fragments_confirm()`; `_shows_confirms()` now tries the whole
    sentence first (unchanged path) and then the fragments. Module docstring updated.
  - `docs/reference/evidence-and-freshness.md` — the "Machine re-verification" section is
    normative for what a confirmation may claim, so it now states what "the `shows` still
    checks out" means, both conditions, and why each is load-bearing.
  - `tests/test_reverify.py` — 11 new fixture-only tests (no network). 29 in the file, all pass.
- commands run (all foreground, all waited on):
  - `uv run python -m build.reverify --limit 25 --pace 1.5 --dry-run --report <path> --body-dir <dir>`
    on `main` (working tree was at `main` before any edit) and again on the branch, same 25 slugs.
  - `uv run python -m build.validate` → **0 errors**, 2 pre-existing warnings (overlapping
    `model_families` glob patterns, unrelated). Re-run on the final tree, same result.
  - `uv run pytest -q` → **2049 passed, 1 skipped, exit 0** (12m33s) on the final three-commit
    tree, run to completion in one go and read before returning. An earlier full run over the
    first commit was also green (2048 passed, 1 skipped).
  - `uv run pytest -q tests/test_reverify.py tests/test_docs_integrity.py` → 44 passed, after the
    second commit.
  - `git diff --name-only 7cea4fbc..HEAD -- sources/` → empty.
  - Corpus statistics for the threshold, over all 3,932 `shows` fields in `sources/scores/`.

### The measurement (same 25 slugs, `--limit 25 --pace 1.5 --dry-run`)

| | `reconfirmed_by_shows` | `drifted` | `transient` | `skipped` | products re-dated |
|---|---|---|---|---|---|
| `main` | 4 | 72 | 2 | 0 | 1/25 |
| branch | 15 | 61 | 2 | 0 | 4/25 |

`reconfirmed_by_spdx` is 1 on both. The eleven sources that moved out of `drifted` are exactly
the eleven that moved into `reconfirmed_by_shows`, and all four whole-sentence confirmations
from the `main` run (`aws-neuron` x2, `groq-inference` x2) are still confirmed on the branch —
the existing path still fires everywhere it already fired. Newly re-dated: `cerebras-inference`,
`lamini`, `llama-cpp` (`amazon-bedrock-custom-models-fine-tuning` was already re-dating).
`git status --porcelain sources/` is empty — `--dry-run` wrote nothing.

## Why

- decisions:
  - **Fragment extraction is a plain double-quote pair scan** over the whitespace-normalized
    sentence: split on `"`, take the odd segments. Curly quotes (`“ ” „ ‟ ″`) are folded to
    straight first, because the corpus uses both. An unbalanced trailing quote opens a pair
    that never closes and that dangling segment is dropped rather than guessed at. Nested
    quotes are not parsed as nesting; the scan pairs off quotes in order, which is what a
    curator writing `"license":{"spdxId":"MIT"}` inside a sentence actually means.
  - **Fragments are matched with `in`, never compiled.** A regex metacharacter, a bracket or a
    bare URL inside a fragment is only ever text. Tested both ways round: the literal string
    confirms, and a body a compiled pattern would have matched does not.
  - **Every fragment must occur, and at least one must be ≥ 24 characters.** Both halves are
    load-bearing and they fail in opposite directions. *Every* is what refuses a partial
    match — "at least one matched" is the rubber stamp this leg exists to stop. *At least one
    long fragment* is what stops a `shows` whose only quoted material is `"license"` and
    `"MIT"` confirming against any GitHub API response ever served. A short fragment must
    still be present; it just cannot be the thing that earns the date.
  - **A `shows` with no quoted fragment is untouched.** No fragment clears the floor, so
    `_fragments_confirm` returns False before looking at the body, and the whole-sentence test
    is the only path — saying less cannot make a source easier to confirm. A placeholder
    `shows` is refused before either path, unchanged.
  - **The floor is 24 characters, picked off the corpus.** Of the 1,714 fragments quoted inside
    a `shows` today, the 563 below 24 characters are almost entirely JSON key names and license
    ids (`license`, `name`, `spdx_id`, `MIT`, `Apache-2.0`, `open_issues_count`,
    `preferredFileType`) and one- or two-word English (`gated`, `open source`, `Enterprise`,
    `Custom pricing`). 426 of those 563 are one or two words. Above the floor, only 5 of the
    1,151 survivors are one or two words, and they are file paths
    (`surfsense_backend/app/proprietary/LICENSE`). At a floor of 20, 93 two-word fragments
    still qualify; at 24, two do. The brief's own test — "a two-word quote occurring somewhere
    on a page proves nothing" — is only actually enforced at 24.
  - **The floor is also what keeps `atropos` drifting**, which is the strongest defense
    available and it is measured rather than argued. On the branch `atropos` is not re-dated:
    the citation establishing `source` from `api.github.com/repos/NousResearch/atropos` quotes
    only `private`, `visibility`, `public`, `archived`, `disabled`, `spdx_id`, `MIT`,
    `pushed_at`, `2026-07-04T17:39:01Z` and `open_issues_count`. Nothing there reaches 24
    characters. At a floor of 20 the timestamp would qualify and `atropos` would be re-dated on
    it alone; at 16 so would `open_issues_count`. `tests/test_reverify.py::
    test_the_atropos_api_citation_does_not_confirm_on_json_key_names` pins that shape.
- alternatives rejected:
  - **Ignore sub-threshold fragments entirely** (require only the long ones to occur). One
    entry apart it measures the same on this sample (`anyscale-fine-tuning` confirms under it
    and not under the shipped rule), but it is looser, and the brief says *every* quoted
    fragment must occur. Rejected for the stricter reading.
  - **A word-count floor, or a hybrid of chars and words.** `"Apache License 2.0"` (3 words,
    18 chars) would survive where `"open_issues_count"` (1 word, 17 chars) would not, which is
    a better outcome for those two cases — but it is a two-clause rule to defend, and the brief
    asks for a single stated minimum length. A hybrid would also have kept `atropos`
    confirmable via `"2026-07-04T17:39:01Z"` on a character clause.
  - **Floors of 16, 20 and 28.** Swept against the 25-slug sample using the saved bodies: 14,
    11, 9 and 8 drifting entries would confirm at 16 / 20 / 24 / 28. 16 and 20 both re-date
    `atropos` on JSON furniture; 28 discards 40% more real material for one further entry.
  - **Extending `html.unescape` handling** (trying the raw body as well as the unescaped one).
    That would change the whole-sentence path too. The brief says fragments match under the
    *same* normalization, so `_shows_confirms` reuses `_read_body` and `_normalize_text`
    verbatim.

## Risks / unsure about

- **The brief's atropos number did not reproduce, though its conclusion did.** The brief says
  `atropos` has "nine fragments of which two occur". Re-probed through `build.fetch_source` on
  2026-09-12, its GitHub repo page `shows` yields 15 fragments and **all 15 occur**, including
  `"Public archive"` and the full archive banner; the LICENSE, repo-API and portal citations
  likewise hit on every fragment. The brief anticipated this ("my probe fetched with `requests`
  … redo it through the real call"). So the *reason* `atropos` still drifts is the length
  floor, not a partial match. The success criterion holds; the stated mechanism behind it does
  not. Worth a reviewer's eye, since it means the partial-match refusal is pinned by a fixture
  rather than by a live product in this sample.
- **Before and after are two live fetches ~25 minutes apart**, so a page could in principle have
  moved between them. The delta is internally consistent (drifted −11, shows-confirmed +11, the
  same eleven URLs), and `transient`/`skipped` are identical, which is what I would expect if
  nothing else moved.
- **Mis-paired quotes fail safe, but they do fail.** An odd quote inside a sentence (an inches
  mark, a nested quote) shifts the pairing for the rest of it and produces junk fragments that
  will not be found, so the source drifts. That is the current behaviour too — such a sentence
  does not whole-sentence match either — but it is a way the fragment path can produce a false
  drift rather than a false confirmation.
- **The floor is a corpus statistic, not a law.** If curators start quoting longer JSON values,
  24 characters will admit material that discriminates no better than `"license"` does. The
  number is a named constant with the reasoning beside it so it can be re-derived; nothing
  re-derives it automatically.
- GitHub HTML is served differently to different clients. `llama-cpp`'s and `lamini`'s new
  confirmations rest on README prose rendered into that HTML. If GitHub ever serves the
  re-verification leg a JS-shell page, those confirmations will flip to drift rather than to a
  false stamp, which is the right direction, but the leg will look noisier.

## Response to round N-1

|---|---|---|---|---|
| — (round 1, no prior review) | | | | |

ccerv1:
## What I did

Fixed the one blocking item. The fragment pair scan no longer repairs an unbalanced `shows`;
it refuses it. Added as a fourth commit, `1ff3e872`, on top of round 1's three — nothing
amended, rebased or reset.

- files touched:
  - `build/reverify.py` — `_shows_fragments()` now returns `list[str] | None` and returns
    `None` when the folded sentence carries an odd number of `"` delimiters.
    `_fragments_confirm()` declines on `None` before any length or containment test. The
    docstrings say why the dangling segment is refused rather than dropped.
  - `tests/test_reverify.py` — `test_an_unbalanced_trailing_quote_opens_no_fragment` is
    replaced by `test_an_unbalanced_quote_refuses_the_fragment_path`, which asserts
    `_shows_fragments(...) is None` and requires the source to **drift** on the same body
    that used to confirm it. Added
    `test_an_unbalanced_quote_still_confirms_on_the_whole_sentence`, so the refusal is
    pinned as fragment-path-only. 30 tests in the file.
  - `docs/reference/evidence-and-freshness.md` — the "Machine re-verification" section is
    normative for what a confirmation may claim, so it now states that a sentence whose
    quotes do not pair off gets no fragment test at all, and why.
- commands run (all foreground, all waited on, all read before returning):
  - `uv run python -m build.reverify --limit 25 --pace 1.5 --dry-run --report <path> --body-dir <dir>`
    on `main` (via a throwaway `git worktree`, since the harness clone must stay on the
    branch; worktree removed afterwards) and on the branch, same 25 slugs, both runs today
    with the final code. 2m12s and 2m16s.
  - `uv run python -m build.validate` → **0 errors**, 2 warnings (overlapping `model_families`
    glob patterns, pre-existing and unrelated).
  - `uv run pytest -q -n 8` → **2050 passed, 1 skipped**, exit 0, 5m30s. Run to completion in
    the foreground in one call. (`-n 8` because the serial suite is ~12 minutes; the same
    2050 tests run either way.)
  - `uv run pytest -q tests/test_reverify.py` → 30 passed.
  - A corpus scan of every `shows` field in `sources/scores/` for odd quote counts after
    curly-folding.
  - `git status --porcelain sources/` → empty.

### The measurement (both runs today, final code, same 25 slugs)

| | `reconfirmed_by_shows` | `drifted` | `transient` | `skipped` | products re-dated |
|---|---|---|---|---|---|
| `main` | 4 | 72 | 2 | 0 | 1/25 |
| branch | 15 | 61 | 2 | 0 | 4/25 |

`reconfirmed_by_spdx` is 1 on both. The same 25 slugs in the same order both times; no
source that confirmed on `main` fails on the branch (`shows-confirmed lost: []`), no new
drift, `transient` is the same two PyPI entries. `atropos` is **not** re-dated on either
side — its `api.github.com/repos/NousResearch/atropos` citation quotes only JSON key names
and values, none of which reaches the 24-character floor. Identical to round 1's table, which
is what I expected: the balance rule cannot move a corpus that has no unbalanced `shows`.

## Why

- decisions:
  - **`_shows_fragments` returns `None` rather than a shorter list.** The caller has to
    distinguish "this sentence quotes nothing" from "this sentence is unreadable", because
    the two mean different things and only one of them is a curator's choice. A sentinel
    return makes the refusal impossible to forget at the call site; a `[]` return would have
    fallen through the existing `any(... >= MIN_FRAGMENT_CHARS ...)` guard and looked like
    the no-quotes case.
  - **Parity of `"` is the whole test, measured after curly-folding.** Curly quotes are
    folded to straight before counting, so a sentence that opens with `"` and closes with `”`
    is balanced. Balance is checked on `len(parts) % 2 == 0` — an even number of split parts
    means an odd number of delimiters.
  - **The whole-sentence path is deliberately untouched.** An unbalanced `shows` that occurs
    verbatim on the page is still a confirmation: the sentence was found, dangling quote and
    all, which is exactly the evidence the original test asks for. Refusing it too would have
    been a second, unasked-for behaviour change, and `test_an_unbalanced_quote_still_confirms_on_the_whole_sentence`
    pins it.
  - **Failing closed here is cheap and I checked rather than assumed.** All 3,932 `shows`
    fields in `sources/scores/` pair off today, so no live source loses a confirmation. The
    cost of the rule is borne entirely by future typos, and an inches mark or a nested quote
    inside a sentence will now cost that source a fragment confirmation until somebody fixes
    the sentence — which is the correct direction for a leg that writes dates.
- alternatives rejected:
  - **Repair the sentence — treat the dangling segment as running to the end.** It invents a
    fragment the curator did not delimit. If the tail is prose rather than a quote, the
    invented fragment will not be on the page and the source drifts for the wrong reason; if
    it is a quote, we guessed right by luck.
  - **Keep dropping the dangling segment but require one more long fragment.** Still confirms
    a source while an entire recorded claim goes untested, which is the reviewer's point, and
    no threshold repairs that.
  - **Refuse unbalanced `shows` on both paths (whole sentence too).** Wider than the finding
    and strictly worse: it would drop confirmations that are literal text matches.
  - **A `check_verification` gate that rejects unbalanced `shows` at the corpus level.** Worth
    having, but that is a corpus-wide gate and a separate change, and with 0 current
    violations it would have landed with nothing to catch. Not this job.

## Risks / unsure about

- **The rule is currently unexercised outside tests.** 0 of 3,932 `shows` fields are
  unbalanced, so the live before/after is unchanged and the whole defense of this commit is a
  fixture plus a corpus count. That is also the honest state of it: the guard exists for the
  first curator who types an inches mark.
- **A legitimate unbalanced sentence now loses its fragment confirmation.** `a 24" rack` or a
  nested quotation will flip a source from confirmed to drifted rather than to a false stamp,
  which is the right direction, but it is a real false-drift mode and it is silent — the
  report says `drifted`, not "unreadable `shows`". If that shape shows up in practice the
  report line should distinguish it.
- **Carried from round 1, unchanged:** the brief's `atropos` figure ("nine fragments of which
  two occur") did not reproduce through `build.fetch_source` — all its repo-page fragments
  occur live, and what keeps `atropos` drifting is the 24-character floor on its API citation,
  not a partial match. The reviewer settled this ("the live atropos fragment counts can differ
  from the earlier probe"), and the partial-match refusal is pinned by
  `test_a_partial_fragment_match_still_drifts` rather than by a live product in this sample.
- **Both measurement runs are live fetches, 6 minutes apart.** The delta is internally
  consistent (drift −11, shows-confirmed +11, same URLs, identical transients) and reproduces
  round 1's numbers from ~25 hours earlier, so I do not think page churn is in these figures.

## Response to round N-1

|---|---|---|---|---|
| Unbalanced quotes fail open: the parser drops the final claim, and the new test confirms a source whose `shows` says the banner reads `"Archived` against a body with no archive banner (`build/reverify.py:259`) | ✅ | | | `_shows_fragments` returns `None` on an odd delimiter count (`build/reverify.py`, commit `1ff3e872`); `_fragments_confirm` declines before any containment test. `test_an_unbalanced_quote_refuses_the_fragment_path` now asserts `is None` and `result.drifted == [("openness", "https://a/LICENSE")]` on the exact body that previously confirmed. `test_an_unbalanced_quote_still_confirms_on_the_whole_sentence` keeps the untouched path honest. `uv run pytest -q -n 8` → 2050 passed, 1 skipped; `build.validate` → 0 errors. Live effect nil and checked, not assumed: 0 of 3,932 corpus `shows` fields are unbalanced, and the re-run before/after table is identical to round 1's. |

Issue #348, opened 2026-08-22
Retirement review: 4 assets with no reviewed consumer after Phase 0b


ccerv1:
Tracking issue for the retirement candidates surfaced by the Phase 0b deployed-model audit
(PR #347). **This is a review record, not an approval to delete.** Section 17 of
`docs/architecture/data-architecture.md` requires explicit maintainer authorization, a consumer
inventory, and a stated rollback path before any deletion. The inventory exists (this audit);
the decision does not.

After reading all 41 deployed model definitions, four **deployed** assets have no reviewed
consumer — no in-repo code reader, no consumer among the org's twenty notebooks, and no deployed
platform model reads them. `external` remains `unknown`.

- `currentai.scores.ossd_coverage`
- `currentai.signal_github.product_adoption` — ⚠️ holds the route precedence `pypi > huggingface > stars` in SQL. **Must not retire before `registry.adoption_routes` compiles that ordering**, or the ordering is lost silently.
- `currentai.signal_semanticscholar.paper_citations` — citation instrument declared in `signal_routing.yaml`, not yet consumed. Pre-positioned.
- `currentai.signal_artificialanalysis.model_evaluations` — platform description says "held deliberately unjoined to gap-map products". Unread by design.

The staged `signal_packages.product_adoption` is **not** a candidate: it is not deployed (issue
#314), and a model not in service cannot be retired.

Each asset's `warehouse/assets.yaml` entry references this issue in `retirement_issue`.

ccerv1:
Three of the four have been dispositioned since this was written. `scores.ossd_coverage`, `scores.repos_summary` and `catalog.stack_map` are all in `warehouse/audits/externalization.json` as `frozen-without-producer` under ADR-003 (#406) — retained live, no repo producer, recorded.

`currentai.signal_github.product_adoption` is the one still outstanding, and it is the one that needs the care: it holds the `pypi` route precedence, so it is not a candidate for the same treatment without deciding where that precedence lives first.

Narrowing this issue to that single asset.

ccerv1:
Cleared by inspection, 2026-09-13. Closing.

The blocker this issue names is that the route precedence `pypi > huggingface > stars` lives only inside `signal_github/product_adoption.sql`, and must be compiled into `registry.adoption_routes` before those tables retire or the ordering is lost silently.

[...]

Issue #314, opened 2026-08-16
Redo the npm/crates signal bridge as a sequenced rollout (was #266)


ccerv1:
#266 is closed as superseded — too stale to rebase safely, and its two irreversible steps never happened. The branch `carl/registry-bridge` is preserved for reference. This issue carries the work.

## Why the sequence matters

Two steps cannot be undone by reverting a commit: **deploying the new package dataset**, and **deleting `signal_pypi`**. #266 carried both alongside a large repo diff, so a single review had to cover reversible edits and irreversible platform actions at once. Splitting them is the whole point of redoing it.

## Order

1. **Correct the platform-mirror manifest** — #313. It listed `packages_product_adoption.sql` against two tables it does not implement and omitted the real `signal_packages.product_adoption`, so the mirror could not serve as the reference for step 2. *(In review.)*
2. Confirm all three `signal_packages` model sources against the platform's released revisions.
3. Deploy and run the new package dataset. **Do not delete `signal_pypi`.** Both live in parallel.
4. **Re-sync the mirror against what was actually deployed.** Refetch the three released `signal_packages` models into `warehouse/platform-mirror/`, replace their `status: staged` entries with full provenance — `model_id`, `revision`, `hash`, `local_sha256`, `synced_at` — and run `tests/test_platform_mirror.py`. `test_deployed_entries_have_revision_provenance` starts covering them the moment `status: staged` comes off, so a half-converted entry fails rather than passing quietly. Do this **before** any consumer is repointed, so the reference is current while the routes are being changed.
5. Compare old PyPI results against the new combined results — per product, not in aggregate. A product that gains or loses a band here is the finding.
6. Open a fresh PR from current `main` with only the still-needed routing, schema, artifact-ID, exemption and score changes. `check_routing` fixes from #266 already landed; do not re-apply them.
7. Flip routes to bridged **only** once the deployed tables answer correctly.
8. **GATE — migrate the `signal_github.product_adoption` consumer.** The deployed model reads `currentai.signal_pypi.package_downloads` in its `already_measured` CTE, which is what stops the stars fallback re-banding products that already have a download signal. Repoint it at the new table, deploy that revision, and confirm the excluded set is unchanged. Until this is done and verified, step 9 silently corrupts adoption for every product the stars route touches.
9. Delete `signal_pypi` **last**, after the gate above plus parity and consumer checks.

[...]

PR #560, opened 2026-09-13
Write down the execution order for the signal_packages bridge

ccerv1:
## Summary

An agent is executing #314, #517, #348 and the #163/#444 request on the platform right now. The sequence it needs existed only in my head and on closed PR #266 — which is a bad place for it, given the sequence ends in an irreversible drop and carries two constraints that fail *silently* if taken out of turn.

- **#348 goes before the retirement, not after.** The route precedence `pypi > huggingface > stars` is implemented only in `signal_github/product_adoption.sql`. If the tables retire before `registry.adoption_routes` carries it, the ordering is lost and nothing fails. `data-architecture.md` §4.1 already names this as the concrete AD-1 violation; the runbook says what to do about it.
- **The reader gets repointed before the drop.** `signal_github.product_adoption` reads `signal_pypi.package_downloads` in its `already_measured` CTE, so dropping first makes the stars fallback re-band products that already have a download signal.
- **The registry static model is a delete-and-recreate**, because it gains `not_primary_channel` and a static model cannot gain a column by re-upload.
- **#517 cannot go first.** Three of its six assets have no governed reader until #314 deploys, and a dependency contract must be reachable from a governed root, so reclassifying early puts them out of scope in both manifests.
- Records the three diff shapes that are legitimate, so the old-versus-new comparison has a pass condition.

## Test plan

- `uv run pytest -q` → 2059 passed
- Docs only; cross-linked from `deploy-models.md`

### DIFF
--- a/docs/operations/deploy-models.md
+++ b/docs/operations/deploy-models.md
@@ -258,5 +258,7 @@ datasets' run history first" — it can mean the warehouse is stale rather than 
 
 - `docs/reference/evidence-and-freshness.md` — what the chain computes and the gates over it
 - `docs/operations/publish-map.md` — serializing and publishing the notebook
+- `docs/operations/deploy-signal-packages.md` — the #314 bridge, its two ordering constraints and
+  the one irreversible step
 - `docs/operations/refresh-data.md` — loading a fetched CSV into the warehouse as a static model
 - `warehouse/assets.yaml` and `docs/architecture/data-architecture.md` — the asset inventory and the model families it covers
--- a/docs/operations/deploy-signal-packages.md
+++ b/docs/operations/deploy-signal-packages.md
@@ -1,0 +1,78 @@
+# Deploy `signal_packages`, then retire `signal_pypi` (maintainer, MCP write)
+
+The npm/crates bridge of issue #314, and the two things that must happen around it: #348 before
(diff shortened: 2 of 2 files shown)


ccerv1:
**Rewritten against what actually happened.** This PR was opened to get the sequence out of one person's head before an agent executed it. The agent executed it, and three of the document's claims did not survive contact.

- **Step 4's advice was wrong and expensive.** It said the fix for the release deadlock was `columns=` on `signal_github.artifact_state`. That was applied on 13 September. It did not unblock the consumer, and because a model's determinism verdict depends on what it *reads*, it locked four further models nobody had touched. All four were deleted and recreated. The step is now marked void, and the repoint it asks for is not needed at all, since #562 retires that model rather than migrating it.
- **`not_primary_channel` exists.** It shipped in #563 the same day. The section that said nothing produces it is replaced with what it means, which is narrower than the name suggests.
- **Steps 1 to 3 are history, and #314 and #348 are closed.**

A determinism section now carries the lesson, which is the reason to keep this document rather than close it: the refusal names the model you are releasing rather than the upstream that moved, a byte-identical control is what isolates the two, there is no rename path so recreation is the only remedy, and recreation restores releasability without immunizing anything.

`uv run python -m build.validate` reports 0 errors.

Issue #562, opened 2026-09-13
Adoption has three layers: signal discovers, observations monitor, evaluation bands


ccerv1:
Adoption work keeps landing in the wrong part of the warehouse, and #314 as written made it worse. This records the separation the repo is migrating toward, and what the remaining work actually is. Revised after review; the first version got three things wrong and they are marked below.

## The separation

This is a **logical separation and a migration goal**, not a description of what every namespace currently holds. `assets.yaml` already classifies by kind rather than by namespace, and the legacy `signal_*` namespaces contain both observations and evaluations.

**Discovery.** Finds things that might *become* products. Reads a universe, not our declarations: `signal_goodailist.repo_catalog`, `signal_hfhub.model_universe`, `signal_openrouter.models`. Discovery may legitimately bucket raw activity to rank candidates — that is not a governed product assessment and nothing here forbids it.

**The curated catalogs, `sources/`.** The source of truth for what is a product and what its artifacts are. Uses discovery to *identify* candidates; always separate from it.

**Observation.** Monitors what we already declared. Reads `registry.product_artifacts` and watches those artifacts on a schedule — licence, stars, downloads, liveness. `observations.product_adoption_current` is this layer, and its classification note is explicit: *artifact-level raw observations, no band and no instrument_type; banding and product-level aggregation are evaluation concerns*.

**Evaluation.** Computes the governed band, once, from the compiled `registry.adoption_*` routing: `build/adoption_measurements.py` into `evaluation.product_adoption_measurements`.

**The test for a model is what it reads.** A universe means discovery. `registry.product_artifacts` means observation. By that test `signal_github.artifact_state` and `signal_packages.downloads_daily` are observation wearing a discovery name — both open with a roster query against `registry.product_artifacts`.

## What is wrong now

Governed product-adoption banding is implemented a second time inside signal datasets:

| asset | status | replacement |
|---|---|---|
| `signal_github.product_adoption` | `compatibility` | `observations.product_adoption_current` |
| `signal_huggingface.product_adoption` | `compatibility` | `observations.product_adoption_current` |
| `signal_packages.product_adoption` | `staged` | **none** |

The third was deployed 2026-09-13 under #314 — a new model of a kind already marked for retirement, carrying no `replacement`, so it has no exit. `data-architecture.md` §11.1 (~line 1454) says the set is "retired together once the centralized evaluator is proven, not one at a time".

## The remaining work, and what each step actually costs

[...]

PR #563, opened 2026-09-13
Declare when a package is not how a product ships

ccerv1:
## Summary

Two deployed warehouse models read a `not_primary_channel` column that **nothing produced** — absent from `registry.product_artifacts`, from `sources/products/*.yaml` and from every `build/*.py`, and found by a repo-wide search only in the two `signal_packages` SQL files. Stubbed NULL, `hexabot` and `yomo` band on a package that is not how they ship and fall 2→1, against the 2026-08-14 ruling holding both at 2.

This builds the declaration those models were written against, and makes the repo's own evaluator honour it too.

- **The artifact wrapper takes an optional `not_primary_channel`**, where the value is the reason. Same shape as `artifact_exceptions`: presence is the exemption, the value says why. The wrapper's description already said adding a second key here would be a schema change rather than a data change; that sentence is now updated, because it has stopped being true.
- **`build/serialize_registry.py` emits it** as a column on `registry.product_artifacts`, so the models already reading it stop resolving NULL.
- **Declared on exactly two artifacts.** Hexabot's npm entry is the embeddable widget a site drops into a page, not the self-hosted Docker stack the product is. Yomo's crate is the Rust SDK for a runtime shipped as a Go binary. Each reason says what the package is and why its installs are a different population from the product's.
- **The evaluator honours it.** `build/adoption_measurements.py` previously reduced declarations to sets of artifact kinds and summed every matching observation, so the declaration would have moved the warehouse band and left the repo's disagreeing.

**Out of the sum, never out of the table.** The artifact stays in `registry.product_artifacts`, its observation stays in `observations.product_adoption_current`, and its per-artifact downloads stay readable. Only the banded figure changes, and `non_primary_artifacts` on the emitted row names what was left out so the exclusion is visible rather than something a reader has to diff the registry to find.

**Why not the ledger.** `sources/resolution_ledger.yaml` has `product_membership` / `not_member_of`, which says the measurement is not the product's at all. That is a stronger and different claim. Here the artifact *is* the product's and its downloads are still published; only the band excludes it.

**Route applicability, not fallthrough.** A kind whose every artifact is declared non-primary names no channel the product ships through, so it never becomes the authoritative route. That is a fact about declarations, which is what applicability is made of. The no-fallthrough rule forbids substituting a weaker route when an authoritative one merely went unobserved — an accident of collection — and that behaviour is unchanged: a kind with a primary artifact and no observation still emits no row. Both products here are all-non-primary on their package kind and fall through to stars, which is where the ruling puts them, and what the warehouse model already does with the same declaration.

## Follow-up this PR deliberately does not do

`registry.product_artifacts` gains a column, and a static model cannot gain one by re-upload. The platform side is a **delete and recreate**: the publisher's `resolve_static_models` creates whatever is missing, so deleting the model lets the next `main` publish rebuild it with the column. The publisher's own delete path only tidies models that never materialized — dropping one that holds rows is deliberately a human's call — so this is a maintainer step, sequenced with the merge.

## Test plan

- `uv run pytest -q` → 2075 passed, 1 skipped
- `build.check_adoption` → 664 products, OK; `build.check_instrument` → 664 records, OK
- Serializer output checked directly: `product_artifacts` carries `not_primary_channel`, and exactly two rows have a value
- No product other than `hexabot` and `yomo` changes

Step 1 of #562.


ccerv1:
## What I did

Built the `not_primary_channel` declaration the two deployed `signal_packages` models were
written against, and made `build/adoption_measurements.py` reach the same answer the warehouse
will. One commit on `loop/not-primary-channel-2026-09-13-a1`, `3bd38ab3`, nothing amended.

- files touched:
  - `docs/schemas/product.schema.json` — `definitions.url` gains an optional
    `not_primary_channel` (`type: string`, `minLength: 20`, matching `artifact_exceptions`).
    The wrapper's own description no longer says nothing has claimed a second key; it names
    this one and keeps the "the next one is still a schema change" half, which stays true.
  - `build/serialize_registry.py` — new `_artifact_entries()` reads an artifact block as
    `(url, reason)` pairs (still tolerating a bare string, as `_urls` did); `product_artifacts`
    gains a `not_primary_channel` column, emitted verbatim, empty elsewhere. Module docstring
    table updated.
  - `sources/products/hexabot.yaml`, `sources/products/yomo.yaml` — the two declarations. Each
    reason says what the package IS and why its population differs from the product's.
  - `build/adoption_measurements.py` — `load_inputs()` now returns a 6-field `Inputs`
    NamedTuple: `declared_artifacts` holds the kinds a product ships *through*, and
    `non_primary_artifacts` holds the `(kind, id)` pairs declared not to be one.
    `measurements()` takes the second as a new parameter, drops those observations from the
    sum, and stamps a new `non_primary_artifacts` column on the row.
  - `build/serialize_evaluation.py`, `build/adoption_reconciliation.py`,
    `build/axis_assessments.py`, `build/check_channel_authority.py` — the four `load_inputs`
    unpack sites; the two that build measurements now pass the new input through.
  - `tests/test_serialize_registry.py` — +2 tests (the reason rides on the artifact row and the
    column is rectangular; the declared set is exactly the two artifacts that were ruled on),
    and the arXiv row fixture gains the new key.
  - `tests/test_adoption_evaluation.py` — +6 tests (below), seven fixture unpacks, and the two
    pinned constants with a dated comment in the file's existing style.
  - `docs/reference/adoption.md` — a subsection under the under-coverage remedies: what the
    declaration means, what it does not touch, and why it is not a ledger ruling.
  - `docs/architecture/data-architecture.md` — the new column in the
    `evaluation.product_adoption_measurements` block, with what it is for.
  - `CHANGELOG.md` — one Added entry, naming the platform consequence.
- commands run (all foreground, all waited on):
  - `uv run python -m build.validate` → **0 errors**, 2 pre-existing warnings (overlapping
    `model_families` glob patterns, untouched). Run on `main` first, identical then.
  - `uv run pytest -q -n 4` → **2074 passed, 2 skipped** (7m20s), on the final tree, read before
    committing. Plus `uv run pytest -q -m serial` → 1 passed, matching what CI splits out. (The
    unsharded `pytest -q` takes 12m45s, past this session's 600s foreground ceiling; `-n auto`
    is what `validate.yml` itself runs.)
  - `uv run python -m build.check_adoption`, `check_instrument` — captured on `main` before any
    edit and again on the final tree; `diff` says **byte-identical** both times.
  - `uv run python -m build.adoption_measurements --allow-dirty [--json]` on `main` and on the
    branch, rows compared product by product.
  - `uv run python -m build.serialize_registry`, then read the emitted CSV, to prove the column
    exists rather than assume it. `uv run python -m build.goldens --check` → the expected stale
    notice (sources/ moved), not a producer failure. `uv run python -m build.check_corpus_diff
    --base 7a1c96f3` → no stage moves, no untouched-product row changes.
  - `build.serialize --date ci-dry-run` ran clean; it rewrote `build/notebook_data.json`, which
    I reverted — a generated file, not mine to commit.

### The measurement (frozen Phase-2 baseline, same observation set both sides)

| | rows | hexabot | yomo | any other row |
|---|---|---|---|---|
| `main` | 377 | no row (npm route wins, unobserved) | no row (crates route wins, unobserved) | — |
| branch | 379 | `github.stargazers_count`, 1,204 stars, **level 2** | `github.stargazers_count`, 1,920 stars, **level 2** | unchanged |

Compared field by field with `declaration_version_id` (commit-scoped) excluded: two rows added,
none removed, and zero existing rows changed a value — the only difference elsewhere is the new
`non_primary_artifacts` column, empty on all 377. `check_adoption` 664 products `[OK]` and
`check_instrument` 664 records `[OK]` on both sides, byte-identical output.

### The six evaluator tests

`test_non_primary_artifact_leaves_the_sum_and_nothing_else` (two PyPI packages, one declared:
route still applies, figure is the primary alone, excluded id still nameable) ·
`test_a_kind_that_is_all_non_primary_falls_through_to_the_next_route` (with the control: the
same observations without the declaration band the package at 40 downloads) ·
`test_an_unobserved_primary_artifact_still_produces_no_row` (the boundary — a declared primary
package that was not collected is still UNMEASURED, not fallen through) ·
`test_load_inputs_reads_the_declaration_off_the_registry` · `test_the_two_declared_products_
band_on_stars_at_two` · `test_no_other_product_carries_an_exclusion`.

### For the PR body

`registry.product_artifacts` gains a column. A static model cannot take a new column on a
re-upload, so the platform side is a **delete and recreate** of the `registry` static model, and
**this PR does not do it** — it is the maintainer step the job file holds. The same applies to
`evaluation.product_adoption_measurements`, which gains `non_primary_artifacts` on the same
terms. The two `signal_packages` models need no change at all: they already read the column, and
this is what stops them reading NULL.

## Why

- decisions:
  - **An all-non-primary kind fails route applicability; it does not match with a null figure.**
    This was the job's first open question. Applicability is a fact about DECLARATIONS, and
    `not_primary_channel` is a declaration: a kind every one of whose artifacts is declared
    non-primary names no channel the product ships through, so it never becomes authoritative.
    The no-fallthrough rule forbids something else — substituting a weaker route when an
    authoritative one was merely not OBSERVED, an accident of collection — and that case is
    untouched and pinned by a test. Three things agree with failing the route: the spec in
    `product_adoption.sql` ("both products then have no primary package channel at all and fall
    through to the stars instrument legitimately"), remedy 3 in `adoption.md` (abstain on the
    minority channel rather than band on it), and the recorded assessments, which are
    `stars_fallback` level 2 on both. Matching with a null figure would emit a row that measures
    nothing, keep both products off stars forever, and make reconciliation read an instrument
    mismatch against their own records.
  - **The minimum reason length matches `artifact_exceptions` at 20.** The second open question.
    A package reason has to say what the package IS, which is the same work the exemption
    reasons do; both declarations run well past 20 characters, so the floor is a floor and not a
    constraint on how these read. Copying the number keeps one answer to "how long is a reason".
  - **The exclusion is visible on the row, not only in stdout.** `non_primary_artifacts` mirrors
    `artifacts_excluded`/`kinds_excluded` on the warehouse model, which is the only way a reader
    of `hexabot`'s stars row can tell why the npm route did not win. Note the shape of the two
    real cases: because both fall through, there is NO package row to hang the exclusion on, so
    a column that only recorded "what this sum left out" would have been null in exactly the two
    cases that motivated the field.
  - **Two readings of one input, not one clever one.** `load_inputs` walks the registry rows once
    and produces both the eligible kinds and the excluded artifacts. Per-artifact ids have to
    reach `measurements()` for the mixed case (one primary npm package beside a non-primary one),
    which no product is in today and which a kind-level filter alone would get silently wrong.
  - **Not the ledger.** `product_membership: not_member_of` says the artifact's measurement is not
    the product's at all. These packages are the products' own, their downloads are real and stay
    published per artifact, and only the banded sum excludes them. Recording the weaker fact as
    the stronger one would retire measurements that should stay visible.
- alternatives rejected:
  - Filtering the non-primary observations at the load boundary (before `measurements`). It sums
    correctly and is invisible: nothing on the row, and it would read as "never observed".
  - Keeping `load_inputs`'s 5-tuple and giving `measurements` a defaulted keyword only. The two
    production callers that build the real table would then have silently kept the old behaviour
    — the exact half-fixed state this job exists to avoid.
  - Excluding the artifact from `registry.product_artifacts` entirely. That is the ledger claim,
    and it would take the downloads off `signal_packages.downloads` too.

## Risks / unsure about

- **Two static models need recreating, not one.** The job's success criterion names the
  `registry` model; `evaluation.product_adoption_measurements` gains `non_primary_artifacts` and
  needs the same treatment. I judged a row-level record of the exclusion to be required ("visible
  in what the builder emits rather than silent") and worth the second migration; if a reviewer
  reads the criterion as satisfied by the stdout summary alone, dropping the column is a small,
  self-contained revert (COLUMNS, one row key, one doc block, two tests, one digest re-pin).
- `MEASUREMENT_COUNT` and `MEASUREMENTS_DIGEST` in `tests/test_adoption_evaluation.py` are
  re-pinned (377 → 379, new digest) with a dated comment. That is the intended signal for a
  declaration change, but it is a golden move and should be read as one.
- The `yomo` reason states the runtime ships as a Go binary, which is the job's own
  characterization and the ruling's; the score record describes a Rust core with an
  Apache-2.0 `Cargo.toml`. The claim that survives either way, and the one the reason leans on,
  is measurement-population identity: crate downloads count Rust builds linking the SDK. If a
  reviewer wants the prose to stop asserting the binary's language, it is a one-line edit.
- `build/check_artifacts.py` could not be run here — it needs `OSO_API_KEY`. Same on `main`.
- `tests/goldens/corpus.json` is deliberately NOT regenerated: the generated-files guard blocks
  it on a proposal PR, `test_goldens` skips equality when the sources fingerprint differs, and
  `check_corpus_diff` (run, clean) is the gate that covers content on a PR.

## Response to round N-1

|---|---|---|---|---|
| n/a — round 1 | | | | |

PR #564, opened 2026-09-13
Read package observations from all three registries

ccerv1:
## Summary

`observations.product_adoption_current` normalises the fetchers into artifact-level, band-free observations, and its package arm was PyPI-only — its own header said the successor was "deliberately NOT read yet". It now reads `currentai.signal_packages.downloads`, so the observation layer sees npm and crates alongside PyPI.

- **`artifact_kind` is projected, never literal.** The arm hardcoded `'pypi'` for both `channel` and `artifact_kind`. The source carries `pypi`/`npm`/`crates` per row, and §4.3 already fixes the channel vocabulary as `github|huggingface|pypi|npm|crates|other`, so one column serves both. Live check: all 187 package rows carry `channel == artifact_kind`, 169 pypi / 17 npm / 1 crates.
- **A replacement, not an addition.** The successor carries the PyPI leg too, so the old arm had to go rather than sit beside it. Proof it is a clean swap: the 169 PyPI observations are identical across both versions — same keys, same `raw_value`, same `observed_at`, zero diffs — and the three non-package arms are byte-identical at 835 rows either side.
- **`not_primary_channel` is deliberately not read here.** The source projects it; this model ignores it. It is a curation fact about how a product ships, and `build/adoption_measurements.py` applies it from `registry.product_artifacts`, leaving the artifact out of a sum rather than deleting its measurement. Reading it here would put curation in the raw layer.
- Manifests and the DAG travel with the read, since `reads` / `read_by` / `required_by` are re-derived from the tree by the scope gates.

## What moved

Observation layer 1004 → 1022 rows. The 18 new rows are 17 npm and 1 crates, over 16 products that had **no package observation at all**.

Measured by running both versions of the SQL in one session and the rollup over each row set with identical repo inputs, so #563's declarations apply to both sides and the observation set is the only variable. Of the 14 products that gain a band, **twelve agree with the level the repo already records**, one rises, one falls.

- `mcp-inspector` **rises** 3 → 4. Recorded at 905,396 on 2026-08-12, just under the 1M boundary; it has since crossed it at 1,043,305.
- `n8n` **falls** 4 → 3, and this is a finding rather than a defect. Its own recorded note says the npm package "is a minority install path, so banding on the registry package alone would have measured the wrong channel and given a materially lower level" — which is exactly what #563 gave a name to. Its npm artifact is a `not_primary_channel` candidate. Out of scope here because this PR moves plumbing rather than declarations, but it is now visible, which it was not before.
- `hexabot` and `yomo` gain observations and **do not fall**. Both are declared `not_primary_channel` as of #563, so the package route never wins and both stay on stars at level 2 — same row, same level, before and after.
- `agent-infra-sandbox` at level 1 agrees; the repo had already moved it from 2 by hand and the warehouse now says the same.

## Grain

`beeai` is the case that would have caught a collapsed grain: it declares `beeai-framework` on **both** npm and PyPI, and produces two rows distinguished by `artifact_kind` rather than one that lost a registry. Its band does not move, because route precedence puts `pypi` ahead of `npm`; the npm row is published without being summed into it. `mastra` declares two npm packages and contributes one row each, summed to 7,857,689 under `sum_usage_across_artifacts`.

The coverage guard never fired across all 1022 rows and `product_type` is non-null throughout, which is the proof that every npm and crates artifact is declared — the successor's roster is `registry.product_artifacts` filtered to the three package kinds, so it cannot emit an undeclared one.

## Downstream that could have narrowed, checked by running it

`build/adoption_measurements.py` filters contributing observations on `artifact_kind` and `metric_type`, never on `channel` or `source_dataset`, so the npm and crates routes start matching rows they could not see before. Nothing else in `build/` keys on `channel == 'pypi'` or on `source_dataset`. `build/check_artifacts.py` and `signal_github/product_adoption.sql` still read `signal_pypi.package_downloads`, neither is in this change's path, and that table's dependency contract stays.

`sources/signal_routing.yaml` is deliberately untouched: its npm and crates sources still say `bridged: false`. Flipping that turns on `check_instrument`'s artifact precondition for every product recording `usage_volume` on a package registry, which is a curation change and a later step of #562. Nothing in the observation or rollup path reads `bridged`.

## Test plan

- `uv run pytest -q` → 2075 passed, 1 skipped
- `uv run python -m build.validate` → 0 errors; `check_adoption` 664 products OK; `check_recipe` 0 failures
- `build.observation_snapshot`'s strict digest accepts the new row set, so the identity path takes npm and crates as vocabulary values without a contract change

Step 2 of #562. Deploying the model to the platform is a maintainer step, sequenced with the merge.


ccerv1:
## What I did

- files touched:
  - `warehouse/models/observations/product_adoption_current.sql` — the package arm now reads
    `currentai.signal_packages.downloads` instead of `currentai.signal_pypi.package_downloads`.
    `'pypi' AS channel` and `'pypi' AS artifact_kind` both become the source's own `artifact_kind`
    column, projected once per alias; `source_dataset` becomes `signal_packages` and `source_table`
    becomes `currentai.signal_packages.downloads`. Columns whose names are stable across the two
    (`package`, `downloads_30d`, `last_day_seen`) are untouched, and the arm wants nothing that was
    renamed: it never read `missing_from_pypi`, so there is no `missing_from_registry` to rename,
    and it never read `adoption_level`, which the successor does not have. `WHERE downloads_30d IS
    NOT NULL` is unchanged and still says the right thing — a declared artifact the fetch has not
    reached, and a `missing_from_registry` 404, both carry a NULL window and drop out here.
    Header rewritten: the normalised-sources block, the line claiming the successor "is deliberately
    NOT read yet", the artifact_kind vocabulary sentence, plus two new notes (why this arm alone
    projects rather than hardcodes, and why `not_primary_channel` is deliberately not read here).
  - `warehouse/assets.yaml` — `observations.product_adoption_current.reads` swaps
    `signal_pypi.package_downloads` for `signal_packages.downloads`; `signal_packages.downloads.read_by.models`
    gains the observations model. Both lists are mechanically re-derived and gated
    (`tests/test_assets_inventory.py::test_reads_matches_the_derived_graph`, `::test_read_by_matches_the_derived_graph`).
  - `warehouse/dependencies.yaml` — `currentai.signal_pypi.package_downloads.required_by` drops the
    observations model. It keeps two readers (`build/check_artifacts.py`,
    `warehouse/models/signal_github/product_adoption.sql`), so the contract stays reachable and gate 4 holds.
  - `docs/architecture/current-state-dag.md` — regenerated from `build.assets.render_dag`. Net effect
    is two lines: `signal_packages.downloads --> observations.product_adoption_current` added in view 1,
    the `signal_pypi.package_downloads --> observations.product_adoption_current` dependency edge removed
    from view 2.
  - `CHANGELOG.md` — one entry under Unreleased / Changed.

- commands run:
  - `uv run python -m build.validate` → `0 error(s), 2 warning(s)` (the two pre-existing
    `model_families` pattern-overlap warnings, unrelated).
  - `uv run pytest -q` → `2075 passed, 1 skipped in 769.26s`, exit 0 (run in the
    foreground on the committed tree)
  - `uv run pytest -q -n auto -m "not serial"` + `uv run pytest -q -m serial` (the CI split) →
    2074 passed, 1 skipped / 1 passed.
  - `uv run python -m build.preflight` → all 17 locally-runnable CI steps pass, 3 accounted for
    (`check_parity` and `check_artifacts --live` need the warehouse; they run in their own workflows).
  - Live warehouse runs of the model SQL, before and after, plus the rollup over both — the
    measurements below.

## Why

- decisions:
  - **Projected, not literal.** `artifact_kind AS channel, artifact_kind` in one SELECT list. §4.3
    fixes the channel vocabulary as `github|huggingface|pypi|npm|crates|other`, and the successor's
    `artifact_kind` values are `pypi|npm|crates`, so the source column is simultaneously a valid
    channel and a valid artifact kind. Live check: the 187 package rows carry
    `channel == artifact_kind` for every row, 169 `pypi` / 17 `npm` / 1 `crates`.
  - **Replacement, not addition.** The successor carries the PyPI leg as well, so the old arm had to
    go rather than sit beside the new one — keeping both would have given every PyPI package two
    rows on one grain. Proof that it is a clean swap: the 169 PyPI observations are identical across
    the two versions, same keys, same `raw_value`, same `observed_at`, zero diffs, and the three
    non-package arms are byte-identical (835 rows either side).
  - **`not_primary_channel` stays out.** The successor projects it and this model ignores it. It is a
    curation fact about how a product ships, and `build/adoption_measurements.py` already applies it
    from `registry.product_artifacts` — dropping the named artifact from the summed figure while the
    observation itself stays published. Reading it here would put the exclusion in the raw layer and
    silently delete a measurement rather than leave it out of one sum. Measured consequence below:
    `hexabot` and `yomo` keep their npm/crates observations AND stay on stars at level 2.
  - **Manifests and the DAG travel with the read.** `reads`/`read_by`/`required_by` are re-derived
    from the tree by gates, so they are part of the change rather than follow-up bookkeeping.
  - **`sources/signal_routing.yaml` deliberately untouched.** Its `npm` and `crates` *sources* still
    say `bridged: false` / `table: null`. Flipping that is the routing step of #562, not this one:
    `bridged` turns on `check_instrument`'s artifact precondition for every product recording
    `usage_volume` on a package registry, which is a curation change. Nothing in the observation or
    rollup path reads `bridged` — `select_route` keys on `artifact_kind` — so the npm and crates
    routes resolve to an observation from this model's first run regardless. The model header now
    says this out loud so the stale flag is a recorded lag rather than a contradiction someone finds.

- alternatives rejected:
  - Adding the package arm alongside the PyPI arm and de-duplicating in `ranked`. It would have
    worked (the tie-break is deterministic) and it would have been wrong: two source rows for one
    artifact, with the winner decided by `source_table` alphabetical order.
  - Filtering `missing_from_registry = false` in the WHERE. Redundant — a missing artifact has a NULL
    `downloads_30d` and is already excluded — and it would have coupled this arm to a column name
    that just changed.
  - Flipping `signal_packages.downloads` in `warehouse/assets.yaml` from `staged` to `active`. See
    Risks.

## Measured band effect (2026-09-13, live)

Computed, not repeated from the brief. Method: run the model SQL at HEAD and the edited SQL against
the warehouse in the same session, then run `build.adoption_measurements.measurements` over each row
set with identical repo inputs (so #563's declarations apply to both sides and the only variable is
the observation set).

Observation layer: 1004 → 1022 rows. The 18 new rows are 17 npm + 1 crates, over **16 products that
had no package observation at all**: `agent-infra-sandbox`, `axon`, `context7`, `filesystem-mcp`,
`gemini-cli`, `hexabot`, `mastra`, `mcp-inspector`, `mcp-typescript-sdk`, `n8n`, `notion-mcp`,
`openclaw`, `opencode`, `orama`, `playwright-mcp`, `yomo`. (`beeai` is the seventeenth npm declarer
and already had a PyPI observation, so it gains a row but not a channel.)

Rollup: 535 → 549 measurement rows. **No product loses a row, none falls, none changes route.**
The 14 gains, against the level the repo records:

| product | npm/crates 30d | measured | recorded | |
|---|---|---|---|---|
| `mcp-typescript-sdk` | 190,532,060 | 5 | 5 | agrees |
| `playwright-mcp` | 22,288,222 | 5 | 5 | agrees |
| `openclaw` | 12,150,078 | 5 | 5 | agrees |
| `opencode` | 8,758,790 | 4 | 4 | agrees (recorded instrument was `reported_traction`; now countable) |
| `mastra` | 7,857,689 (two packages summed) | 4 | 4 | agrees |
| `orama` | 4,499,682 | 4 | 4 | agrees |
| `context7` | 3,558,420 | 4 | 4 | agrees |
| `filesystem-mcp` | 2,782,041 | 4 | 4 | agrees |
| `gemini-cli` | 1,343,244 | 4 | 4 | agrees |
| `mcp-inspector` | 1,043,305 | 4 | 3 | **rises** — recorded 905,396 on 2026-08-12, just under the 1M boundary; it has since crossed it |
| `notion-mcp` | 600,217 | 3 | 3 | agrees |
| `n8n` | 438,134 | 3 | 4 | **falls** — see below |
| `agent-infra-sandbox` | 4,837 | 1 | 1 | agrees; the repo had already moved this one from 2 to 1 by hand ("the level was previously placed at 2"), and the warehouse now says the same |
| `axon` | 103 | 1 | 1 | agrees |

Twelve agree, one rises, one falls. The one that falls is the interesting row and is not a defect in
this change: `n8n`'s recorded note bands on Docker Hub pulls plus npm and says in as many words that
"the npm package is a minority install path, so banding on the registry package alone would have
measured the wrong channel and given a materially lower level". That is exactly the shape #563 gave a
name to, and `n8n`'s npm artifact is a `not_primary_channel` candidate. Out of scope here — this PR
moves plumbing, not declarations — but it is now visible, which it was not before.

`hexabot` (npm widget) and `yomo` (Rust SDK crate) gain observations and **do not fall**: both are
declared `not_primary_channel` as of #563, so the kind names no shipping channel, route selection
never reaches it, and both stay on `github.stargazers_count` at level 2 — the same row, the same
level, before and after.

## Grain and double counting

- Grain is `(product_slug, artifact_kind, artifact_id, channel, metric_type, measurement_window_days)`.
  Live: 0 duplicate keys in the 1022 rows, 1022 distinct `observation_id`. The union's arms sum to
  465 + 325 + 187 + 45 = 1022, and the `rn = 1` filter drops nothing, so there were no duplicates
  before the dedup either, not merely none after it.
- `mastra`, the two-npm-package case, produces **two** rows — `@mastra/core` (5,363,932) and `mastra`
  (2,493,757) — one per declared artifact, neither lost and neither duplicated. The rollup then sums
  them to 7,857,689 under `sum_usage_across_artifacts`, which is the map's family unit.
- `beeai` is the case that would have caught a collapsed grain: it declares `beeai-framework` on
  **both** npm and PyPI. Two rows, distinguished by `artifact_kind`, not one row that lost a
  registry. Its band does not move — route precedence puts `pypi` ahead of `npm`, so it still bands
  on its PyPI row at level 2, and the npm row is published without being summed into it.
- The coverage guard never fired on any of 1022 rows and `product_type` is non-null throughout, which
  is the proof that every npm and crates artifact is declared: the successor's roster is
  `registry.product_artifacts` filtered to the three package kinds, so it cannot emit an undeclared one.
- `build.observation_snapshot`'s strict digest accepts the new row set (content digest `a282df37785b…`,
  snapshot id `ccaf3e83e9a0…`), so the identity path takes `npm`/`crates` as nonempty vocabulary
  values without a contract change.

## Downstream that could have narrowed

- `build/adoption_measurements.py` filters contributing observations on
  `artifact_kind == route["artifact_kind"]` and `metric_type`, never on `channel` or `source_dataset`.
  The npm and crates routes (`npm.downloads_30d`, `crates.downloads_30d`) therefore start matching
  rows they could not see before — confirmed by running the rollup, not assumed.
- Nothing else in `build/` keys on `channel == 'pypi'` or on `source_dataset`. `check_channel_authority`
  tests `route.get("artifact_kind") != "pypi"` but reads the routing tables and `sources/products/*.yaml`,
  not this model. `adoption_reconciliation` passes `channel` through from the measurement row.
  `observation_snapshot` requires `channel` to be a nonempty string and does not enumerate it.
- `build/check_artifacts.py` and `warehouse/models/signal_github/product_adoption.sql` still read
  `signal_pypi.package_downloads`; neither is in this change's path, and the dependency contract for
  that table stays.

## Risks / unsure about

- **`warehouse/assets.yaml` calls `signal_packages.downloads` `status: staged`, `materialized: false`,
  `refresh: 'not deployed; see issue #314'` — and it is deployed.** I queried it live: 188 rows,
  `last_day_seen` 2026-09-12, three registries. The #563 changelog entry already refers to "the two
  deployed `signal_packages` models". So my change is safe (it reads a table that exists), but the
  inventory label is stale. I did **not** fix it: `tests/test_platform_mirror.py::test_the_staged_package_models_are_all_present`
  asserts all three `signal_packages` models are `staged`, and flipping one to `active` also subjects
  it to the retirement/no-reviewed-consumer predicates that exempt staged assets. That is a decision
  with consequences, not a typo, and it wants its own change. Flagging it as the most likely reviewer
  objection to this PR.
- **`sources/signal_routing.yaml` now under-describes reality.** Its npm `bridge_note` says "no signal
  model reads the npm registry" and the route note says "this route resolves to no observation". Both
  become false with this merge. I left the file alone for the reason in Why, and recorded the lag in
  the model header, but a reviewer may reasonably want the prose corrected in this PR even if
  `bridged` stays false.
- **`n8n` will now reconcile as a disagreement** (measured 3, recorded 4) where before it produced no
  measurement at all. That is the bridge working, and it needs a curation follow-up
  (`not_primary_channel` on the npm artifact, or an explicit expected difference). It is a new red row
  on a board that did not have one.
- **The successor's join is case-insensitive on the package name** (`LOWER(w.package) = LOWER(r.package)`)
  while `fetch_status` joins exactly. If two differently-cased packages of the same kind ever land in
  the daily history, one roster row could fan out to two windowed rows and break this model's grain.
  It does not today (188 source rows for 188 declared artifacts), and it is the successor's bug to fix
  rather than this model's, but this change is what makes it reachable from the observation layer.
- The band effect is a measurement of the warehouse as of 2026-09-13. The npm figures move weekly; the
  route assignments and the agree/rise/fall split will not.

## Response to round N-1

|---|---|---|---|---|
| n/a — round 1, no prior review | | | | |

PR #567, opened 2026-09-13
Correct the cause of the 2026-09-13 release freeze

ccerv1:
## Summary

The runbook merged in #560 says our own `columns=` change on `signal_github.artifact_state` froze five models. **That is wrong**, and it is the version a maintainer would act on.

Twenty of the frozen models never read that table. A large share of the org could not be released that day, including `entities`, `metrics`, `events`, `state_of_os_ai` and `ai_demand_curve`, none of which is gap-map work. The two events happened within minutes of each other and the second was read as a consequence of the first for most of a day.

Telling a maintainer to expect a cascade from their own edit is the opposite of the right instinct, which is to run a byte-identical control and find out.

## What replaces it

- **A byte-identical control answers the question per model, never org-wide.** On the same day `currentai.metrics.daily` re-released cleanly while `currentai.scores.taxonomy` was refused.
- **Do not read the verdict off an unreleased draft.** For SQL the schema is derived and frozen at release, so a fresh revision can report non-deterministic with zero columns and still release fine, coming back with its full schema. Reading drafts produced a wrong count twice in one evening.
- The recreation paragraph drops a stale count and says what is actually true for Python: the schema comes from the in-code `columns=` declaration, so a model whose schema was passed as an argument cannot be re-released, and declaring it only takes effect on a newly created model.

## Test plan

- `uv run python -m build.validate` — 0 errors, 2 pre-existing warnings
- Docs only; no code path touched

https://claude.ai/code/session_016PavZokTp99gCG1VYjX9nN


ccerv1:
**Both blocking findings addressed.**

**1. Step 4 still carried the wrong story** while the section below it retracted the same story. That is this repo's eighth trap for the third time today: correcting one passage and leaving the other. It is retracted in place rather than deleted, since it is the claim a maintainer would have planned around.

**2. The control instruction was unsafe as written,** and review is right that "re-release the previous revision" could mean an older release or an unreleased draft, either of which would ship a different definition.

It now says the **currently live released revision**, and carries the part I learned the hard way the same evening: **a control can succeed, and then it is live.** That happened to `currentai.metrics.daily`. The code was identical so nothing changed in behaviour, but re-releasing the original afterwards does not undo it — the release row is updated in place and keeps its original timestamp, so it cannot supersede the newer one. The instruction now says to run a control only where shipping an identical definition is acceptable.

It also now requires the refusal to be **the determinism-verdict message specifically**. A release refused for another reason says nothing about this diagnosis, which review flagged and I had left implicit.

On the non-blocking note: agreed that the contrasting outcomes show each model must be tested independently and do not establish the platform's root cause. The document does not claim to — the root-cause material went to OSO as a separate reproduction with model and revision ids, not into this runbook.

`uv run python -m build.validate` — 0 errors, 2 pre-existing warnings. Docs only.

PR #568, opened 2026-09-13
Bridge the package routes onto signal_packages

ccerv1:
## Summary

Step 3 of #562. The `pypi` route pointed at `signal_pypi.package_downloads`; the `npm` and `crates` routes pointed at `table: null`. All three now read `currentai.signal_packages.downloads` with an `artifact_kind` filter, which is the table the observation layer already reads.

**Bridging npm and crates is the substantive half.** It turns on `check_instrument`'s artifact precondition for every product recording `usage_volume` on those registries, so a recorded download count there becomes falsifiable rather than merely re-checkable. That is the curation change the observations model header warned belonged to this step, and it lands clean.

```
check_instrument   664 adoption records          [OK]
check_adoption     664 products with a band      [OK]
check_routing      adoption 560 routed, 0 blocked
```

## The checker widens rather than swaps

`check_artifacts.pypi_missing` becomes `package_missing` and reads `missing_from_registry` on the successor, so it now covers **npm and crates as well as PyPI**. A reserved or deleted npm package was previously invisible to every check in that file. The `artifact_exceptions` enum in the product schema moves with the rename; no product declares that exemption today.

## One test rewritten rather than retargeted

`test_an_unbridged_source_does_not_satisfy_recomputation` asserted `"npm" not in artifacts["usage_volume"]`. That was true only while npm happened to be unbridged, so bridging it broke the test without touching the invariant it protects.

Swapping in whichever source is currently unmeasured would just move the problem. It now runs against a synthetic routing file with one bridged and one unbridged source, so it tests the rule and stops failing for good news. The history that produced it is kept in the docstring, since the reason it exists is worth not losing.

## Two statements that had become false

- The `npm` bridge_note said "no signal model reads the npm registry". It does now; 17 npm artifacts are measured.
- `check_channel_authority`'s docstring said npm and crates are unbridged. Still true of **its** leg, for a different reason now worth stating: `signal_packages.downloads` carries volume, not a published version, and that leg needs a registry version to compare against a source one.

## Not in scope

`signal_pypi` is untouched and still read by `signal_github/product_adoption.sql`. Dropping it is step 5, it is irreversible, and it should wait for OSO's answer on the release freeze.

## Test plan

- `uv run pytest -q` — 2075 passed, 1 skipped
- `uv run python -m build.validate` — 0 errors, 2 pre-existing warnings
- `check_instrument`, `check_adoption --strict`, `check_routing`, `check_retirement` all clean
- `current-state-dag.md` regenerated

https://claude.ai/code/session_016PavZokTp99gCG1VYjX9nN


ccerv1:
**Review found two blocking issues and corrected my framing. All three addressed.**

**1. `package_missing` overstated what `missing_from_registry` proves.** The column is `no history AND (kind is pypi OR a status came back)`. On npm and crates a 429 or a 5xx sets it, because a status came back and no history did — a **failed measurement**, reported as absence. Under `--strict` that fails CI for weather.

It now reports only a 404 on npm and crates. For pypi it says what that leg can actually support, "no downloads recorded on PyPI in the window", since the PyPI arm windows `oso.pypi_downloads` and never fetches, so a live package with no installs is indistinguishable from a deleted one. The predecessor's "absent from PyPI" could not support that claim either.

Live effect: one finding, `monocle_apptrace`, now reported with the honest reason.

**2. The npm and crates ROUTE notes still said UNBRIDGED.** I corrected the `sources:` notes and missed the route notes three hundred lines below, which contradicted the declarations directly. That is this repo's eighth trap almost verbatim — closing a hole in one place and leaving it open in the other. Both corrected, plus the file header's version-2 note.

**3. My framing was wrong, and the PR description above is corrected.** I said bridging turns on a mandatory artifact precondition and tightens the gate. It does not. `check_instrument` accepts a readable artifact **OR** dated, digested evidence, so widening the first leg can only let more records pass. All thirteen newly recomputable products already satisfied the evidence leg, which is why no finding moved. The gates passing unchanged is the expected result, not evidence the change did what I claimed.

**One thing review did not raise that it led me to.** The branch naming what a product declares hardcoded `("npm", "crates", "docker")` in Python — the mirrored-in-Python drift the sibling test explicitly forbids, and stale the moment npm and crates became countable, since both can now reach the recomputable path instead. Derived from the routing file now.

Four more present-tense statements corrected: the schema's `pypi` description, the artifacts workflow header, `check_instrument`'s own docstring, and the observations model's reader list.

`uv run pytest -q` — 2075 passed, 1 skipped. `check_instrument`, `check_adoption --strict`, `check_artifacts` all clean.

PR #596, opened 2026-09-15
Create the ml_orchestration category and move ray into it (#429)

### LINKED ISSUE #429: [category-proposal] ML orchestration & pipelines — and is Ray misfiled in deployment?
## The question this has to answer first

**If Ray, Flyte and ZenML are one kind of thing, who are their actual peers?**

Three products is not a category. This issue should not be closed by creating one until there is a peer set deep enough to support a single coherent capability dimension — and if that set cannot be assembled, the honest outcome is that these three stay where they are or leave the map, not that a thin category gets created to hold them.

## Why it was opened

The `deployment` category is "Sandboxes, runtimes, and serverless model hosting", and its capability ladder is in practice **a single ordering on isolation strength**: every one of its 14 calibrated products is a sandbox, a VMM or a sandbox service, banded on isolation tier, cold start, per-instance overhead and fleet scale.

A category pass over 13 candidates found six that cannot be banded on that axis at all — they would *use* Firecracker, not compete with it. Four were general-purpose infrastructure and are excluded in #428. Two are genuinely AI stack products with nowhere to go:

- **`flyteorg/flyte`** — orchestrates ML pipelines, models and agents in pure Python, tasks declared against a `TaskEnvironment` and executed on a Kubernetes-backed control plane
- **`zenml-io/zenml`** — pipelines over pluggable infrastructure stacks, containerizing and tracking code and delegating execution to SageMaker, Vertex or Kubernetes

## `ray` is the migration candidate, not the precedent

`ray` already sits in `deployment` at capability 4, and its own capability note concedes the problem: it is *"a distributed compute, training, tuning, data and serving framework rather than an isolation boundary."*

So `ray` should be read as **the anomaly in `deployment`, not the precedent that admits Flyte and ZenML**. Either it belongs with them in a new category, or it is itself misfiled and that should be settled once rather than per-product.

## Why not just add a second ladder to `deployment`

The governing precedent is the 2026-08-18 move of Qdrant, Milvus and Pinecone out of `agent_tools_protocols` into `storage`. That category's own comments record the reason: *two capability recipes banding the same kind of product meant their bands could not be compared in the published map.* An orchestration rung set inside `deployment` recreates exactly that.

The forcing is also measurable. Banded against `firecracker`, five of the six orchestrators land on rung 3 — the container rung — because shelling out to a container is the only isolation fact they expose. A rung that admits a third of a category has stopped discriminating, which `capability.md` says directly.

## Provisional scope, if the depth test passes

Systems that coordinate distributed ML jobs, training pipelines, workflows, tuning and execution across infrastructure. Explicitly **not** `ml_frameworks`, whose definition is foundational libraries imported by the rest of the stack.

The proposal needs, before anything is created:

1. a peer set — Kubeflow-like systems, Metaflow, Prefect/Dagster/Airflow if they qualify as AI stack rather than general workflow, Argo Workflows, SageMaker Pipelines and Vertex Pipelines as the hosted end
2. **one** capability dimension the whole set can be ordered on, with an anchor that carries an evidenced `value` (see #427)
3. its Columbia arc/layer assignment, since `serialize` emits a per-category `layer`
4. a check on whether the general workflow engines excluded in #428 would have to be re-admitted under the new definition — if they would, the definition is too broad

## Meanwhile

`flyte` and `zenml` are recorded `unresolved` in the ledger. `deployment` is not broadened while this is open.

ccerv1:
Closes #429.

25 head products across 24 organizations, created and published in one commit. Not staged through
`preliminary`: that status omits a category from the notebook payload, so moving `ray` into one
would have dropped a published product off the map and `check_retirement` would have demanded an
alias for a product nobody retired.

## Review sheet

| category | stage | gaps | products | tier changes |
|---|---|---|---|---|
| deployment | 5 -> 4 | [] -> ['resiliency'] | +0 / -1 | none |
| ml_orchestration | None -> 4 | [] -> ['resiliency'] | +25 / -0 | none |

`untouched-product row changes: none`.

**`deployment` drops a stage, and it is arithmetic rather than a re-scoring.** `ray` was one of its
four mature fully-open products (openness 5, adoption 5, capability 4, `B` = 4.6 against the 4.5
threshold). Stage 5 needs four; three remain. No score in `deployment` was changed to produce this.
Carries `stage-move`.

## The membership test

Two limbs on the product's **own job model**. Either (a) it declares ML work as a first-class type
it schedules and tracks — a training run, a tuning trial, a model or checkpoint artifact — or (b)
its own scheduler consumes an accelerator request in its own schema and uses it to select or
provision the machine. Passing a backend's pod spec through does not satisfy (b), and neither does
a resource field no shipped scheduler reads.

One rule excludes all eight neighbours, each on a named fact rather than on positioning:

- **`airflow`** defines `Resources(cpus, ram, disk, gpus)` in its Task SDK, but a repository-wide
  search finds it referenced only by the operator definition, serialization and tests. Recorded at
  that strength — it establishes no first-party consumer, not a universal negative across provider
  packages.
- **`prefect`, `dagster`, `argo-workflows`** pass a Kubernetes or Docker spec through.
- **`kestra`, `hatchet`, `dagu`, `atmos`** are #428's four, unchanged.
- **`mlflow`** models runs and artifacts as first-class types but **schedules nothing**, which is
  why "schedules" is in limb (a) and why it stays in `telemetry_observability`.

`kubeflow-pipelines` is the admit that proves the test does work: it runs on Argo Workflows, which
is excluded, and the difference is one file — the KFP SDK defines `system.Model`, `system.Dataset`
and `system.Metrics` as typed artifacts, and Argo has no equivalent.

## Capability: one quantity, three anchored bands

How far the product's own scheduler gets toward putting work on a machine. Band 1 places nothing
and hands the pipeline over; band 3 decides when each unit runs and which pool it goes to; band 5
chooses the machines and reserves a multi-machine set as one allocation. Bands 2 and 4 are for
products between anchors — 4 holds two, 2 is empty.

Substrate breadth is deliberately **not** in the definition: whether a product can place on two
clouds or one is a fact about what the vendor sells. Neither is commercial tier — this category
scores wholly closed products, so conditioning capability on openness would put two questions on
one axis, which is why `clearml`'s enterprise-only GPU allocation and `polyaxon`'s commercial
queueing both count.

Distribution: 1 / 12 / 4 / 8 across bands 1, 3, 4, 5. Widest band 48%. `capability.md` names a
third as its diagnostic; measured across the live corpus, **18 of 20 published categories exceed
it** (median ~48%, `safeguards` 81%). The number is reported rather than used to trim the roster.

## Openness

15 fully open, 3 open-core, 1 source-available, 6 closed. Three calls worth reading:

- **`zenml` 4/open_core** on packaging rather than on a feature table: ZenML Pro ships as separate
  container images *"privately hosted in the ZenML GCP Artifact Registry and ZenML AWS ECR
  repositories. Access to these registries is granted upon request"*, pulled by a self-hosted
  deployment with an image-pull secret. A closed component of the deployment you run.
- **`clearml` 2/source_available**: the SDK is Apache-2.0 but `clearml-server`, the thing you run,
  is SSPL-1.0.
- **`openpai` 5** with its MIT body read behind a Microsoft copyright line, and **`cube-studio`
  held** for the opposite reason — a genuine MIT body, but two repositories carry the name and the
  larger one has a 25-entry tree, so identity is unsettled.

## `ray`

Rebanded 4 -> 5 and written the capability `value` it lacked, retiring a weak comparison root under
#427. The band is not carried across: `deployment` orders on isolation strength, a ladder ray's own
note conceded it did not belong on.

`modal-sandboxes` recorded `at ray` and is repointed to `one_below firecracker` — the comparison its
own note already argued, and a root that records a value, which `ollama` does not. Its score is 4
before and after. `check_capability` caught this; the first draft of `deployment`'s comments
asserted ray anchored nothing, and that was wrong.

## One thing that needs a maintainer ruling, and does not block

`slurm` records **GPL-2.0**, which the shared `osi` tier does not name, so the lookup abstains and
the 5/open_source pair is hand-placed with the reason in `scoring_recipe.deferred`. GPL-2.0 is
OSI-approved and inside the tier's own definition — the same shape as the BSD-2-Clause and
PostgreSQL entries already there. Measured before raising it: **no other software-type product on
the map records GPL-2.0 under `license`**, so naming it in the tier would move this record and no
other. Extending a shared tier reaches eleven categories, so it is not part of a promotion.

## Adoption reads low here by construction

Seven products — `volcano`, `kueue`, `kai-scheduler`, `kubeflow-trainer`, `openpai`, `armada` and
`slurm` — ship as container images, Helm charts or distribution tarballs and publish nothing on any
route `signal_routing.yaml` carries. They band on `stars_fallback`, capped at 3. Slurm runs a large
share of the world's GPU training fleets and bands at 2. That is a missing instrument, not a
measurement, and it is the #446 family of gap.

`kubeflow-trainer` is deliberately **not** banded on the `kubeflow` PyPI package's 100,711 monthly
downloads: that is the umbrella Kubeflow SDK, a different population.

## Test plan

- `uv run python -m build.preflight` — **all 16 locally-runnable CI steps pass**; the four it cannot
  run locally are accounted for.
- `uv run pytest -q -n auto -m "not serial"` — **2093 passed, 1 skipped**, exit 0.
- `uv run pytest -q -m serial` — **1 passed**, exit 0.

Three pinned censuses moved and each carries its reason beside it rather than a bare new number:
the parity deferral count 4 -> 5 (`slurm`'s tier gap); the channel-authority prose leg 17 -> 18
(`slurm` on stars, with nothing to remedy it to); and the adoption measurements digest — where the
row sets were dumped on both sides and diffed rather than assumed: **378 rows before and after,
exactly one row different, and the only field that differs on it is `category_slug`**. The 24 new
products contribute no measurement rows, because measurements come from the frozen observation
snapshot rather than from declared artifacts.

https://claude.ai/code/session_017YpDugrn1ApNRnMK1yjuYf


ccerv1:
## Review round 1 — eight blocking findings, all addressed

An external review of the diff (not of the design, which was settled over four earlier rounds) returned `fail` on eight points. Every one landed; the fixes are in `dbdd7185`.

**A recursive tree establishes filenames.** Fourteen records rested a `core-gated` value on one while claiming the absence of license-key machinery, a closed dependency or a vendor edition — and `slurm`'s tree citation attributed SchedMD's commercial model to a file listing. Every detail now claims path absence only, and every tree citation ends *"A tree lists paths; it is cited for that and for nothing about the vendor's offerings."* Five `commercial-tier` values naming offerings no source established are gone.

**`sagemaker-pipelines` drops 4 → 3.** Its own note assigned the multi-instance allocation to SageMaker Training, so band 4 credited the pipeline product with its downstream service's placement. `nvidia-run-ai` becomes a root at 4 on its own evidence; `pai-dlc` sits `at` it.

**`slurm` and `openpai` keep band 5** on citations that reach it — Slurm's quickstart on allocating *"access to resources (compute nodes)"* and running a parallel job *"on the set of allocated nodes"*, and HiveD's user manual on topology cells above the leaf level.

**Band 3's definition narrowed to its evidence.** It claimed the scheduler picks the pool as well as the timing; six citations didn't show that. Products that do pick a queue are noted as exceeding the minimum. No score moved.

**`zenml`'s value claimed it owns none of its orchestrators** — false, the local and Docker flavors are its own. Band unchanged, and now rests on what those flavors do rather than on a miscount.

**Unsupported market-share claims removed** from `slurm`'s description and adoption note, the strapline and the category comments. An absent channel shows a gap, not its size.

### Two corrections to the PR description above

**`slurm` no longer defers, and no longer needs a ruling at merge time.** GPL-2.0 was added to the shared `osi` tier on the owner's ruling, with the measurement recorded beside it in `software.yaml`: no other software-type product on the map records that license, so the name moves this record alone. `check_rubric` now reports **ml_orchestration: 25/25 reproduced** — the score is computed rather than authored, and the category's only deferral closed. The parity deferral census therefore stays at 4 rather than rising to 5.

**The capability distribution in the description is wrong twice over.** It should read **1 at band 1, 13 at band 3, 2 at band 4, 9 at band 5 — widest band 52%**. The first figures were transposed; the second change came from the `sagemaker-pipelines` reband. Generated from the files rather than typed this time.

One census moved in the opposite direction to the first round: removing `slurm`'s unsupported "understates it badly" line stopped the under-coverage detector firing on it, so the channel-authority prose leg returns to **17**. That is the right outcome — the set holds notes that *assert* under-coverage, and `slurm` no longer asserts it. Seven products in this category still band on stars for want of an instrument; none of them now claims to know what the missing measurement would say.

### Re-verified after the fixes

- `build.preflight` — all 16 locally-runnable CI steps pass
- `pytest -q -n 4 -m "not serial"` — **2093 passed, 1 skipped**, exit 0
- `pytest -q -m serial` — **1 passed**, exit 0

ccerv1:
## Review round 2 — four blocking findings, and two of them were mine to cause

Round 2 is answered in `50bd1f0`, and it is worth naming that **two of the four were defects introduced by round 1's own fixes** — the failure mode this repo has hit before, where the repair opens the next hole.

**Narrowing band 3 collapsed band 1.** Removing the "chooses the pool" clause left band 3 reading "decides when work runs and dispatches it", which is exactly what ZenML's *own* local and Docker orchestrators do. `zenml` moves **1 → 3**, and band 1 is now defined, empty, and documented as empty: even a framework built to delegate ships an executor. Keeping it occupied would have needed a local-versus-remote distinction the ladder does not make.

**The rewritten tree citations contradicted their own tags** — text disclaiming that a listing shows anything about a vendor, while still marked `establishes: core-gated`. Now trees establish `source` only, and `core-gated` is carried by evidence that can bear it: CNCF project pages for Volcano, Kubeflow, Armada and HAMi; Kueue's *"You can install Kueue on top of a vanilla Kubernetes cluster"*; dstack's own guide naming Self-hosted dstack, dstack Factory and dstack Sky separately; SkyPilot's navigation; SchedMD's support page. The code-search claims left the notes, since a search result is not re-fetchable.

**`openpai` 5 → 4** — a topology representation is not the allocation operation band 5 asks for. **`pai-dlc` 4 → 3** — the same ownership inference removed from `sagemaker-pipelines` one record over. And `slurm`'s openness note still described a tier that does not name GPL-2.0, a hand-assigned pair and a deferral, all three false after the ruling.

## Three products added

| Product | Band | What puts it there |
|---|---|---|
| **HAMi** (4,600★, CNCF) | 4 | its own protocol doc: *"HAMi scheduler needs to patch schedule decisions into pod annotations"*, `hami.io/vgpu-node: {schedule decision node}` |
| **HTCondor** (328★) | 4 | `request_GPUs`, `require_gpus`, `gpus_minimum_capability` in its own submit language, *"matching jobs that need GPUs to machines that have them"* |
| **DLRover** (1,687★, Ant Group) | 3 | elastic training runs on Kubernetes with failure detection and resume |

Each new band carries a `comparison` block attested against a **same-day re-read of its peer**, rather than claiming a date its peer predates.

Two measurement calls worth flagging: HTCondor bands on stars rather than on the `htcondor` PyPI package, because that package is the Python bindings and would measure API users rather than clusters; DLRover bands on its PyPI package at level 1 (295/month) rather than reaching for the higher star count, because the package is genuinely its own channel.

DLRover is the third product in this PR whose licence reports `NOASSERTION` over a standard open body behind a preamble, after `slurm` and `cube-studio`. Three catches in one change suggests a gate rather than three readings.

## IBM blocks the fetcher, like Huawei

`ibm.com` returns **403** to `build.fetch_source`, so IBM Spectrum LSF is held for the same reason as Huawei ModelArts. Both are recorded in the category's `comments` with the consequence stated: two large enterprise vendors blocking the fetcher means the closed side here **skews toward vendors whose documentation is openly readable**. That is a property of this map's instrument, not of their products.

## Current shape

**28 products.** Capability **16 at band 3, 4 at band 4, 8 at band 5** — widest 57%, generated from the files rather than typed.

| category | stage | gaps | products |
|---|---|---|---|
| deployment | 5 → 4 | [] → ['resiliency'] | +0 / −1 |
| ml_orchestration | None → 4 | [] → ['resiliency'] | +28 / −0 |

`untouched-product row changes: none`.

Re-verified: `build.preflight` all 16 steps pass; `pytest -n 2 -m "not serial"` **2093 passed, 1 skipped**.

Still held, each with an unblock condition: `modelarts-training` and `lsf` (403), `modal` (needs an identity ruling against `modal-sandboxes`), `cube-studio` (two repositories carry the name). `yunikorn` and Nebius serverless jobs are unevaluated candidates carried forward.

ccerv1:
## Review round 3 — two under-bandings, and the category's headline claim was wrong

Shipped in `8803c538`.

**`nvidia-run-ai` 4 → 5.** I banded it from the product marketing page, which attributes gang scheduling to Grove with KAI. Its **own Scheduler documentation** says the Scheduler *"always uses the pod group associated with each pod and handles the pod as part of a larger group of pods… such as gang scheduling, minimum replicas or workers… applied consistently across the entire pod group"*, over *"distributed workloads using multiple pods, each running on a node."* That is a multi-machine set reserved as one allocation by the platform's own scheduler.

**This falsifies what the category was claiming.** The strapline said no closed product reaches the top band. It now says what is true: nine of the ten products that reserve their own multi-machine allocation are open source or open core, and the tenth is NVIDIA Run:ai.

**`htcondor` 4 → 5** on `machine_count`, *"the number of machines to be dedicated toward running the job"* in the parallel universe. Same error as Run:ai — I read the single-job GPU page and took its silence for a ceiling. `openpai` anchors band 4 now that Run:ai has left it; `hami` is attested against `openpai`.

**`htcondor`'s adoption moves to its own package** — 28,301 monthly downloads, level 2. Excluding it because "the bindings measure a subset of cluster users" is the minority-share argument this project has already settled against; the package covers this product and no other, so the measurement belongs to it.

### The `ungated` standard is now written down

Three review rounds argued whether tree path-absence, foundation governance or a vendor's product page establishes `core_gated: ungated`. That argument is now settled in `sources/rubrics/software.yaml`, which owns the dimension:

> "`ungated` is a negative — nothing is withheld — and no document ever asserts it directly. A reviewer can therefore always say that the cited evidence establishes something weaker, and be literally right every time. That is unfalsifiable, so the ladder states what it accepts instead."

Path absence, **never sufficient alone**, plus a statement from whoever would do the withholding — a vendor naming its offering as a separate product, or foundation governance meaning no commercial party exists to withhold anything. Against that sits what makes a product `gated`, which is directly provable: a closed package the open one pulls, a license key in the published source, an `ee/` tree. Gating is provable and openness is not, which is why the burden sits where it does.

This is the standard `ray`'s 5/open_source has always rested on. Raising it would unsettle every foundation project on the map at once — a corpus migration with its own ruling, not something a category promotion may do on the side.

### Nothing was lost

Checked against `main` rather than asserted:

```
main placed: 686    branch placed: 713
products that lost their place: none
products that changed category:  {'ray': ('deployment', 'ml_orchestration')}
product files deleted: none
```

### Final shape

**28 products** — 18 fully open, 3 open-core, 1 source-available, 6 closed. Capability **16 at band 3, 2 at band 4, 10 at band 5**, widest 57%, generated from the files.

| category | stage | gaps | products |
|---|---|---|---|
| deployment | 5 → 4 | [] → ['resiliency'] | +0 / −1 |
| ml_orchestration | None → 4 | [] → ['resiliency'] | +28 / −0 |

`untouched-product row changes: none`.

Verified: `build.preflight` all 16 steps; `pytest -n 2 -m "not serial"` **2093 passed, 1 skipped**; `pytest -m serial` **1 passed**.

Held with unblock conditions, in the category's `comments`: `modelarts-training` and `lsf` (Huawei and IBM both return 403 to the repository's own fetcher), `modal` (needs an identity ruling against `modal-sandboxes`), `cube-studio` (two repositories share the name), `yunikorn` and Nebius serverless jobs (unevaluated).

Issue #605, opened 2026-09-16
[hygiene] Category leaks: garak (eval in safeguards), and document-processing tools in agent_tools_protocols


ccerv1:
## What
Two placements the omissions review flagged, verified against the current corpus:

1. **`garak` is in `safeguards`** but it is red-team / vulnerability **evaluation** tooling. It reads more like an `evaluation_code` product (its peers there are `pyrit`-adjacent scanners). Decide whether it moves, or whether `safeguards` deliberately holds red-team tools (in which case say so in the category comments).

2. **Document-processing tools sit in `agent_tools_protocols`** — `marker`, `mineru`, `docling`, `markitdown` — but their job is document → text extraction, which is `dataset_processing_tools`' concern (`unstructured` is already a candidate there, #598). Decide: move, or dual-home with a stated reason.

## Why it matters
Both are "category leaks": a product banded on one category's capability ladder while conceptually belonging to another can't be compared against its real peers in the published map — the same failure the 2026-08-18 Qdrant/Milvus move into `storage` fixed.

## Not asking for
No roster change in this issue — just a decision (`update-product` moves, or a category-comment recording why they stay). `garak` and the four processors are already head products, so any move is an `update-product` category-membership change, not a new product.

_Filed from the 2026-09-16 AI Stack Map omissions review triage._

ccerv1:
The converter half is resolved by #610: `docling`, `marker`, `mineru` and `markitdown` left `agent_tools_protocols` for the new `document_conversion` category rather than for `dataset_processing_tools` as proposed here. The category comment records why - `dataset_processing_tools` bands on two instruments already and its recipe carries no written rungs, so moving four more products in on a third would relocate the defect instead of fixing it.

The garak half is #608, still open.

ccerv1:
Half of this is already done; recording which half so it is not re-derived.

**Part 2 — document-processing tools — RESOLVED.** `marker`, `mineru`, `docling` and `markitdown` are no longer in `agent_tools_protocols`. They sit in `document_conversion`, the category created by the #430 split (#610), which is a better answer than either option this issue offered.

**Part 1 — garak — still open, and it is now two products.** `garak` and `pyrit` are both still in `safeguards.yaml`. Carl's ruling: **move both to `evaluation_code`.** They are red-team and vulnerability evaluation tooling, and the category holding them should be the one that says so.

[...]

Issue #430, opened 2026-08-31
[category-proposal] Split agent_tools_protocols — three hypotheses, decided by capability coherence


ccerv1:
`agent_tools_protocols` is "Tools, protocols, and retrieval for agents", which reads coherent at the label and hides **several incomparable capability questions**.

## The evidence that it is not one category

Its capability anchor is `docling` at 5, whose `value` is document-conversion breadth: formats in, formats out, layout, tables, formulas, OCR, ASR.

A category pass asked six new products to be banded against it. **None of them converts documents.** Five are MCP servers, a package manager or a wire protocol. The reviewing agent's own account of what it did:

> In practice I banded every product in this category against `playwright-mcp`, `composio` or `serper` and then converted to a docling relation for the arithmetic. That conversion is honest bookkeeping but it is not evidence.

The shared instrument across the existing calibration is not the anchor's at all — it is *how much of a target surface can this reach*, which is how `firecrawl`, `playwright-mcp`, `composio` and `apify` all reached 4.

Two products show the seam plainly: **`apm`** is an agent-config manifest format with a lockfile, and **`agent-client-protocol`** is a wire protocol published as JSON Schema plus SDKs. Neither has a comparable dimension with a PDF parser.

## Sub-anchors will not fix this

Sub-anchors work when products share one underlying capability dimension and occupy different portions of it. They do not help when the capabilities answer **different questions**. The governing precedent is the 2026-08-18 move of Qdrant, Milvus and Pinecone to `storage`: two capability recipes for one product kind made published bands incomparable. Here it is worse — Docling and an agent wire protocol are not the same product kind at all.

## Three hypotheses, not proposed final categories

1. **Agent protocols & interoperability** — MCP, A2A, ACP, and SDKs/inspectors where they are genuinely protocol infrastructure
2. **Agent tools & connectors** — MCP servers, Composio-like tool integration platforms, browser and tool execution surfaces
3. **Retrieval & document context** — search and retrieval APIs, document ingestion and conversion — **test for a further split**

## Bucket 3 must not be decided from naming

Run the capability-coherence test: **can `firecrawl`, `serper`, `docling`, `marker` and `jina-reader` be meaningfully ordered on one capability ladder?**

[...]

PR #609, opened 2026-09-16
Promote document_conversion with twenty products, and reverse a #413 ledger exclusion to do it (#430, #605)

ccerv1:
Stacked on #607. The second half of the #430 split: the document-converter group leaves
`agent_tools_protocols` and becomes a published category of twenty.

## The ledger reversal, first, because it gates the roster

Six repositories were recorded in `sources/resolution_ledger.yaml` as `excluded_boundary`
from a #413 sweep on 2026-08-30. `excluded_boundary` means outside the whole map, and the
sweep's own note cited an in-map analogue in `dataset_processing_tools` - so the decision
it was recording was a placement call inside `agent_tools_protocols`, encoded a category
wider than it was made. Codex was asked whether reinterpreting the verdict was enough and
ruled that it is not: the vocabulary is explicit, so the correction has to be a formal
reversal with a new `decided_in` and a new date, and the original verdict, boundary,
issue reference and date preserved in the note. That is what the six entries now carry.
`deepseek-ocr` and `paddleocr` resolve to `existing_product`; `dolphin`, `grobid`,
`layout-parser` and `tesseract.js` go to `unresolved`, which states what is still open
rather than admitting them by default.

## The category

Twenty products: `docling`, `paddleocr`, `chandra`, `marker`, `mineru`, `markitdown`,
`nanonets-ocr`, `dots-ocr`, `olmocr`, `surya`, `tesseract`, `deepseek-ocr`, `lightonocr`,
`google-document-ai`, `azure-document-intelligence`, `mistral-ocr`, `textract`,
`adobe-pdf-extract`, `mathpix`, `reducto`. Five publish source outright, four publish
weights, four are source-available under revenue-capped weights licences, seven are
closed. That is the open/open-ish/closed mix the bar asks for, at the 20+ figure that
every published category already meets.

Four products move in from `agent_tools_protocols` and one from
`dataset_processing_tools`. `#605` proposed sending the converters to
`dataset_processing_tools` instead; the category comment records why not - that
category's bands already span two instruments and its recipe carries no written rungs,
so the move would relocate the defect rather than fix it.

**Capability is one nested quantity: how much of a document survives the conversion.**
Text at 3, structure at 4, meaning - typed entities the consumer defines - at 5. Two
things are excluded by name because earlier drafts banded on them and both are a second
quantity: input breadth (how many document kinds, not how much of one) and pipeline
position (where the output goes next).

**The distribution is bad and is reported rather than engineered away:** 1 at band 3, 15
at 4, 2 at 5, 2 abstaining. Almost everything here recovers structure into Markdown or
JSON and the two ends are narrow. Manufacturing rungs to spread the count would encode a
distinction the evidence does not carry. A shared benchmark would fix it and there is not
one - olmOCR-Bench is AI2's and olmOCR wins it, OmniDocBench is OpenDataLab's and their
MinerU wins that one - so benchmark scores are cited per record as corroboration against
whatever currently leads each board, with the home-field problem stated.

**Two openness ladders**, `{model: model, software: software}`, the per-type form
`safeguards` already uses. Four products are typed `model` as a stated judgement per
record. An earlier draft claimed the line was mechanical - software ships an installable
package - and that premise was false for the one product it was written to decide, so it
is withdrawn rather than repaired.

Three deferrals, all license-tier and all naming what would settle them: `chandra` and
`surya` on Datalab's modified OpenRAIL-M, `nanonets-ocr` on a card that declares no
licence at all.

## Review

Three Codex rounds, 5 + 7 + 1 blocking findings, every one verified against the files
before fixing. The third round confirmed the adoption arithmetic, the class counts, the
capability distribution and every recorded comparison edge independently. What the rounds
caught, in case the pattern is useful: a declared Hugging Face URL that 404s and an
adoption band resting on stars while the real record published downloads; a `code`
component reading `closed` while its own detail said inference code is published; six
capability notes still arguing a ladder the category had already replaced; a band-3
assignment that asserted structure does not survive when nothing cited tested it (both
such records now abstain); a note claiming the record defers while publishing a score;
and a note claiming the openness walk returns a score it does not - replaying
`score_openness` over `surya` returns nothing at all, so the 2 is hand-assigned and now
says so.

A feature matrix is recorded in the recipe note as the future direction, with the three
conditions it has to meet first, rather than shipped as a token count that remixes the
quantities the ladder just removed.

## Test plan

- `uv run pytest -q` - 2094 passed, 1 skipped, exit 0, at the head commit
- `uv run python -m build.check_refetch --product <slug> --all --strict` run per product
  that gained a source

Closes part of #430. Resolves the converter half of #605.


ccerv1:
Stage impact, measured by serializing at `main` and at the head commit and diffing every category's stage:

```
document_conversion: none -> 4 (Competitive Open Ecosystem)
```

That is the only change. `agent_tools_protocols` loses four products and `dataset_processing_tools` one, and neither stage moves - none of the five was a fully-open product at 4.5 or above in the category it left, so the count that defines the stage is untouched on both sides. The new category enters at 4 on `docling`, `markitdown`, `paddleocr` and `olmocr`.

Labelled `stage-move` because a category appearing on the map at stage 4 is a deliberate change to the published gap set, not a side effect.

ccerv1:
Correction to the stage comment above - the measured diff was right, the reason I gave for it was not.

Counting mature open products (open bucket, maturity >= 4.5) on both sides:

```
agent_tools_protocols     8 -> 6   stage 5 holds (needs 4)
                                   loses docling and markitdown, both of which ARE mature open
dataset_processing_tools  0 -> 0   stage 2 holds; olmocr was never in the count
document_conversion       - -> 2   enters at stage 4 on docling and markitdown
```

So it is not that none of the products leaving was mature and open. Two of them were, and `agent_tools_protocols` keeps stage 5 only because the threshold is four and it still has six. `paddleocr` and `olmocr`, which I also named as anchoring the new category's stage, do not reach 4.5 maturity and contribute nothing to the count.

ccerv1:
Closed by GitHub when its base branch was merged and deleted. Continued as #610, same content rebuilt on `main`. The four review rounds and the stage-impact measurement are here.

PR #611, opened 2026-09-16
Withdraw the third-of-a-category rung rule from capability.md

ccerv1:
`capability.md` says a capability rung holding more than a third of its category is
unacceptable and is "not discriminating between anything; it is a label". Measured across
the corpus, nearly nothing meets it.

| | |
|---|---|
| Published categories with at least one banded product | 21 |
| Widest rung above a third | **19** |
| Median widest rung | 50% |
| Range | `base_pretrained` 28% to `document_conversion` 83% |

With five bands, an evenly split three-product category breaches it. A rule that nearly
every category fails is not a standard anyone is held to - it is a line that gets quoted
at whichever category happens to be under review, which is what prompted this: it was
raised against `document_conversion` in review while 18 other categories sat in the same
state unremarked.

## What is kept

The count itself, and the `storage` example behind it, which is the genuinely useful part:
a top rung written as "a distributed retrieval platform that also ranks or runs inference
in the serving path" admitted seven of twenty-seven products; reworded to "hosts and
evaluates ranking or embedding models inside the serving path" it admitted two, with
nothing about the products changing.

## What replaces the threshold

A reading rather than a number. When one rung is wide, ask which of two cases it is - the
definition generalized until almost everything satisfies it, or the products genuinely
cluster - and report the answer in the category's `scoring_recipe.note`. The checklist item
changes from "no rung quietly holds a third of the category" to requiring that a wide rung
is either reworded or reported with which case it is.

`sources/categories/scientific_ai_models.yaml` had already made this measurement against
the old eighteen-category corpus and argued from it. Its figures are refreshed and it now
points at the withdrawal instead of the rule.

Recommended by Codex during the `document_conversion` review (#610), where the same rule
would have required manufacturing rungs to spread a count the evidence does not support.

## Test plan

- `uv run pytest -q` - 2094 passed, 1 skipped, exit 0


ccerv1:
Codex round 1: two blocking findings, both true, both fixed in `39e90c0d`.

**The rationale argued from the wrong thing.** "19 of 21 categories fail it" is as much an argument for fixing the corpus as for dropping the rule. The withdrawal now rests on what a distribution can decide: a wide rung has two causes - a definition loose enough that nearly everything satisfies it, and a category whose products genuinely share a capability - and a count cannot separate them. Splitting the second records a distinction no evidence supports. The corpus measurement stays, as scale and as the explanation for why the line was being quoted selectively, not as the reason it goes.

Codex also caught that **the threshold was never calibrated against its own founding example**: the `storage` rung this section condemns admitted 7 of 27 products, or 26% - below the third the rule went on to forbid. What made that rung wrong was read from the definition against the products, not from a percentage. That is now in the document.

One arithmetic error of mine went with it: I wrote that an evenly split three-product category breaches the rule. It does not - one third each is exactly a third, not above it. Removed.

**The replacement was unfalsifiable.** As drafted, a curator could satisfy it by declaring the products cluster. Above a third is kept as a trigger, as a number so it fires the same way for everyone, and what it demands is a record in the recipe note: the count and its denominator, the rung's wording, which case it is, and the evidence - the reword, re-counted afterwards, or what the products actually share. The checklist item matches.

Two non-blocking corrections applied in `scientific_ai_models`: three categories occupy all five rungs, not four, and its fat middle is where the observed attribute combinations fall rather than something arithmetic - 20 products record two or three of the four attributes, against 14 recording none, one or all four.

Preflight 16/16, suite 2094 passed, 1 skipped, exit 0.

PR #614, opened 2026-09-17
Promote search_retrieval with twenty-one products, and close the #430 split (#430)

### LINKED ISSUE #430: [category-proposal] Split agent_tools_protocols — three hypotheses, decided by capability coherence

ccerv1:
The last piece of #430. Eleven products move in from `agent_tools_protocols`, ten promote
from the seed registry, and the category publishes at 21.

## The ladder

Capability measures **how far along one workflow the product's own documented interface
takes the caller**. The workflow is stated in the recipe because the order has no meaning
without it: a caller holds a question and an HTTP client and wants text to put in front of
a model, and the steps are decide where to look, find candidate sources, retrieve and clean
them, compose an answer.

| band | what the interface returns |
|---|---|
| 1 | the content of an address you name |
| 3 | ranked references for a query |
| 4 | references **and** the content behind them |
| 5 | a composed answer over retrieved sources, with citations |

**The bands do not nest, and the recipe says so.** The first draft claimed they did and it
is false at the first step - SearXNG returns references and will not fetch a URL you name,
Crawl4AI fetches anything you name and finds nothing. This is an ordinal scale over
interface function, resting on one stated assumption: given a candidate URL, retrieval is
the step a caller is most likely to already have and discovery the step it is least likely
to have. That is about availability, not difficulty, and it does not claim a higher band
means less work.

Band 2 is unused and must not be assigned.

## The scoring unit, and where it does the work

Three parts: documented endpoints count **in combination**, a product is **what the vendor
sells under that name**, and the **highest step reached** is the band.

The second part carries this category. Brave, Kagi, Parallel, Exa and You.com all sell an
answer product, and all five stay at band 4 because that product is a separate plan or a
separate line on their own price list. Linkup reaches 5 because `sourcedAnswer` is a
parameter of the one `/search` route it sells, at $0.006 against $0.005. Move the test and
the distribution moves with it, which is an argument for stating it precisely rather than
softening it.

Distribution 1/4/11/5 over 21, widest rung 52% against a corpus median of 50%, reported in
the recipe note with which of #611's two cases it is (clustering, not a loose definition).

## Boundary calls, both recorded where a reader would look

- **`whoogle` is rejected, not promoted**: the repository is archived. Padding a roster to
  reach a count with a dead product is worse than a shorter roster.
- **`algolia-ai-search` does not move here.** It searches the catalogue the customer loads
  into it, so it belongs in `storage` beside Elasticsearch, Meilisearch, Typesense and
  Orama - the 2026-08-18 Qdrant/Milvus/Pinecone reasoning. It gets its own PR, because the
  move changes `storage`'s roster and its capability value has to be read against
  `storage`'s instrument rather than assumed to port.
- **Four records are scoped to one SKU** of a vendor that sells several, and each says so.

## Adoption

The client-package rule is stated once in the category comments and applied uniformly: a
first-party SDK counts as a usage-volume channel only where the vendor sells one product
(`serpapi` at 4, `linkup` at 3, both medium confidence); a multi-product SDK does not
(`parallel-search`, `you-com-search-api` abstain despite large numbers); a different-protocol
wrapper measures that protocol rather than the service (`brave-search-api` abstains).

Nine of twenty-one abstain, every one a closed service with no countable channel, each
naming the page the figure was looked for on.

## Review

Two Codex rounds on the design before any file was written, and one on the built result.
The design rounds killed the original nesting claim and forced the scoring unit to be
declared. The build round caught three places where the product-boundary test had been
applied to some vendors and not others - `ddgs` banded 3 for doing exactly what
`jina-reader` was banded 4 for, and `exa` and `tavily` at 5 on evidence that did not meet
band 5's own wording. All three moved, which is why the distribution in this PR is worse
than the one in its first commit and more defensible.

## Test plan

- `uv run pytest -q` - 2094 passed, 1 skipped, exit 0
- `build.check_refetch --product <slug> --all --strict` run on all 21 products: no
  fabricated digests anywhere
- Fixtures re-derived rather than carried over: identity coverage (github 258/377,
  huggingface 72/106, homepage 7/56), the identity edge fixture (171 tail rows), and
  `MEASUREMENTS_DIGEST`, whose move was measured by dumping and diffing the row sets - 378
  rows on both sides, five differing, only `category_slug` changing on each.

Closes #430.


ccerv1:
Stage impact, measured by serializing at `main` and at the head commit and diffing every category's stage:

```
search_retrieval:       none -> 3 (Viable Alternatives)
agent_tools_protocols:  5 -> 5, unchanged
```

That is the only change, and the second line is the one worth reading. `agent_tools_protocols` loses eleven products and its mature-open count does not move at all: it was 6 before and is 6 after, because **not one of the eleven that left was a fully-open product at 4.5 or above**. The retrieval layer it was carrying contributed nothing to its stage.

`search_retrieval` enters at 3 with **zero** mature open products of its own. Stage 3 is what the formula returns for a category with open products that do not reach maturity - `searxng`, `jina-reader` and `ddgs` are all open source and none of them clears 4.5 - and that is the finding this category exists to publish rather than a defect in the roster.

Labelled `stage-move` because a category appearing on the map is a deliberate change to the published gap set.

ccerv1:
CI note, and a papercut worth knowing about. This PR's `validate` failed twice on `stage moved without the stage-move label` while the label was already on it.

The label is read from the event payload - `contains(github.event.pull_request.labels.*.name, 'stage-move')` - and `validate` triggers on `pull_request` default types only, not `labeled`. So a payload frozen at open time never sees a label added afterwards, and `gh run rerun` replays that same frozen payload rather than fetching a new one. Closing and reopening the PR fires a fresh event and the run goes green.

Two ways to avoid it next time: add the label before opening the PR, or add `labeled` to the workflow's `pull_request` types. The second is a one-line change and I have not made it here, because it is unrelated to this PR.

ccerv1:
Adoption ruling applied, per Carl: the hyperscaler web tools band on `reported_traction` rather than abstaining, and Azure sits one below the other two.

```
openai-web-search     abstain -> 5   the Responses API surface the tool is enabled on
google-grounding-api  abstain -> 5   the Gemini and Vertex AI surfaces grounding is available on
azure-grounding-bing  abstain -> 4   the Microsoft Foundry Agent Service surface it ships in
```

The mechanism was already in the map and I had not applied it: `banded_quantity` exists for "a parent platform's reach, the product being a feature inside something larger", and `gemini-pro` bands 5 on the surfaces its tier powers rather than on a per-model count. Each record now carries the surface it was banded on, so a consumer reading `level: 5` off the warehouse sees what was actually read.

Azure is 4 on the nearest precedent - `azure-ai-model-inference` reads the Azure enterprise base the same way, and an enterprise cloud surface is broad rather than mass-market.

`google-programmable-search` is deliberately NOT included and still abstains: a metered legacy API with a hundred free queries a day is not a default path anything reaches the web through, so the parent-surface reading does not apply to it.

Category adoption distribution is now 5 at level 3, 8 at level 4, 2 at level 5, 6 abstaining - down from nine abstentions, and all six that remain are independent closed services with no countable channel.

Preflight 16/16, suite 2094 passed, 1 skipped, exit 0.

PR #598, opened 2026-09-16
Register 83 tail candidates from the omissions review

ccerv1:
## What

Adds **83** signal-only tail rows to `sources/registry/` for products the AI Stack Map does not yet carry, drawn from a targeted review of coverage gaps. Rebased onto current main (`fd793664`) and re-assessed against ADR-005 after the #430 split — see the review thread for the point-by-point response. Each candidate fits an existing published category, dedups clean, and every artifact identifier was confirmed reachable on 2026-09-16. No scores, product records or category rosters change.

This is the "before the map changes" step (`discover-candidates`). Rows are promoted later via `add-product` (all target categories are published).

## Reconciled counts (existing-category triage)

```
raw_signals       110   (review items that fit an existing published category)
duplicate_signals  10   (already in the corpus at branch time)
unique_candidates 100
accepted           83   (rows in this PR)
parked             17
```
`110 = 10 + 100` and `100 = 83 + 17`.

## Parked

- **ADR-005 (closed, long-tail / bundle — not the frontier):** `nvidia-nim` (closed platform bundle, fails "surface not bundle"), `msty` (closed consumer app), `neptune`, `comet-ml` (closed experiment-tracking long tail; `weave`/`opik` already represent that frontier).
- **Now head products on main** (dropped on re-dedup after the rebase): `clearml`, `puppeteer`, `selenium`, `ag-ui`.
- **Artifact already belongs to a head product** (pre-rebase parks): `simpleqa`, `browsecomp` (openai/simple-evals), `open-instruct` (tulu), `alignment-handbook` (zephyr), `cyberseceval` (llamafirewall), `vllm-ascend` (vllm).
- **Identity/family deferrals:** `mixtral`, `gpt-neox`, `magentic-one`.

Structural gaps needing a category that does not exist (image/video generation, speech, robotics, datacenter accelerators, model hubs, classic ML/CV) are routed to `category-proposal` issues (epic #606), not registered here.

## Split-taxonomy placement (#430)

`llama-parse` → `document_conversion`. The rows that had targeted `agent_tools_protocols` are gone entirely (`ag-ui`/`puppeteer`/`selenium` are now head products), so no pre-split file is recreated.

## Identity coverage re-pin

[...]

PR #616, opened 2026-09-17
Split agent_tools_protocols into protocols and connectors, at 20 and 21 products (#430)

### LINKED ISSUE #430: [category-proposal] Split agent_tools_protocols — three hypotheses, decided by capability coherence

ccerv1:
The last shape problem #430 opened. After the document converters left for
`document_conversion` and the retrieval layer for `search_retrieval`, what remained was
still two things — and its capability anchor was `docling`, which left on 2026-09-16.

**Its `scoring_recipe` carried an openness note and no capability section at all.** So this
is the first ladder either half has had, not a rewrite.

## Why it is two categories

Two ladder designs for a single combined category were written and **both failed review on
the same point**: a connector's capability can only be stated as a proportion of its target,
the targets are incomparable — GitHub's API, a Postgres instance, a browser, a sandboxed
directory — and any rule loose enough to span them is gamed by narrowing the target. The
second attempt's own worked examples contradicted its band table.

The repository's rule applied: *a rule that takes more than two attempts to state
consistently is drawing a line that is not there.* The line that **is** there separates
implementing a specification from adapting a system.

## `agent_protocols` — new, 20 products, every one open source

Capability is **how much of a named specification at a stated version** the product
implements, over role × primitives × transports. The version requirement is what makes the
band reproducible, and it does real work:

| band | who | what decides it |
|---|---|---|
| 5 | Go, Rust SDKs | cite the **current** 2026-07-28 revision *and* document sampling, elicitation, roots |
| 4 | Python, TypeScript, C#, Kotlin, PHP, Swift, `mcp-go`, `fastmcp` | each misses one leg, and a different one — C# and Kotlin name no version, Swift targets the superseded 2025-11-25, PHP names the current revision but no optional features |
| 3 | `mcp-inspector` | client role only |
| 1 | `fastapi-mcp` | one role, one primitive |

Eight abstain: four wire protocols, an extension spec, a packaging format, the registry —
and the Java SDK, which implements the contract but whose primitives no reachable document
enumerates.

## `agent_tools_protocols` — kept slug, renamed, 21 products

Capability is **what the adapter adds**: one system or many, does it hold credentials, does
it host execution. The credential test is the 3/4 line, and it is whether the *product* runs
the flow — not whether "OAuth" appears in a README, which is how `slack-mcp` was banded 4 in
an earlier draft and is now 3 beside `postgres-mcp`.

The finding: **three of the six band-5 products publish their source.** OpenConnector and
ACI.dev publish the OAuth machinery that Zapier and Pipedream sell.

## Roster growth

The protocol side found its roster where the #430 sweep never looked — that sweep ranked
repositories by stars across six subcategories and never enumerated the
`modelcontextprotocol` org, so seven official SDKs were never candidates. Ten promotions
each side; `puppeteer` and `selenium` come from draft PR #598.

## Review

Three Codex rounds — two on the designs, which killed the combined-category approach, and
one on the build. The build round caught: `slack-mcp` banded on the word OAuth; no record
naming a spec version though the ladder promised one; `algolia`'s note inventing a
requirement absent from storage's ladder; `java-sdk` banding down on missing documentation;
`klavis` scoring two different product scopes on its two axes; three records claiming
`ungated` from path absence against the rubric's own acceptance standard; and five stale
superlatives the promotions had falsified.

## Also in here

- `algolia-ai-search` moves to `storage`, re-read against that ladder. Its band rests on an
  evidence gap that is stated plainly: read NeuralSearch as satisfying storage's top rung and
  it is a 5, which is a maintainer's ruling and not something a category move should decide.
- `liteparse`, `opendataloader-pdf` and `omniparse` move to `document_conversion`'s registry.
- Identity coverage went **up** — thirteen new orgs, all declaring handles, github 258/377 →
  270/380 — so the baseline carries no `lowered_because`, which its guard would reject.

## Test plan

- `uv run python -m build.preflight --skip-tests` — all 17 steps pass
- `uv run pytest -q` — 2120 passed, 1 skipped, exit 0
- `check_refetch --product <slug> --all --strict` on all 20 promotions: no fabricated digests
- Two censuses re-pinned against measured diffs: channel-authority prose leg 18 → 19, and
  `MEASUREMENTS_DIGEST` (378 rows both sides, two differing, only `category_slug` changing)

Closes #430.


ccerv1:
Stage impact, measured by serializing at `main` and at the head commit. **This is the largest stage move any of the #430 pull requests has made, and it needs a decision rather than just a label.**

```
agent_protocols:        none -> 5 (Mature Open Ecosystem)    mature-open 0 -> 6
agent_tools_protocols:  5    -> 3 (Viable Alternatives)      mature-open 6 -> 0
```

The six mature open products did not disappear - they moved. `model-context-protocol`, `mcp-python-sdk`, `mcp-typescript-sdk` and `fastmcp` were four of the six holding `agent_tools_protocols` at stage 5, and all four are specification-layer products that now sit in `agent_protocols` alongside `ag-ui` and `mcp-apps`. `playwright-mcp` and `browser-use` stay in the connector category and neither clears the 4.5 maturity bar.

So the drop is not a regression in the connector layer; it is the removal of a score the connector layer was never earning. Stage counts fully-open products at 4.5 or above, and every one of that category's six was a protocol or an SDK. What is left - the MCP servers, the browser tooling, the connector platforms - is genuinely a stage-3 layer today: sixteen of twenty-one publish their source, and none of them reaches 4.5 maturity, mostly because adoption for a single-target MCP server bands on stars.

Two things follow, and the second is yours:

1. The published map will show `agent_tools_protocols` falling from Mature Open Ecosystem to Viable Alternatives on the day this merges. That is the honest number, and anyone reading the front end will see a two-stage drop without the explanation above unless the category comment carries it. It now does.
2. **If that reads wrong to you, the lever is not this PR** - it is whether stage should count a category's products or its layer's. Nothing here moved a product's score; the products were re-filed into the category that measures what they do.

Labelled `stage-move`.

PR #618, opened 2026-09-17
Rename agent_tools_protocols to agent_tools_connectors

ccerv1:
The slug stopped describing the category when #430 moved the protocols and the SDKs out into
`agent_protocols`. Its display name already reads "Agent tools & connectors"; this brings the
slug with it.

## The rule for the rest of the corpus

Category slugs carry no alias mechanism the way product slugs do, so the renamed category's
comments hold the mapping. The rule applied everywhere else:

- **Updated** — references that name the category as a thing that exists *now*, because a
  reader following one has to be able to find it. Three, across `agent_protocols` and
  `search_retrieval`.
- **Left alone** — references that record what was decided while that was the name. 31 in
  `sources/resolution_ledger.yaml` under #413 verdicts, and four score notes saying a product
  "moved out of agent_tools_protocols". Rewriting them would make the record say something
  untrue of its own date, which is the opposite of what a ledger is for.

## Two mechanisms did not model a rename

Both were taught rather than worked around, and both are reusable.

**The externalization receipt** watches every path under `sources/` and reads a `git mv` as
an unaccounted deletion — correctly, because neither an externalization nor a retirement had
happened. The file moved. Its nearest available word, a retirement, names a warehouse
**table** and archives its bytes; filing a renamed category file as one would have put a false
statement in an audit document to make a gate pass. Schema version 5 adds a `renames` block:
deliberately thinner than a retirement — no hashes, because nothing is archived — and `to` is
checked to exist so an entry cannot launder a deletion by naming a file nobody wrote.

**`build/render.py`** took `_ORDER` from the committed payload and straplines from live
sources, then indexed one into the other. Those lag each other by design — the payload is
bot-owned and regenerated after merge — so a removed category raised `KeyError` at module
scope, which is an import error, which took the whole of `tests/test_render.py` down as a
**collection** error rather than one readable failure. A category *new* in sources was
already tolerated; the opposite case now behaves the same way. This was not a rename bug; the
rename is just the first thing to hit it.

## What this does not break

Checked before starting, because I had assumed the opposite:

- `roadmap.json` sets `map_category_ids` on **zero** products, so the FK check that exits 1 has
  nothing pointing at a category.
- `gap-map-links.json` carries no category references.
- `categoryId` in the monorepo's `roadmapPublic.generated.ts` is **derived** — the generator
  reads whichever category a product is in from the synced payload — so the next
  `npm run build:roadmap` writes the new value itself.
- Category no longer appears in deep links; those moved to `productUrl(slug)`.

Worth telling whoever picks up the front end: the next data sync regenerates
`roadmapPublic.generated.ts` with new `categoryId` values. Two of those (Firecrawl, SearXNG)
are already stale from #614 and are not caused by this change.

## Test plan

- `uv run python -m build.preflight --skip-tests` — all 17 steps pass
- `uv run pytest -q -n auto -m "not serial"` — 2123 passed, then 1 passed serial
- Four new tests: three on the `renames` block (accounts for the deletion, rejects an entry
  pointing at nothing, and the live receipt resolves), one on the render lag
- `MEASUREMENTS_DIGEST` re-pinned against a measured diff — 378 rows both sides, six
  differing, only `category_slug` changing on each
- `tests/goldens/corpus.json` deliberately NOT regenerated: `build.goldens --check` reports it
  stale as a **notice**, which is the expected state before the bot's post-merge commit


ccerv1:
Two more mechanisms had to learn about renames, beyond the externalization receipt already in this PR.

**`build/render.py`** took `_ORDER` from the committed payload and straplines from live sources, then indexed one into the other. Those lag by design - the payload is bot-owned and regenerated after merge - so a removed category raised `KeyError` at module scope, an import error, which took all of `tests/test_render.py` down as a **collection** error. A category *new* in sources was already tolerated; the opposite case now behaves the same way. Not a rename bug - the rename is just the first thing to hit it.

**`build/check_corpus_diff.py`** requires every axis row of an untouched product to be byte-identical, and the rename moves `category_slug` on all 21 rows while touching none of those products' files. Renamed slugs are now detected from the diff rather than declared, and only that one column is projected out, only for those products. A product moving between two categories that both still exist is still flagged.

The gate now reports the rename cleanly:

```
agent_tools_connectors: None -> 3   +21 / -0
agent_tools_protocols:  3 -> None   +0 / -21
untouched-product row changes: none
```

One thing worth saying plainly about the shape of this PR. **Three independent mechanisms treated a category slug as immutable** - the externalization receipt, the notebook renderer and the semantic-diff gate - and none of them was wrong to; they were written for a corpus where categories are added and populated, not renamed. That is why my first instinct was that renaming would be expensive. I was right about the cost and wrong about where it sat: not in the front end, which derives the slug, but in three gates that had no word for the event. Each now has one, and each is reusable for the next rename.

Suite: 2126 passed, 1 skipped, then 1 serial. Preflight 17/17.

ccerv1:
Labelled `stage-move`, and the label is doing its job rather than being worked around: no product's stage changed, but the published stage table gains a row and loses one, which is exactly the visible change the label exists to make a human affirm.

```
agent_tools_connectors: None -> 3
agent_tools_protocols:  3 -> None
```

I considered teaching this leg about renames the way the other three were taught, and did not. The other three were reporting something false - a deletion that was a move, a crash on a lagging payload, a silent rewrite that was a column rename. This one is reporting something true.

Issue #573, opened 2026-09-14
Two docstrings in check_channel_authority contradict their own code, and nothing gates code prose


ccerv1:
Found while answering a question about whether this repo carries too many comments. The answer was no — the ratio is not the fault — but the investigation turned up something worse than verbosity, and a gap that let it survive.

## The two defects, both verified

**`parse_version` describes the opposite of what it does.** Its docstring offers `openlit-2.0.0` as an example of a tag that carries no version:

> A tag like `2026-08-01` or `openlit-2.0.0` is not evidence of an absent lag; it is evidence that this comparison could not be made.

It parses fine. `_VERSION` has a digit-lookbehind precisely so a prefixed tag still yields a version:

```
openlit-2.0.0  -> (2, 0, 0)
2026-08-01     -> None          # this half of the example is correct
v1.2.3         -> (1, 2, 3)
```

A reader is told a supported tag shape is undecidable.

**`newest_pypi_release` names the wrong source for its own return value.** The docstring says:

> Read off the release history rather than off `info.version`, because the upload date is half the finding.

The code reads `info.version` and uses `releases[version]` only to find that version's upload timestamp. The reason given is right; the mechanism described is not.

## Why they survived

**Nothing gates prose in `build/`.** Every prose gate in this repo points at `sources/` or `docs/` — `test_score_notes.py`, `test_product_prose.py`, `test_externalized_table_docs.py`, `test_docs_integrity.py`. No test reads a docstring in `build/`, so a docstring can contradict its function indefinitely.

The registry half is well covered by comparison: five gates plus the `clean-score-notes` skill, which already carries the right keep/cut test — a clause stays if deleting it would leave a reader unable to reconstruct the score.

## What is worth building, and what is not

**Not a prose-cleanup pass.** Measured: `build/*.py` is 17,460 code lines against 8,663 of comment and docstring, and the registry's 2,058 score notes have a median of 48 words with only 3% over 150. The volume is not the problem, and the comments are load-bearing — four separate times in one session a comment in `build/` was the thing that settled a question. A cutting exercise would put those at risk for no gain.

**Two mechanical gates, modelled on the score-note ones that already work:**

1. **No live census in an executable docstring.** This module's own docstring carries "Four records today", "Two records today" and a seven-product roster, while explicitly saying it avoids duplicating a moving count. It appears in `--help`. Same fault the date rule catches one directory over.
[...]

PR #624, opened 2026-09-20
Correct two docstrings that contradict their code, and gate prose in build/

ccerv1:
# build-prose-gates — pass

**Goal.** Two docstrings in `build/check_channel_authority.py` state the opposite of what their own code

## Success criteria
- [ ] `parse_version`'s docstring no longer offers `openlit-2.0.0` as an example of a tag carrying no version. It parses to `(2, 0, 0)` — `_VERSION` has a digit-lookbehind precisely so a prefixed tag still yields a version. The `2026-08-01` half of the example is correct and stays.
- [ ] `newest_pypi_release`'s docstring names the source it actually reads. It reads `info.version` and uses `releases[version]` only to find that version's upload timestamp; the stated *reason* ("the upload date is half the finding") is right and should be kept.
- [ ] **Gate 1 — no live census in an executable docstring.** The module docstring carries "Four records today", "Two records today" and a seven-product roster while saying it avoids duplicating a moving count, and it appears in `--help`. Model the check on the score-note date rule one directory over; the test names the file and the offending phrase.
- [ ] **Gate 2 — a docstring that quotes an example input is not contradicted by its function.** Extract the quoted value, call the function, compare against what the prose claims. This is the check that would have caught `parse_version` on the day it was written. It needs to work on at least `check_channel_authority.py`; state in the PR body which modules it covers and how a module opts in.
- [ ] Both gates fail on the pre-fix content and pass after — demonstrate by running them against the parent commit, and quote the failure text in the PR body
- [ ] `uv run pytest -q -n auto -m "not serial and not network"` passes, and the PR body quotes the count
- [ ] PR title names the change. Not "Apply #573" and not a truncated goal sentence.

## Outcome
`pass` after 2 round(s). Branch `loop/build-prose-gates-2026-09-20-a1`.

## Final verdict
```json
{
 "verdict": "pass",
 "blocking": [],
 "non_blocking": [],
 "settled_now": [
  "Round 1 blocker is fixed: Gate 2 extracts the parent commit\u2019s original prose and reports parse_version(`openlit-2.0.0`) -> (2, 0, 0), docstring says None.",
  "All four gate checks pass on current content. The extraction convention is sufficiently narrow for the opted-in module."
 ],
 "escalation": null
}
```

## Settled ground this run added (8)
- The comment volume in `build/` is not the problem and is not in scope. The issue investigated
- Comments in `build/` are load-bearing. Correct wrong sentences; do not delete correct ones.
- `check_instrument.py`'s prose-reading detector is out of scope and must not change behaviour.
- Both corrected function docstrings accurately describe their implementations.
- Gate 1 detects the parent commit's live counts and product rosters.
- Describing _VERSION as a non-digit prefix match rather than a lookbehind is correct.
- Round 1 blocker is fixed: Gate 2 extracts the parent commit’s original prose and reports parse_version(`openlit-2.0.0`) -> (2, 0, 0), docstring says None.
- All four gate checks pass on current content. The extraction convention is sufficiently narrow for the opted-in module.

Later runs of this job inherit `/workspace/GitHub/personal/agent-loops/jobs/settled/build-prose-gates.json`; prune it there.


ccerv1:
## What I did
- files touched:
  - `build/check_channel_authority.py` — corrected `parse_version`'s and `newest_pypi_release`'s docstrings, and removed the live counts/rosters the module docstring carried into `--help`. No behaviour change: prose only, plus two new `Examples:` blocks (`parse_version`, `is_prerelease`).
  - `tests/test_build_docstrings.py` (new) — the two gates.
- commands run:
  - `uv run python -c "from build.check_channel_authority import parse_version; ..."` → `openlit-2.0.0` → `(2, 0, 0)`, `2026-08-01` → `None`. Both defects verified first-hand, not taken from the issue.
  - `uv run pytest -q tests/test_build_docstrings.py` → 4 passed.
  - Gate 1 + gate 2 against the parent commit's module (`git checkout HEAD -- build/check_channel_authority.py`, run, restore):
    ```
    AssertionError: 5 `--help` docstring(s) state a live count of the corpus:
        check_channel_authority.py: 'Four records'
        check_channel_authority.py: 'Two records'
        check_channel_authority.py: '`areal`, `gpt-researcher`, `langtrace`, `openlit`, `sageattention`, `swe-agent`, `xtuner`'
        check_channel_authority.py: '`flash-attention`, `khoj`, `lance`, `mineru`'
        check_channel_authority.py: 'seven fires'
    AssertionError: build.check_channel_authority is in CHECKED_MODULES but documents no `Examples:` block, so nothing in it is checked.
    ```
  - Gate 2 against the pre-fix *claim* written in the new convention (`` `openlit-2.0.0` -> None ``):
    ```
    AssertionError: 1 documented example(s) in build.check_channel_authority contradict the function:
        parse_version(`openlit-2.0.0`) -> (2, 0, 0), docstring says None
    ```
  - `uv run python -m build.validate` → `0 error(s), 2 warning(s)` (both pre-existing model_families pattern-overlap warnings).
  - `uv run pytest -q -n auto -m "not serial and not network"` → **2213 passed, 1 skipped in 414.90s**.
  - `uv run python -m build.check_channel_authority --help` → renders.
  - Tests were run to completion before `git add`, never concurrently (the `build/assets.py` `git ls-files` race).

## Why
- decisions:
  - **`parse_version`.** Kept the `2026-08-01` half, dropped `openlit-2.0.0`, and added the positive statement: `_VERSION` anchors on any non-digit before the number (`(?:^|[^0-9])`) rather than on the string start, so a package-prefixed tag parses like any other — which is what lets `openlit` reach a verdict at all. Note the issue/goal text calls this a "digit lookbehind"; the lookbehind `(?<=\d)` is in `_PRERELEASE`, not `_VERSION`. I described what the code does, not what the issue said.
  - **`newest_pypi_release`.** Now names `info.version` and says the release history is read only to date it (earliest file upload under `releases[version]`). The stated reason — the upload date is half the finding — is kept verbatim in substance.
  - **Module docstring.** Removed: "Four records … — `flash-attention`, `khoj`, `lance`, `mineru`"; "Two records today: `chroma` and `mlflow`"; "three pre-release exclusions and fourteen undecidable where this gate has four and two"; "The seven fires are identical — <seven slugs>"; "four categories". Every claim they supported is kept in a form that doesn't copy a moving number ("the report prints the excluded records", "the simulation left far more products undecidable and excluded fewer", "the fires are identical — the run prints them"). Also fixed the undecidable bullet, which listed "per-package monorepo tags" as unparseable — contradicting the same `parse_version` defect.
  - **Gate 1 shape** follows `test_score_notes.py`'s date rule: a regex, a named allowlist frozen with justification, and a staleness test so a rewritten phrase must leave the list. It covers every `build/*.py` containing `description=__doc__` — i.e. every module docstring that is `--help` output — so a module is in scope by construction, not by opting in. The detector is a cardinal ≥ two plus one of the nouns this repo counts and a build module recomputes (`records/products/fires/notes/slugs`), plus a roster rule: three or more comma-separated backticked tokens that are *all* real product slugs. The slug test keeps field lists (`note`, `sources`, `confidence`) out; the threshold of three keeps "Both catch `areal` and `xtuner`" (reasoning) in.
  - **Gate 1 backlog.** 15 phrases in 10 other modules, frozen and named per phrase. Two of them (`check_capability` "two different products", `serialize_registry` "Two structural notes") are the detector reading an argument's shape as a census; they are listed with that justification rather than excused by loosening the regex, because every loosening I tried also let a real census through. `check_instrument.py` gets one roster + three count entries in the allowlist — **an allowlist entry only, no edit to that file and no change to its under-coverage detector**.
  - **Gate 2 convention** (written into the gate's own docstring, so gate 1 holds me to it): inside a function docstring, a line `Examples:` opens a block, and each following `` `<argument>` -> <repr of expected return> `` line is executed — the backticked token is passed as a string, and `repr()` of the return is compared to the claim. The block ends at the first non-example line. Deliberately single-string-argument only: that is the shape of both functions whose examples were wrong, and `lag_verdict` (two parameters) documents its cases in prose and is out of scope. Modules opt in via `CHECKED_MODULES` (`build.check_channel_authority` today) **and** must document at least one example — a second test asserts the coverage, so a module cannot join the list and check nothing.
  - **New test module, not an existing one.** `test_score_notes.py` and `test_product_prose.py` both take the corpus fixture and are about a curator's prose; these two parse `build/*.py` and are about a maintainer's. They share no input and no fixture, and the file's docstring says so. This is the "unless that forces a worse design" branch the plan allows; everything else stayed in `tests/` where the prose gates live.
- alternatives rejected:
  - Gate 1 as an opt-in module list (like gate 2): rejected — the plan says model it on the date rule, which is corpus-wide with a named allowlist, and an opt-in list would have covered exactly the one file already fixed.
  - Gate 2 extracting examples from free prose ("A tag like `X` … is not evidence"): rejected as an NLP guess; a declared marker is checkable and the plan sanctions it.
  - Re-banding or rewriting anything in the other ten modules: out of scope for #573.

## Risks / unsure about
- Gate 2 covers one module today. That is the floor the goal set, but until a second module opts in the convention is unproven against another author's docstring style.
- Gate 1's backlog is keyed on whitespace-normalized phrases. A rewording that keeps the count (e.g. "472 products" → "the 472 products") fails the gate rather than silently passing — correct direction, but it will read as a surprise to whoever reflows one of those docstrings.
- `CENSUS` excludes the singular ("one product") and the noun `categories`; a docstring can still state a live count as "sixteen categories" or in a sentence with no cardinal ("this gate has four and two" — which I removed by hand, not by gate). The gate is a floor, not a proof.
- The commit message and this note assert `_VERSION` has no lookbehind, contradicting the issue's wording. If the reviewer believes the issue, the code is the tiebreak: `_VERSION = re.compile(r"(?:^|[^0-9])v?(\d+)\.(\d+)(?:\.(\d+))?")`.

## Response to round N-1
|---|---|---|---|---|
| (round 1 — nothing to respond to) | — | — | — | — |

ccerv1:
## What I did
- files touched:
  - `tests/test_build_docstrings.py` — gate 2 gained a second extractor, `prose_none_claims`, so it reads the prose shape the #573 claim was actually written in and not only the `Examples:` block convention this PR introduced. The coverage assertion now passes `prose=False`. Nothing else changed; `build/check_channel_authority.py` is untouched this round.
- commands run:
  - `uv run pytest -q tests/test_build_docstrings.py` → **4 passed** on current content.
  - Both gates against the parent commit (`git show cf363838^:build/check_channel_authority.py > build/...`, run, restore), **with the parent's own unmodified prose** — nothing rewritten into the new convention first:
    ```
    AssertionError: 5 `--help` docstring(s) state a live count of the corpus:
        check_channel_authority.py: 'Four records'
        check_channel_authority.py: 'Two records'
        check_channel_authority.py: '`areal`, `gpt-researcher`, `langtrace`, `openlit`, `sageattention`, `swe-agent`, `xtuner`'
        check_channel_authority.py: '`flash-attention`, `khoj`, `lance`, `mineru`'
        check_channel_authority.py: 'seven fires'

    AssertionError: 1 documented example(s) in build.check_channel_authority contradict the function:
        parse_version(`openlit-2.0.0`) -> (2, 0, 0), docstring says None

    AssertionError: build.check_channel_authority is in CHECKED_MODULES but documents no `Examples:` block, so nothing in it is checked.
    ```
    `3 failed, 1 passed` — the contradiction test now fails on the original sentence itself, which is the round-1 gap.
  - `uv run python -m build.validate` → `0 error(s), 2 warning(s)` (the two pre-existing `model_families` pattern-overlap warnings).
  - `uv run pytest -q -n auto -m "not serial and not network"` → **2213 passed, 1 skipped in 363.52s**.
  - Suite run to completion before `git add`; never concurrent (the `build/assets.py` `git ls-files` race).
  - Commit added on `loop/build-prose-gates-2026-09-20-a1`: `41c496ee`. Nothing amended or rebased.

## Why
- decisions:
  - **The second extractor, and how narrow.** The parent's claim was a sentence, not a block: "A tag like `2026-08-01` or `openlit-2.0.0` is not evidence of an absent lag". The convention now reads one prose shape as well — a run of backticked tokens joined by `,` / `or` / `and`, immediately followed by a denial phrase — and the claim it asserts is always `None`, each token passed on its own. Written into the gate's own comment block, so gate 1 holds me to it.
  - **Two guards so it is a rule and not a reading of English.** The denial vocabulary is closed and listed in the source (`is/are not evidence`, `carries/carry no version`, `does/do not carry|parse`) rather than inferred from the sentence; and the function must both declare `None` in its summary line and take exactly one required parameter. A sentence therefore cannot assert a return the function has no way to produce, and a two-parameter function (`lag_verdict`) or a defaulted one (`release_lines`, whose "a transport failure is not evidence of a trailing channel" has no backticked token before the denial anyway) is out of reach.
  - **Positive prose claims stay unread.** "so `openlit-2.0.0` yields a version like any other tag" is not extracted. Reading an expected *value* out of English is the guess I refused in round 1 and still refuse; a positive claim goes in an `Examples:` block, where the value is declared.
  - **Coverage counts declared blocks only.** `test_a_checked_module_actually_documents_an_example` now calls `_checked_functions(module_name, prose=False)`. The prose extractor fires on whatever sentences a module happens to contain, so counting its hits as coverage would let a module join `CHECKED_MODULES` and satisfy the assertion without anyone writing a checkable example — exactly the hollowness that assertion exists to prevent.
  - **`check_instrument.py` untouched**, as in round 1: it holds allowlist entries in `CENSUS_BACKLOG` and nothing else. Its under-coverage detector is unchanged and unread by either gate.
- alternatives rejected:
  - Extracting from any backticked token in a docstring and calling the function on it: would have caught this one, and also called `parse_version("_VERSION")`, `parse_version("/tags")` and every other incidental code span. Noise, not a gate.
  - Inferring the expected value from the surrounding sentence: an NLP guess dressed as a check. The closed denial vocabulary is the same idea reduced to something a reader can audit in four lines.
  - Leaving round 1's demonstration as-is and arguing the rewrite-then-run was sufficient: it was not. Run against the parent's own bytes, gate 2 extracted zero examples and passed. That is the reviewer's point and it is correct.

## Risks / unsure about
- The denial vocabulary is three phrases. A fourth way of writing "this yields nothing" (`returns nothing for`, `is ignored`) passes silently. The gate is a floor; widening it is a one-line change with a test to prove the widening catches something.
- `_takes_one_required_string` checks arity, not annotation, so a single-required-parameter function taking a non-string would be called with a string if someone wrote a denial sentence about it. In `check_channel_authority` there is no such function; across a second opted-in module there could be, and the failure would be a `TypeError` in the gate rather than a wrong-answer report. Legible, but not a clean message.
- Gate 2 still covers one module (`CHECKED_MODULES = ("build.check_channel_authority",)`). The prose leg makes joining cheaper — a module now gets some coverage without anyone writing a block — but the convention remains unproven against a second author's style.
- Round 1's caveats stand: `CENSUS` excludes the singular and the noun `categories`, and `CENSUS_BACKLOG` is keyed on whitespace-normalized phrases, so reflowing one of the 15 backlogged phrases while keeping its count fails the gate rather than passing silently.

## Response to round N-1
|---|---|---|---|---|
| `tests/test_build_docstrings.py:196` — gate 2 extracts zero examples from the parent commit, so the contradiction test passes without calling `parse_version`; the demonstration rested on rewriting the claim into the new convention first | ✅ | — | — | Added `prose_none_claims` with a closed denial vocabulary plus the `None`-in-summary and single-required-parameter guards. Run against the parent commit's unmodified `build/check_channel_authority.py`, gate 2 now fails on the original sentence: `AssertionError: 1 documented example(s) in build.check_channel_authority contradict the function: parse_version(`openlit-2.0.0`) -> (2, 0, 0), docstring says None`. Parent run is `3 failed, 1 passed` (census, contradiction, coverage); current content is `4 passed`. Coverage assertion switched to `prose=False` so it cannot be satisfied by incidental prose hits. Commit `41c496ee`. |

Issue #604, opened 2026-09-16
[policy] No written inclusion rule for closed / platform products


ccerv1:
## The gap
Closed and platform products are included **inconsistently**, with no stated rule:

- **In:** ChatGPT, Claude, Gemini, Grok (`ui_api`); Bedrock Guardrails (`safeguards`); Vertex/Bedrock evaluations (`evaluation_code`); many hosted fine-tuning APIs (`finetuning_code`).
- **Out:** SageMaker, Vertex AI, Azure ML (as platforms); Weights & Biases* ; NVIDIA Triton Server*; Unstructured*; Replicate / Fal / RunPod.

(* candidates for these are registered in #598, but the platform/cloud ones above still have no rule that says whether they belong.)

There is no documented predicate that explains this split, so every borderline closed/platform product is decided ad hoc — which is exactly how the inconsistency above arose.

## Ask
A short ADR (or a rule in `docs/reference/identity.md` / a boundary doc) defining **when a closed or platform product is in scope**, e.g. by asking: does it occupy a stack layer the map already scores, and can at least one axis (openness/adoption/capability) be evidenced for it? The review called this its sharpest governance point: the aperture, not the source registries, is what is inconsistent.

This unblocks clean decisions on the cloud-platform and serving-fabric candidates parked out of #598 (SageMaker/Vertex/Azure ML, Replicate/Fal/RunPod), and on a possible model-serving category (KServe/BentoML/Triton/Seldon) distinct from the orchestration question in #429.

_Filed from the 2026-09-16 AI Stack Map omissions review triage._

ccerv1:
Settled by ADR-005, accepted 2026-09-20 and merged in #628: `docs/architecture/adr-005-closed-product-inclusion.md`.

The rule Carl gave: the map needs representations of the best-in-class closed products to mark the frontier, and does not need the long tail of closed products. Score 4 or higher is the initial screen — **as a principle, not a hard rule**, which is why the ADR opens by saying it is guidance a curator applies rather than a predicate a build evaluates, and why nothing in `build/` or `tests/` fails on a product below the line.

The tail was surveyed against it in the same PR: 81 above, 61 below, 3 unmeasured of 145.

PR #628, opened 2026-09-20
Write the inclusion principle for closed products, and survey the tail against it

ccerv1:
# closed-product-inclusion-rule — pass

**Goal.** Closed and platform products are included with no written guidance, so every borderline case is

## Success criteria
- [ ] An ADR under `docs/architecture/` stating the three tests, recording that 1 and 3 codify existing practice, and quoting Carl's ruling as the source of test 2
- [ ] The ADR states which population it governs (openness score, openness class, or the serializer's closed bucket) and why, and does not use the other two interchangeably
- [ ] A `build/` module producing the census: every closed product failing test 2, with the values it was judged on and the comparison against its category
- [ ] **Each census row carries a measured stage-and-gaps delta** for removing that product alone, computed by rebuilding the payload with it withheld. A row asserting "no effect" without the rebuilt numbers is the defect the review caught; do not ship one
- [ ] The census names any product whose removal WOULD move a stage or a gap, and marks it as a decision for Carl rather than a candidate
- [ ] The survey applies the line as stated — `overall_score >= 4`, reading `capability >= 4` where `overall_score` is null — and reports three distinct states, never two: above the line, below it, and **unmeasured**. "Below" and "unmeasured" need opposite work and must not be summed
- [ ] The survey reproduces the 59 / 48 / 38 split and the 81-above figure, or explains the difference before going further
- [ ] **No test asserts the line.** A test that fails when a product sits below it turns the principle into a gate, which is the one thing this job must not do
- [ ] **Nothing is removed from `sources/`.** No product file is edited or deleted
- [ ] No gate is written for any of the three tests, and the ADR says why: tests 1 and 3 describe practice, test 2 is a judgement. If a future gate is wanted for test 1, the PR body may note it as an option — it does not open an issue committing to one
- [ ] `uv run pytest -q -n auto -m "not serial and not network"` passes, count quoted in the PR body

## Outcome
`pass` after 1 round(s). Branch `loop/closed-product-inclusion-rule-2026-09-20-a4`.

## Final verdict
```json
{
 "verdict": "pass",
 "blocking": [],
 "non_blocking": [],
 "settled_now": [],
 "escalation": null
}
```

## Settled ground this run added (3)
- Carl's ruling: best-in-class closed products in, long tail out — **as a principle**. The
- The three killed designs above: leader-relative predicate, invariance-by-assertion, lexical gate.
- Nothing is removed in this PR. The census is advisory.

Later runs of this job inherit `/workspace/GitHub/personal/agent-loops/jobs/settled/closed-product-inclusion-rule.json`; prune it there.


ccerv1:
## What I did

Started from `origin/recovered/closed-product-a3` as the plan directed, cherry-picked it onto
current `main` (`1fe61b55`, four merges on from the base it was written against — clean
auto-merge in `AGENTS.md`, no conflicts), then read it as someone else's work and added a
second commit for what that read found.

- files touched:
  - `docs/architecture/adr-005-closed-product-inclusion.md` (new, 172 lines) — the ADR: three
    tests, which population it governs, why there is no gate, the measurement it was decided on.
  - `build/closed_inclusion.py` (new, 399 lines) — the survey, the census, and the per-product
    withheld-rebuild delta. `uv run python -m build.closed_inclusion`, `--json`, `--all`.
  - `tests/test_closed_inclusion.py` (new, 240 lines) — 13 tests.
  - `AGENTS.md` — a `Surveys` family in the `build/` map, saying never to wire one into a gate.
  - `docs/README.md` — ADR-005 in the architecture table.
  - `docs/workflows/add-product.md` — points a closed or platform-surface product at the ADR.
  - **Nothing under `sources/`.** `git diff --name-only 1fe61b55 | grep -c '^sources/'` → 0.
- commits (added, never amended):
  - `1d5f85f5` docs(architecture): write the closed-product inclusion principle and survey the tail
  - `c840ff2e` Stop the census test depending on where the line falls, and say what really holds test 3 up
  - `uv run python -m build.closed_inclusion` → 145 population, **81 above / 61 below / 3
    unmeasured**; sub-counts `overall_score >= 4: 59`, `< 4: 48`, `overall_score null: 38`
    (capability ≥ 4: 22, < 4: 13, none: 3). Reproduces the pinned 59 / 48 / 38 split and the
    81-above figure exactly; 59 + 48 + 38 = 145 and 59 + 22 = 81.
  - `uv run python -m build.closed_inclusion --all --json` → all 145 withheld one at a time,
    each with a rebuilt payload: **0 products move a stage or a gap**. This is what lets the
    ADR say it about the whole population rather than only the 64 in the default census.
  - `uv run python -m build.validate` → **0 error(s), 2 warning(s)** (both pre-existing
    `model_families` pattern-overlap warnings, untouched by this branch).
  - `uv run pytest -q tests/test_closed_inclusion.py` → 13 passed in 16.01s.
  - `uv run pytest -q -n auto -m "not serial and not network"` → **2239 passed, 1 skipped in
    371.32s**.
- proposed PR title: **Write down why closed products are on the map, and survey what sits below
  the line**. PR body quotes: 2239 passed, 1 skipped (`-n auto -m "not serial and not network"`);
  `build.validate` 0 errors; survey 81 / 61 / 3 over a population of 145. The body notes, as an
  option and not a commitment, that test 1 is the only one that could carry a gate one day and
  would need a declared category capability to do it. No issue is opened for that.

## Why

- decisions:
  - **Population is `openness.score <= 1`, stated in the ADR and enforced nowhere else.** Three
    sets in this repo are called "closed": the score, `openness.class`, and the serializer's
    `_gap_bucket`, which folds `documented` and `restricted` into `closed` so they can fire an
    openness gap. The ADR reads the score because the inclusion judgement is about how much of
    the product is open, and because the screen's fallback axis (`capability`) is on the same
    scale. Verified the ADR's claim that the score set and the class set coincide today: the
    corpus has 2 products at `(closed, 0)` and 143 at `(closed, 1)` = 145, and every other class
    scores 2 or above.
  - **The delta is measured, never asserted.** `_stage_and_gaps` computes `mature_anywhere` over
    *all* products, so a closed product can hold a category off Stage 0 and can fire the
    `openness` gap. Every census row carries its category's stage and gap set rebuilt with that
    product withheld from every roster it appears in, printed whether or not it differs, and the
    diff is taken across all categories rather than only the product's own.
  - **The test that asserted the corpus's position against the line is gone.** The first draft
    opened its census test with `assert rows`. That reads as a sanity check and is really a
    claim about where the corpus sits: it goes red the day nothing is below the line, which is
    the day the principle has been fully acted on. It is the forbidden gate pointed the other
    way. It is now checked against the survey it was drawn from
    (`len(rows) == counts[BELOW] + counts[UNMEASURED]`, plus set equality on
    `(category, slug)`), which holds at any count including zero.
  - **ADR-005 overclaimed what enforces test 3.** It said `check_citations` and the verification
    workflow "already require of every product" at least one evidenced axis. They do not:
    `docs/schemas/score.schema.json` requires the three axis blocks and no `sources` entry under
    any of them, and the evidence gates constrain the evidence a score *records* rather than
    requiring it record any. Checked the corpus instead — all 145 closed products carry at least
    one source on at least one axis, with nothing enforcing it — and rewrote the paragraph to
    say that. This repo gated docstrings that contradict their code four commits ago (#624); an
    ADR making the same mistake about its own gates would not have survived review.
  - **`--all` now names the population it printed.** It censused 145 rows under a heading reading
    "products not above the line".
  - **The no-gate surface check globs `*.yaml` as well as `*.yml`.** Every workflow is `.yml`
    today, so the gate was correct and narrow; a gate that is right only for the filenames that
    happen to exist is the class of gap, not an instance.
  - Also verified the ADR's factual claims rather than inheriting them: `sagemaker-pipelines`,
    `azure-machine-learning-pipelines`, `triton`, `claude-ai`, `weave` and exactly four
    `vertex-ai-*` products exist, so #604's "missing" table is wrong; `vertex-ai-pipelines` and
    `vertex-ai-tuning` sit in `ml_orchestration` and `finetuning_code`, two categories;
    `google-document-ai`, `azure-document-intelligence` and `nvidia-run-ai` are all above the
    line at capability 5 with a null overall score, so the null fallback is load-bearing exactly
    where the ADR says.
- alternatives rejected:
  - **A gate on any of the three tests.** Tests 1 and 3 describe practice, so a gate would
    re-implement a convention the corpus already keeps. Test 2 is a judgement, and `4` is a
    policy number Carl set as a starting point; a build failing because a product scored 3.7
    would convert it into a threshold, which is the part he declined to make binding. The ADR
    says this in *Why there is no gate* and a test asserts the survey appears in neither
    `build/preflight.py` nor any CI workflow.
  - **A leader-relative predicate** ("in if it matches the best open product in its category") —
    fails both ways: where the open peers sit at 5/1 and 1/5 it turns away a closed product at
    4/4 that beats both on blended score; where they are weak it admits a proprietary tail.
  - **Invariance by assertion** ("removing a non-leading closed product cannot move a stage") —
    false, per `mature_anywhere`, and the reason the census rebuilds instead of promising.
  - **A lexical gate on test 1 (surface, not bundle).** Nothing about a bundle is visible in a
    slug, and `spaces`, `aws-lambda` and `google-cloud-run` are legitimately rostered in
    `deployment`.
  - Keeping a live count in the ADR's *Decision* body. The finetuning_code claim stays
    qualitative ("most of the closed side"); the numbers live in a section dated as a
    measurement, because a number in an ADR's argument rots silently.

## Risks / unsure about

- The 145-wide `--all` sweep returning zero movement is a fact about today's corpus, not a
  property. A future roster where a closed product is a category's only category-leading
  product would produce a *decision for Carl* row. That path is only covered by the synthetic
  corpus test, which is deliberate — covering it on the real corpus would mean pinning a
  product's position, which is the gate this job must not write.
- The withheld rebuild drops a slug from category rosters only, not from the organization
  roster. Stage and gaps are computed per category from the category rows, so that is the whole
  input to what is diffed; the org roster feeds the payload's organization block, which is not
  read here. The new roster-integrity assertion proves the mutate-and-restore comes back clean,
  but it does not prove the organization side is irrelevant — I reasoned that from
  `_stage_and_gaps`'s inputs rather than measuring it.
- The ADR quotes Carl's ruling from the job brief, which is where I have it. I did not have the
  #604 thread in this clone to check the quote's surrounding context, only its content.
- The survey is unscheduled and ungated by design, so nothing will tell anyone when its numbers
  move. That is the intended cost of not making it a gate, and the ADR says so, but it does mean
  the "measured 2026-09-20" figures in the ADR will drift and only a reader running the module
  will notice.

## Response to round N-1

|---|---|---|---|---|
| n/a — round 1; no reviewer round preceded it. The a3 build this started from never reached a reviewer (its run died in harness setup), so its three defects above were found by re-reading rather than raised. | — | — | — | `c840ff2e` |

Issue #355, opened 2026-08-23
observations.source_runs: live platform run-manifest emitter + authoritative row-to-run binding


ccerv1:
## Context

Phase 2 populates `observations.source_runs` from **real** control-plane run metadata (never inferred from surviving rows). A read-only binding check (2026-08-23) established what is and isn't available today:

- **Run execution status is authoritative and available** via the platform `runs` API: `id`, `triggerType`/`runType`, `status`, `queuedAt`/`startedAt`/`finishedAt`, `lastHeartbeatAt`, `steps`.
- **Run→model/materialization binding is authoritative** via `runs → steps → materializations` (materialization `id`, `runId`, `tableId` = the model id, `datasetId`, `createdAt`). A run can be bound to the exact source model it materialized, no timestamp inference.
- **Expected scope is NOT exposed** anywhere on a run (`metadata` is empty on success) → recorded `expected_scope: unknown`, `scope_status: unknown`.
- **Row→run binding is NOT available**: `signal_github.repo_state` / `signal_huggingface.hub_state` rows carry only `fetched_at`, no run/load id.

The interim implementation (Review Unit 1) is a **repo-side snapshot** of the `runs` API (`authority: repo`, staged): real `execution_status`, `scope_status: unknown`, run→model binding via materializations. It is a snapshot, not a live contract, and carries no row-level lineage.

## An open feasibility question, NOT yet proven

The control-plane API knowing a run id does **not** mean a Python UDM can access its own run/materialization id at execution time. Before Part B may add a `source_run_id` column to the `artifact_state` outputs, prove ONE of:

1. **The UDM runtime exposes the current run/materialization id to model code** (so the model writes it into each row); or
2. **The current table version can be atomically and unambiguously bound to its materialization id** (so a row's version resolves to exactly one materialization).

If neither holds, the row→run column is not achievable in the fetcher output and still requires the live platform emitter below. **Do not bridge the gap with timestamps.**

## Done when

- [ ] **Runtime feasibility proven** — (1) or (2) above demonstrated read-only, or explicitly ruled out.
- [ ] **Row→run lineage** — observation rows carry their producing `run_id` (added to the `artifact_state` outputs in Part B) *only if* feasibility is proven; otherwise via the emitter below.
- [ ] **Live emitter** — the platform writes a run-manifest row into `observations.source_runs` as each run completes (the §4.3 "required from day one" run contract), replacing the repo-side snapshot.
[...]

PR #635, opened 2026-09-20
Schedule the adoption reconciliation, and date adoption from the run that confirmed it

ccerv1:
# adoption-reconciliation-weekly — reviewer-unavailable

**Goal.** The comparison this needs **already exists**. `evaluation.adoption_reconciliation` compares what

## Success criteria
- [ ] The reconciliation runs on the weekly cadence, on the same schedule discipline as the signal tables. An observed SCHEDULED run is the evidence it works, not a configured cron
- [ ] Its disagreements are emitted as the tier-change queue: measured level != recorded level, one line per product, naming both values and the route
- [ ] **Adoption's `last_verified` derives** from the observation snapshot behind the last run that AGREED for that product. Not the execution date. Not `accessed`
- [ ] **A disagreement leaves the date where it was** and raises the product instead. A test asserts this: a product whose measured level differs must not have its date advanced
- [ ] `reverify` no longer touches adoption, and a test asserts it cannot
- [ ] The 354 counts-endpoint citations stop being evidence of currency. Either they go, or the PR argues why a hashed snapshot of a daily-changing number is worth keeping
- [ ] **No adoption score changes.** `level` and `reach` identical for all 763 products before and after, or every difference named. This job changes what dates a score, not what it is
- [ ] `docs/reference/evidence-and-freshness.md` states the derived rule. Reference document: what is true, no issue numbers, no chronology
- [ ] `uv run python -m build.validate` reports 0 errors; `check_payload` ok
- [ ] `uv run pytest -q -n auto -m "not serial and not network"` passes, count quoted

## Outcome
`reviewer-unavailable` after 1 round(s). Branch `loop/adoption-reconciliation-weekly-2026-09-20-a1`.

## Final verdict
```json
{
 "verdict": "fail",
 "blocking": [
  {
   "where": "build/adoption_freshness.py:94",
   "issue": "A zero delta becomes AGREED even when reconciliation reports source_unavailable and unknown freshness. The existing contract explicitly forbids treating unbound observations as agreement. Require valid run-bound evidence before dating an axis; retain diagnostic queue output meanwhile.",
   "failure_prevented": "Publishing the 296 advanced dates as confirmations without an attributable agreeing source run."
  },
  {
   "where": "build/adoption_freshness.py:284",
   "issue": "The gate checks only the snapshot-wide date window, not the product's agreeing measurement. In memory, changing accelerate's level, route to nonexistent.route, and date from August 20 to August 24 still returned no problems. Persist and validate product-specific agreement evidence.",
   "failure_prevented": "Accepting an unsupported date or retaining derived confirmation after the recorded score no longer agrees."
  },
  {
   "where": ".github/workflows/adoption-reconciliation.yml:21",
   "issue": "Only a configured cron exists; the builder confirms neither a scheduled execution nor the live path has been observed. Supply a successful schedule-triggered run showing the live reconciliation and emitted queue.",
   "failure_prevented": "Declaring weekly automation complete when production execution remains unproven."
  }
 ],
 "non_blocking": [],
 "settled_now": [
  "Using the existing reconciliation resolver preserves the required mechanism.",
  "The oldest contributing observation date and exclusion of cross-instrument matches are appropriate.",
  "All 763 products retain identical adoption level and reach. Direct Python validation reported 0 errors; check_payload passed.",
  "The builder reports 2284 tests passed and 2 skipped; independent uv execution was blocked by the read-only filesystem."
 ],
 "escalation": null
}
```


ccerv1:
## What I did

Put the existing reconciliation on a weekly cadence, wired both of its outcomes to something, and
made adoption's date come from the observation the route measured rather than from a person.

Four commits on `loop/adoption-reconciliation-weekly-2026-09-20-a1`:

- `0e16f6e2` feat(freshness): the mechanism
- `a820b736` chore(freshness): the first run's output (296 score files + the snapshot ledger)
- `459f04fa` test(freshness): the two date guards that restated the old rule as data
- `7a03862d` docs(operations): how to tell the weekly run actually fired

**Proposed PR title:** `Date adoption from the weekly reconciliation, and queue the disagreements`

- files touched:
  - `build/adoption_freshness.py` **(new, 379 lines)** — the derivation and the queue. `verdict()`
    reads one reconciliation row (`delta == 0` is an agreement on the same instrument, non-zero a
    tier change, differing levels with no delta a route disagreement); `queue()` / `render_queue()`
    emit the disagreements one line per product naming both levels and the route; `plan()` is
    advance-only and never emits a non-agreement; `apply()` re-checks every change before writing
    and raises on a disagreement, on a date that is not the observation date, and on a backwards
    move; `derivation_problems()` / `supports_date()` are the one predicate both date gates use.
  - `build/observation_snapshot.py` — `observed_window()`, `snapshot_record()`, `load_ledger()`,
    `record_snapshot()`, and the window added to `resolve_baseline()`. This is plan step 3: the
    snapshot id is a content hash and resolved to no date at all.
  - `sources/snapshots/observation_snapshots.yaml` **(new)** — id → observed window, so a derived
    date can be audited offline. The baseline snapshot `9bd4d93a6fc6` observed 2026-08-16 .. 08-24.
  - `build/check_verification.py` — the invariant accepts a derivation in place of a citation read
    since, for adoption only, and fails on a malformed one, an unrecorded snapshot, a date outside
    the window, or a date that is not the observation's.
  - `build/check_redate.py` — same, for the "moved forward" gate: a derived move needs the
    measurement, not an `accessed` on the day.
  - `build/reverify.py` — refusal messages and docstring now name adoption's actual owner. The
    three refusals (flag, planner, writer) were already there.
  - `build/serialize.py` — `derived_from` stays in the repo and out of the payload.
  - `docs/schemas/score.schema.json` — `adoption.derived_from` (`observation_snapshot_id`,
    `route_id`, `measurement_as_of`), `additionalProperties: false`.
  - `docs/reference/evidence-and-freshness.md` — new section "How an adoption date is earned" plus
    edits to "Who may write `last_verified`", the machine-re-verification paragraph and the
    per-axis automation table. Reference register: no issue numbers, no chronology.
  - `docs/reference/adoption.md`, `docs/operations/deploy-models.md` (schedule row),
    `docs/operations/deploy-evaluation.md` (§5, the weekly run), `warehouse/assets.yaml` (the
    reconciliation now has a consumer; the retirement finding is narrowed to the deployed table).
  - `.github/workflows/adoption-reconciliation.yml` **(new)** — Monday 05:00 UTC, `--live --apply`,
    gates, queue to the run summary whether or not a date moved, one PR per run.
  - `tests/test_adoption_freshness.py` **(new, 26 tests)**, `tests/test_reverify.py` (+1),
    `tests/test_check_verification.py`, `tests/test_check_adoption.py`, `tests/test_schedule_doc.py`.
  - `sources/scores/*.yaml` — 296 files, `adoption.last_verified` + `derived_from`, nothing else.

  - `uv run pytest -q -n auto -m "not serial and not network"` → **2284 passed, 2 skipped in
    348.83s** (baseline before the change: 2257 passed, 1 skipped). The extra skip is
    `tests/test_goldens.py` — `sources/` differs from the golden's fingerprint, which is the
    expected state for a corpus change and is gated by `check_corpus_diff` on a PR instead.
  - `uv run pytest -q -m serial` → 1 passed, 2286 deselected.
  - `uv run python -m build.validate` → **0 error(s), 2 warning(s)** (both warnings pre-existing
    `model_families` pattern overlaps).
  - `uv run python -m build.serialize --date 2026-09-20 && uv run python -m build.check_payload` →
    `check_payload: 763 products, 421 organizations, ok`.
  - `build.check_verification --verbose` (4/4 OK), `build.check_adoption --strict` (OK),
    `build.check_instrument` (OK), `build.check_citations` (OK), `build.check_redate` (296 moves,
    all supported), `build.check_retirement` (0 retired, all routed),
    `build.check_corpus_diff --base dacfc679` (no stage moves, no untouched-product row changes).

**Before / after.**

| | before | after |
|---|---|---|
| corpus median axis age | 38d | **33d** |
| oldest axis | 43d | 43d |
| newest axis | 3d | 3d |
| adoption `level` / `reach` / `signal_type` / `banded_quantity` / `confidence`, 763 products | — | **0 differences** |
| serialized payload, adoption block, 763 products | — | **byte-identical** |
| payload fields that changed at all | — | `adoption.last_verified`, `freshness.date`, `freshness.latest_axis_confirmation` |

The run: 763 rows, 297 agreements on the same instrument, 11 tier changes, 56 route
disagreements, 399 not compared. 296 dates earned; `gsm8k` declined because a person confirmed it
on 2026-09-01, after the snapshot was observed. A second run over the same snapshot plans **0**
changes, so the mechanism is idempotent.

## Why

- decisions:
  - **The date is `measurement_as_of`, the OLDEST contributing observation** — not the run's
    execution date, not `accessed`. An aggregate is only as current as its stalest part, and
    stamping a run with today would claim a currency nobody has.
  - **The support record lives on the axis, not in a side ledger.** `derived_from` travels with
    the date it supports. Split across files, a hand edit could move the date and leave the
    support pointing at something else, and the gate would read a support that does not belong to
    the claim.
  - **The snapshot ledger exists because a content hash carries no date.** Plan step 3 asked
    exactly this, and the answer was that it resolved to nothing. Without it a score file records
    an opaque 64-hex string and no reader can say whether it names last week's observations or
    March's. The ledger also gives the gate a real check: a date outside the snapshot's observed
    window fails.
  - **Agreement is `delta == 0`, which means same instrument.** The reconciliation already refuses
    to subtract across instruments; 11 cross-instrument rows have equal levels and none of them
    dates anything, because equal numbers on two different instruments are a coincidence.
  - **The queue carries both kinds of disagreement, labelled.** 11 tier changes (re-band or
    dispute the figure) and 56 route disagreements (the recorded instrument and the applicable
    route measure different quantities). Same line shape, one marker, because they are different
    repairs.
  - **The invariant had to learn a second support.** All 296 derived dates would have failed the
    `accessed` floor — measured: 296 of 297 agreements have no adoption citation read on or after
    their observation date. That is not a hole in the gate, it is the finding the job names: the
    citation was never what made the figure current.
  - **The 354 counts-endpoint citations stay, as provenance.** My census by URL shape
    (`huggingface.co/api/`, `pypistats.org/api/`, `api.github.com/`, `api.npmjs.org/`) is 663
    corpus-wide adoption citations, 483 of them on a purely numeric endpoint across 357 products —
    so I could not reproduce 354 exactly and have reported what I measured instead. **314 of them
    sit on the 296 now-derived axes, and every one is older than the date it carries**: none of
    them supports a date any more, and the gate reads the observation instead. They are kept
    because a digest over a counts body is the only proof the fetch was real, which is what the
    sampled re-fetch checks and what keeps a fabricated citation catchable — and because where no
    route can measure, a person's reading is the only support there is. Deleting them would remove
    the fabrication check and leave the undated-by-measurement axes unsupported.
  - **`derived_from` is stripped from the payload.** It answers an audit question, not a reader's,
    and the payload's transport is the map's tightest constraint.
  - **Two commits for mechanism and corpus**, so a reviewer can read the 296-file diff separately
    from the 1,215 lines of mechanism.

- alternatives rejected:
  - *Making the reconciliation emit `agree` as a status.* The status vocabulary is about run
    binding (#355) and is deliberately `source_unavailable` everywhere until that lands. The
    freshness derivation compares `measured_level` against `recorded_level` directly and never
    reads `status`, so this job does not re-litigate that ruling — noted in `assets.yaml`.
  - *Deriving the date in the gate by recomputing the reconciliation.* Deterministic today over
    the committed baseline, and broken the moment the baseline is replaced: a recorded date would
    stop being reproducible. The recorded derivation is durable.
  - *Advancing every agreement unconditionally.* One product (`gsm8k`) has a human confirmation
    newer than the snapshot. Overwriting it would be a tool's arithmetic beating a person's
    observation, which the guide already forbids.
  - *Committing the regenerated payload.* `generated-files-guard` forbids it; the bot regenerates
    on merge. I ran `serialize` + `check_payload` and reverted `build/notebook_data.json`.

## Risks / unsure about

- **The one criterion I cannot satisfy from here: an observed SCHEDULED run.** This clone has no
  remotes and no network, so the workflow has never executed. What I did instead: put the cron
  under the repo's existing schedule discipline (`tests/test_schedule_doc.py` now holds
  `adoption-reconciliation` to the doc table, as it does the other eight gates), and write into
  `docs/operations/deploy-evaluation.md` §5 that a configured cron is not an observed run, that a
  `schedule` entry in the run list is the evidence, and the point-the-cron-a-few-minutes-out trick
  this repo already uses to settle it in minutes. Someone with the repo has to look, once.
- **The `--live` path is unexercised here.** The weekly job reads the deployed
  `observations.product_adoption_current` through `load_current_observations()`. `assets.yaml`
  carries an older note that `--live` is broken on an uncoerced VARCHAR `observed_at`;
  `_coerce_timestamp` in `build/adoption_measurements.py` now handles exactly that, so I believe it
  is fixed, but I could not run it (no key, no network). The first dispatch will say. Everything in
  this PR was run over the committed baseline snapshot, which is the same code path minus the read.
- **296 re-dated axes is a large one-time move**, and it is a real claim: those bands were
  re-measured from observations taken 2026-08-16..24 and came back identical. The median improves
  38d → 33d rather than dramatically, because the observation snapshot itself is a month old. Once
  the weekly job runs against live observations the dates will track the run.
- **A route measuring the wrong thing now dates an axis silently.** If a declared artifact is not
  actually the product's (the `not_primary_channel` / withdrawn-declaration class of bug), the
  route will keep agreeing with a band derived from the same bad artifact and keep re-dating it.
  The reconciliation cannot see that, and neither can this. It is the pre-existing hazard of
  declarations, not a new one, but the date now moves on it.
- **`freshness.latest_axis_confirmation` appears in the payload** for 278 products that did not
  carry it before. It is an existing `freshness_payload` field that surfaces when axis dates
  diverge; nothing I wrote emits it, and no gate objects. Worth a glance from someone who knows the
  front end.

## Response to round N-1

|---|---|---|---|---|
| (round 1 — no prior review) | | | | |

ccerv1:
## What I did

Made the mechanism refuse to date an axis it cannot attribute to a run, gave the derivation a
second record to be checked against, and unwound the 296 dates round 1 published on evidence that
does not meet that bar.

Two commits added on `loop/adoption-reconciliation-weekly-2026-09-20-a1` (nothing rewritten):

- `a05a4242` chore(freshness): unwind the 296 derived dates until a bound run earns them
- `84cc2c6c` feat(freshness): require a bound run and a second record before a date is earned

**Proposed PR title:** `Schedule the adoption reconciliation, and let only a run-bound measurement date an axis`

- files touched (this round):
  - `build/adoption_freshness.py` — `binding_problems()` is the new precondition: a read is usable
    only when the bracket named one materialization at both ends and the binding carries
    `run_id` / `materialization_id` / `model` / `read_at`. `plan(rows, binding, root)` declines
    every match when it is not, one line per product, and `apply()` refuses a change with no
    `source_run_id` and re-reads the file to refuse one whose recorded level is not the level the
    run measured. `AGREED` is gone: a zero delta is now `BAND_MATCH`, because the comparison
    cannot say more than that. `derivation()` records `source_run_id` and `measured_level` beside
    the snapshot, route and observation date; `agreements()` is the ledger's half of it;
    `known_route_ids()` compiles the route ids a gate checks against; `_already_supported()` keeps
    a weekly re-run from rewriting 300 files to swap one run id. `derivation_problems()` now also
    fails a level the score has left, a route the tables do not compile, a product the ledger
    records no measurement for, and any field on which the two records disagree. The unused
    `supports_date()` helper is deleted — round 1 described it as the shared predicate and nothing
    called it.
  - `build/observation_snapshot.py` — the ledger entry carries `agreements`: per product, the run,
    route, level and observation date that dated it. `record_snapshot(record, agreements=...)`
    merges them (a product this run did not date keeps the record of the run that did) while the
    content half stays immutable.
  - `build/check_verification.py` — passes the compiled route ids into the derivation check, and
    compiles them only when something in the corpus derives.
  - `build/check_redate.py` — a derived move is put to the full derivation check against the
    ledger, not only to the "date equals the measurement" shape check.
  - `docs/schemas/score.schema.json` — `derived_from` gains `source_run_id` and `measured_level`,
    both required.
  - `docs/reference/evidence-and-freshness.md` — "A match is not an agreement until the run is
    attributable" and "The measurement is recorded twice", in the reference register.
  - `docs/operations/deploy-evaluation.md` §5 — only a live bound read can date anything; what
    `source run  NONE` means and when a week of it is a problem; the Sunday 03:30 UTC sweep this
    job runs behind.
  - `docs/operations/deploy-models.md` — the schedule table gains the `observations` dataset row.
  - `.github/workflows/adoption-reconciliation.yml` — the summary now carries the binding line as
    well as the queue; the comments say why the job reads live and never the baseline.
  - `warehouse/assets.yaml` — the consumer note says the derivation requires run-bound evidence of
    its own rather than reading `status`.
  - `tests/test_adoption_freshness.py` — 45 tests (was 26). New: the three ways a read fails to
    bind, an unbound read dating nothing while still queueing, `bound_read` driven through a fake
    control plane in both the stable and the mid-bracket-refresh case, the writer refusing a
    run-less date and a re-banded score, and the reviewer's own repro as
    `test_the_three_hand_edits_a_self_describing_record_could_not_catch`.
  - `sources/scores/*.yaml`, `sources/snapshots/observation_snapshots.yaml` — reverted. `git diff
    dacfc679 HEAD -- sources/` is **0 lines**.

- commands run (all foreground, in the build clone):
  - `uv run pytest -q -n auto -m "not serial and not network"` → **2304 passed, 1 skipped in
    331.11s**. (Round 1's run had 2 skips; the second was `test_goldens.py` skipping on a changed
    corpus, and the corpus is unchanged again.)
  - `uv run pytest -q -m serial` → 1 passed, 2305 deselected.
  - `uv run python -m build.validate` → **0 error(s), 2 warning(s)** (both pre-existing
    `model_families` pattern overlaps).
  - `uv run python -m build.serialize --date 2026-09-20` then `uv run python -m build.check_payload`
    → `check_payload: 763 products, 421 organizations, ok`. Regenerated payload reverted, per
    `generated-files-guard`.
  - `uv run python -m build.check_verification --verbose` → 4/4 OK over 2289 dated axes;
    `check_adoption --strict` → OK; `check_redate --base dacfc679` → 0 re-datings;
    `check_freshness` → **2289 axes | median 38d | oldest 43d | newest 3d**, the merge base's own
    numbers.
  - `uv run python -m build.adoption_freshness --apply --allow-dirty` (the baseline path) →
    `band_match 297, tier_change 11, route_disagreement 56, not_compared 399`,
    `source run NONE — the frozen baseline records no source run…`, `dates earned 0`, full queue
    printed, and `git status` clean afterwards.
  - Platform reads through the OSO MCP connection (read-only): `observations` dataset cron
    `30 3 * * 0` UTC, `lastRunAt 2026-09-20T03:30:13Z`; its run list shows run
    `980a87f1-3274-408e-928e-70eee35dd8a2`, `triggerType: SCHEDULED`, `status: SUCCESS`,
    2026-09-20T03:30:16Z → 03:30:26Z. `observations.product_adoption_current` revision 3 still
    projects `CAST(NULL AS VARCHAR) AS source_run_id -- blocked on #355`.

## Why

- decisions:
  - **The reviewer is right about §4.3, and the consequence is the corpus, not just the code.**
    Requiring run-bound evidence and leaving 296 dates standing that were earned without it would
    have been a gate that exempts its own first run. So the dates go back and the mechanism ships
    with no corpus effect at all; the first scheduled run against the deployed table earns them.
  - **The evidence is a bracketed read, and `build/read_binding.py` already existed for exactly
    this.** It reads the newest materialization of `observations.product_adoption_current` before
    and after the row read; one materialization at both ends means no run completed across the
    read, so the rows were served by that materialization and are bound to its `run_id`.
  - **This does not claim to close #355, and the status vocabulary is untouched.** #355 is
    row-to-run binding of each observation to the FETCHER run that measured it. What a bracket
    gives is read-grain attribution: which model run served these rows. I say so in the module, in
    the reference doc and in `assets.yaml`, and `adoption_reconciliation.py` is unchanged — every
    measured row is still `source_unavailable`.
  - **Read-grain attribution has a hole, and the date is what fills it.** A failed collector can
    leave the current table serving last week's figures while the model run that read it succeeds.
    The date written is the oldest contributing observation's own date, so that case dates the axis
    to last week rather than to today. Staleness is recorded, not laundered. This is the one place
    where I am asking the reviewer to accept a bound read as sufficient, and it is sufficient
    because of what the date is, not because the binding is stronger than it is.
  - **A self-describing record cannot be audited.** Round 1's `derived_from` was checked against
    the snapshot window and against itself, which is why the reviewer could move a date four days,
    rename the route and re-band the level without a complaint. The same measurement is now
    written into the ledger under the product's slug and the gate requires the pair to agree, so
    each of those three edits fails — plus the level check catches an honest re-banding, which is
    the case that actually happens.
  - **`measured_level` is part of the support, not decoration.** The run confirmed a band. Once a
    curator moves the level, that confirmation is about a band the score has left, so the date must
    be earned again. The writer enforces it at write time too, re-reading the file, because a plan
    built minutes earlier can be stale by the time it writes.
  - **The run id is excluded from "has this already been dated".** Otherwise a weekly run over
    unchanged observations rewrites ~300 files to swap one opaque id for another. What identifies
    the confirmation is the measurement: snapshot, route, level, date.
  - **`--apply` over the baseline is not an error.** It prints what it would have dated and writes
    nothing. The comparison is the useful half without a run behind it, and the queue is what the
    job is for on a week that dates nothing.

- alternatives rejected:
  - *Reading `status == "agree"` from the reconciliation.* That vocabulary is reserved for the
    #355 binding and is `source_unavailable` everywhere by design. Flipping it here would claim
    row-to-run binding this does not have.
  - *Keeping the 296 dates and adding the binding requirement going forward.* It would leave dates
    in the corpus that today's gate would reject, and `check_verification` would have had to be
    taught an exemption for the run that wrote them.
  - *Requiring the collector runs themselves to be attested from `observations.source_runs`.* The
    receipt is a point-in-time snapshot of platform run history with `expected_scope: unknown` by
    construction, so a SUCCESS there cannot say the collection was complete. Making it a
    precondition would look like proof and not be any.
  - *Putting the per-product evidence in a new `warehouse/audits/` receipt.* The ledger already
    exists, already resolves the snapshot, and is already what the gate loads. A second file would
    be a third record to keep in step.
  - *Inferring a run id for the baseline from its capture timestamp.* That is the timestamp
    inference `read_binding` refuses in its own docstring.

## Risks / unsure about

- **The live path has still never executed.** `bound_read` is exercised here only through a fake
  control plane; the pyoso read and the real GraphQL response shape are not. The first dispatch is
  what settles that, and the failure mode is loud (`BindingLookupError`, or an `unstable` bracket
  that dates nothing) rather than silent.
- **The bracket could be chronically unstable.** If the Monday 05:00 slot regularly overlaps a
  materialization, every week reports `source run NONE` and no date is ever earned. I wrote the
  rule for reading that into §5 — a week is fine, a run of weeks means move the slot, not relax the
  rule — but nobody has watched it yet.
- **Nothing in this PR moves a date, so the freshness numbers do not improve.** That is the honest
  state: the corpus is exactly the merge base's, medians 38d / 43d / 3d. The 296 dates come back on
  the first bound run, and if the reviewer wants to see them in a diff, that diff has to come from
  a run with the key.
- **The ledger will grow.** One entry per snapshot, with an agreement per dated product. At ~300
  products a week that is tens of KB per new snapshot. Nothing prunes entries no axis references
  any more; worth a follow-up once there is more than one.
- **`freshness.latest_axis_confirmation` no longer appears** for the 278 products round 1 surfaced
  it on, because no axis dates diverge any more. Same reason as everything else here.
- **A route measuring the wrong artifact still dates silently.** If a declaration names an artifact
  that is not really the product's, the route keeps agreeing with a band derived from the same bad
  artifact. Unchanged from round 1, unchanged by the binding, and not something this mechanism can
  see.

## Response to round N-1

|---|---|---|---|---|
| `build/adoption_freshness.py:94` — a zero delta becomes AGREED although reconciliation reports `source_unavailable` and unknown freshness; require run-bound evidence, keep the queue meanwhile | ✅ | | | A zero delta is now `BAND_MATCH`, not an agreement. `plan()` takes a binding and `binding_problems()` gates it: only a bracketed read of `observations.product_adoption_current` that named one materialization at both ends, carrying `run_id`, `materialization_id`, `model` and `read_at`, may date anything. The baseline and an `unstable` bracket date nothing and still emit the queue — verified by running `--apply` over the baseline: `dates earned 0`, full queue printed, `git status` clean. `apply()` refuses a change with no `source_run_id`. The 296 dates round 1 published are reverted in `a05a4242`; `git diff dacfc679 HEAD -- sources/` is 0 lines. Tests: `test_an_unbound_read_dates_nothing`, `test_an_unbound_read_still_emits_the_queue`, `test_a_refresh_landing_mid_read_dates_nothing`, `test_the_writer_refuses_a_date_with_no_source_run`, `test_a_read_served_by_one_materialization_is_bound_to_its_run`. Reconciliation's `status` is untouched and #355 is not claimed closed. |
| `build/adoption_freshness.py:284` — the gate checks only the snapshot-wide window; changing accelerate's level, its route to `nonexistent.route` and its date Aug 20 → Aug 24 returned no problems. Persist and validate product-specific evidence | ✅ | | | The measurement is now written twice: `derived_from` on the axis (snapshot, run, route, measured level, observation date) and the same record in the snapshot ledger under the product's slug. `derivation_problems()` fails a level the score has since left, a route the compiled tables do not carry, a product the ledger records no measurement for, and any field where the two records disagree; `check_verification` supplies the compiled route ids and `check_redate` runs the same check on a derived move. Your exact repro is now `test_the_three_hand_edits_a_self_describing_record_could_not_catch`, which asserts all three edits are caught, alongside `test_a_re_banded_score_loses_its_derived_confirmation`, `test_a_route_the_tables_do_not_compile_fails`, `test_a_date_the_ledger_does_not_record_fails`, `test_a_product_the_ledger_records_no_measurement_for_fails`. `apply()` re-reads the score file and refuses to date a band it did not measure. |
| `.github/workflows/adoption-reconciliation.yml:21` — only a configured cron exists; supply a schedule-triggered run showing the live reconciliation and the emitted queue | | | ⚠️ | Cannot be produced from here and the attempt would be outward-facing: this clone has no remote, no `OSO_API_KEY`, and no GitHub Actions to fire — producing the run means pushing the branch to `currentai-org/os-ai-map`, giving it the secrets and dispatching it, which is a change to the live repository rather than to this branch. **What I did instead of claiming it:** made an unproven schedule harmless. Nothing in this PR dates an axis, and nothing can until a live bound read succeeds, so "production execution unproven" no longer means "296 unattributed dates are already published". Related evidence I could get read-only: the chain this job runs behind IS observed on schedule — `observations` dataset cron `30 3 * * 0` UTC, run `980a87f1-3274-408e-928e-70eee35dd8a2`, `triggerType: SCHEDULED`, `status: SUCCESS`, 2026-09-20T03:30Z. `docs/operations/deploy-evaluation.md` §5 says what to look at on the first fire: the trigger on the run entry, the binding line in the summary (`source run <id>` vs `source run NONE`), and the queue, which is written whether or not a date moved. Someone with the repo has to press it once. |

ccerv1:
Superseded by #637, which carries the same job's work plus the review fixes: a staleness check on the binding (a SUCCESS/SCHEDULED run stays SUCCESS/SCHEDULED forever, so age was the missing test) and pipefail on the reconciliation step. Closing rather than merging — two PRs for one job, and #637 is the reviewed one.

Issue #376, opened 2026-08-26
observations dataset has no cron — blocking for any observations cutover


ccerv1:
## Problem

The `observations` dataset has **no configured cron**. Any model that becomes *canonical* in `observations` — i.e. a table a deployed consumer reads as its source of truth — will not refresh on a schedule, so a consumer repointed onto it goes stale.

## How it surfaced

During the Phase 5 pilot (#375), `observations.product_adoption_current` was repointed to read `observations.paper_citations`, whose dataset has no cron. The old `signal_semanticscholar` Sunday cron (`0 1 * * 0` UTC) refreshes only the retained compatibility table — it cannot keep the new canonical table current.

Per Phase 1's schedule-normalization discipline, a configured cron is not itself proof that a scheduled run fired (§13 requires an **observed `SCHEDULED` run** as evidence). A cutover onto an unscheduled canonical table is therefore not sound.

## Blocking condition

Before any consumer is repointed onto a canonical `observations` table:

1. the `observations` dataset must have a UTC cron configured; **and**
2. an observed `SCHEDULED` run must be recorded (not merely configured).

**This is a blocking gate for observations cutovers, not a non-blocking follow-up.**

## Note

#375 itself is being rolled back on separate architectural grounds (data-architecture.md §4.3: source-specific ingestion remains in `signal_*`; a `kind: observations` table in a `signal_*` namespace is permanently legitimate), which removes this specific instance. This issue tracks the general scheduling contract for the `observations` namespace so it is not lost.

ccerv1:
**Condition 1 is met again. Condition 2 resets to 2026-09-20, and the evidence that would have satisfied it was destroyed today.**

The `observations` dataset carries a UTC cron: `30 3 * * 0`, Sunday 03:30, placed after the 03:00 GitHub leg it reads and before entities at 04:00. `nextRunAt` is 2026-09-20T03:30Z.

The awkward part is worth stating plainly rather than letting it look like progress. This is a **new** `observations` dataset. The old one could no longer be released after `signal_github.artifact_state` turned schema-deterministic, so it was recreated and the superseded dataset deleted. That old dataset had carried a cron since 2026-09-03 and had two runs behind it, 2026-09-06 and 2026-09-13, which is very likely the observed `SCHEDULED` run this issue asks for. Deleting it took that run history with it.

So the state is:

[...]

PR #637, opened 2026-09-20
Schedule the adoption reconciliation, and date adoption from the run that confirmed it

ccerv1:
# adoption-reconciliation-weekly — reviewer-unavailable

**Goal.** The comparison this needs **already exists**. `evaluation.adoption_reconciliation` compares what

## Success criteria
- [ ] The reconciliation runs on the weekly cadence, on the same schedule discipline as the signal tables. An observed SCHEDULED run is the evidence it works, not a configured cron
- [ ] Its disagreements are emitted as the tier-change queue: measured level != recorded level, one line per product, naming both values and the route
- [ ] **Adoption's `last_verified` derives** from the observation snapshot behind the last run that AGREED for that product. Not the execution date. Not `accessed`
- [ ] **A disagreement leaves the date where it was** and raises the product instead. A test asserts this: a product whose measured level differs must not have its date advanced
- [ ] `reverify` no longer touches adoption, and a test asserts it cannot
- [ ] The 354 counts-endpoint citations stop being evidence of currency. Either they go, or the PR argues why a hashed snapshot of a daily-changing number is worth keeping
- [ ] **No adoption score changes.** `level` and `reach` identical for all 763 products before and after, or every difference named. This job changes what dates a score, not what it is
- [ ] `docs/reference/evidence-and-freshness.md` states the derived rule. Reference document: what is true, no issue numbers, no chronology
- [ ] `uv run python -m build.validate` reports 0 errors; `check_payload` ok
- [ ] `uv run pytest -q -n auto -m "not serial and not network"` passes, count quoted

## Outcome
`reviewer-unavailable` after 0 round(s). Branch `loop/adoption-reconciliation-weekly-2026-09-20-a2`.


ccerv1:
## What I did

Started from a1 (PR #635), whose round 2 had already answered the first two findings before the
reviewer ran out of quota mid-read. I verified those answers rather than assuming them, then did
the thing a1 could not: **made the trigger of the platform run part of the binding**, so a date
cannot be earned except from a run the platform recorded as `SCHEDULED` and successful — and then
ran it, live, against the deployed table, binding 460 dates to the `observations` dataset's Sunday
03:30 UTC sweep.

**Proposed PR title:** `Put adoption's date on the weekly cadence, and let only a scheduled run earn one`

Three commits on `loop/adoption-reconciliation-weekly-2026-09-20-a2`, nothing rewritten:

- `6d4affa6` merge of `loop/…-a1` (its six commits are preserved intact as the base)
- `7f9f15ed` feat(freshness): only a scheduled, successful run may date an adoption axis
- `8015020d` chore(freshness): date 460 adoption axes from the scheduled run that re-measured them

- files touched (this round, on top of a1):
  - `build/read_binding.py` — a bound bracket now also looks the run up. `RUN_QUERY` /
    `run_record()` read `triggerType`, `runType`, `status` and `startedAt` from the control plane;
    `bound_read` puts `run_trigger_type` / `run_status` / `run_started_at` on the binding. The
    graphql-and-token plumbing moved into `_control_plane()` so the two queries share it. An id the
    control plane cannot resolve raises `BindingLookupError` rather than degrading to "trigger
    unknown" — the id came out of a materialization the bracket already proved, so a control plane
    that will not answer for it is contradicting itself.
  - `build/adoption_freshness.py` — `BINDING_KEYS` gains the two new fields, and
    `binding_problems()` gains the cadence rule: a run that is not `SUCCESS` dates nothing, and a
    run that is not `SCHEDULED` dates nothing because it is a person. `run_receipt()` is the
    ledger's record of the dating run. `derivation_problems()` asks the ledger, not the control
    plane, whether the run named by a stored `derived_from` was the scheduled one — three distinct
    failures: nothing recorded about the run, a run recorded `MANUAL`, a run recorded unsuccessful.
    The run line in the CLI output now prints `SCHEDULED/SUCCESS started …`.
  - `build/observation_snapshot.py` — the ledger entry carries `source_runs`, one record per run
    (`trigger_type`, `status`, `started_at`), merging like `agreements` so a later run cannot erase
    the run an older agreement still points at.
  - `.github/workflows/adoption-reconciliation.yml` — the header no longer lets its own cron stand
    for the cadence, and says why `workflow_dispatch` is safe. Also records a measured operational
    fact: GitHub starts this repo's Monday scheduled runs five to six hours after their cron, so the
    ordering that matters is against the Sunday platform sweep, not against the sibling workflows.
  - `docs/reference/evidence-and-freshness.md` — "The run has to be the scheduled one", in the
    reference register: what it rules out, and why a configured schedule is not evidence a schedule
    ran.
  - `docs/operations/deploy-evaluation.md` §5 — the enforced half separated from the observed half,
    and the PR-reading checklist extended to the `source_runs` entry.
  - `warehouse/assets.yaml` — the consumer note says the derivation also requires the run to be a
    successful SCHEDULED one.
  - `tests/test_adoption_freshness.py`, `tests/test_read_binding.py` — 50 tests (was 45). New: a
    MANUAL-materialized table dates nothing and still queues, a `RUNNING`/`FAILED` run dates
    nothing, a bound read carries the trigger the control plane reports (both ways), a stored date
    fails the gate on all three ledger states, and a second run does not erase the first run's
    receipt. Both control-plane fakes now answer the run query.
  - `sources/scores/*.yaml` (460 files), `sources/snapshots/observation_snapshots.yaml` — the first
    bound run's output.

- commands run (all foreground, in the build clone):
  - `uv run pytest -q -n auto -m "not serial and not network"` → **2308 passed, 2 skipped in
    346.21s**. Both skips are structural: `test_identity_eval.py:1149` needs `origin/main`, which a
    remoteless clone has not got, and `test_goldens.py:45` skips because `sources/` moved, which is
    what `check_corpus_diff` gates on a PR. `uv run pytest -q -m serial` → 1 passed, 2310 deselected.
  - `uv run python -m build.validate` → **0 error(s), 2 warning(s)** (both the pre-existing
    `model_families` pattern overlaps).
  - `uv run python -m build.serialize --date 2026-09-20` then `uv run python -m build.check_payload`
    → **`check_payload: 763 products, 421 organizations, ok`**. The regenerated payload was
    reverted; it is one of the bot-owned pair (`tests/test_preflight.py::test_the_generated_files_
    are_the_bot_owned_pair`).
  - `uv run python -m build.check_verification --verbose` → 4/4 OK over 2289 dated axes.
    `check_adoption --strict` → OK, 724 products with a band. `check_redate` → 458 re-datings,
    every one supported. `check_corpus_diff --base main` → **no stage moves, no tier changes, no
    untouched-product row changes** — an empty review sheet.
  - `uv run python -m build.check_freshness` → before **2289 axes | median 38d | oldest 43d | newest
    3d**; after **median 33d | oldest 43d | newest 0d**. The oldest number is unchanged but the
    axis under it is not: `finetuning_code`'s oldest was `trl.adoption` and is now
    `trl.capability`, both at 43d. This job moves adoption dates and nothing else, so the age gate
    is unchanged at the ceiling and five days better at the median.
  - `uv run python -m build.adoption_freshness --live --apply --queue …` with a real `OSO_API_KEY`
    → snapshot `36a826325844…`, observed 2026-09-08 .. 2026-09-20, 1111 rows; `source run
    980a87f1-3274-408e-928e-70eee35dd8a2 SCHEDULED/SUCCESS started 2026-09-20T03:30:16.737Z
    (materialization 84a794f8-57ca-46c0-8363-efdf7553d046)`; band_match 460, tier_change 56,
    route_disagreement 69, not_compared 178; dates earned 460, matches not dated 0; wrote 460 score
    files.
  - Two ad-hoc proofs, scripted and re-runnable, in the session scratchpad: the 763-product
    level/reach comparison, and the reviewer's own three hand-edits plus a fourth on the trigger.

## Why

- decisions:
  - **The cadence is enforced, not configured.** Finding 3 asks for an observed `SCHEDULED` run and
    cannot be satisfied by a YAML cron. The cron a1 added is still there, but it no longer carries
    any weight: `binding_problems()` refuses a materialization whose run was `MANUAL`, so if the
    weekly chain ever stops firing on its schedule, every match declines, the run prints why, and
    the queue is the week's whole output. A person refreshing the table by hand and re-running the
    job earns nothing. That is the #376 discipline applied where it can be checked rather than
    asserted, and it is the literal reading of the goal — adoption's freshness coming *from the run
    rather than from a person*.
  - **So the evidence is now in the corpus, not in a promise.** Every one of the 460 derived dates
    names run `980a87f1-3274-408e-928e-70eee35dd8a2`, and
    `sources/snapshots/observation_snapshots.yaml` records that run as `SCHEDULED` / `SUCCESS`,
    started 2026-09-20T03:30:16Z. That is the `observations` dataset's Sunday 03:30 UTC sweep — the
    same run issue #376 was closed on today. The weekly cadence the signal tables run on is not
    something this PR promises to join; it is the run the dates already rest on.
  - **The trigger goes in the ledger, not only in the binding.** The control plane will not answer
    for a run forever, and the gate has to be able to ask about a date months after it was written.
    One record per run, beside the agreements, merging the same way so an older agreement's run
    survives a newer run's write.
  - **The live path executed, which is the thing a1 listed as its largest open risk.** `bound_read`
    had only ever run against a fake control plane. It now has a real GraphQL response behind it,
    a real bracket over a real pyoso read, and a real run lookup.
  - **`workflow_dispatch` stays, and is now safe by construction.** A manual dispatch over a
    scheduled materialization earns exactly what the scheduled dispatch would. What is checked is
    the run that produced the data, not the hand that started the job.
  - **The counts-endpoint citations stay, as provenance, and I re-measured rather than inheriting
    a1's census.** Corpus-wide there are 1204 adoption citations; 665 sit on a purely numeric
    endpoint (`huggingface.co/api/`, `pypistats.org/api/`, `api.github.com/`, `api.npmjs.org/`,
    `crates.io/api/`, `api.semanticscholar.org/`) across 505 products. 519 of those are on an axis
    whose date is now derived, and **517 of the 519 were accessed earlier than the date the axis now
    carries** — so they demonstrably are not what makes any of those dates current. They are kept
    because a digest over a counts body is the only proof the fetch happened, which is what the
    sampled re-fetch checks and what keeps a fabricated citation catchable, and because on the 178
    products no route could measure at all, a person's reading is the only support there is. The figure in the
    goal is 354; I could not reproduce it under any shape list I tried and have reported what I
    measured.
  - **Two commits, mechanism and corpus.** A reviewer can read the rule without the 461-file diff.

- alternatives rejected:
  - *Folding the job into `reverify.yml` or `freshness.yml`, whose scheduled legs are already
    observed.* It would have produced a citable `schedule` run for the containing workflow, and it
    is the wrong trade: `reverify`'s scheduled runs are 1-for-2 over the last fortnight, so adoption
    would inherit another job's availability, and `freshness.yml` is a read-only gate with neither
    warehouse access nor PR permissions. Coupling two weekly jobs to borrow a green checkmark is
    worse engineering than making the checkmark unnecessary.
  - *Pushing the branch to `currentai-org/os-ai-map` and dispatching the workflow to produce the
    run.* Outward-facing, out of scope for a loop round, and it would still not be a `schedule`
    trigger — a workflow only fires on its cron from the default branch.
  - *Requiring the FETCHER runs to be scheduled too.* `observations.source_runs` records those, but
    with `expected_scope: unknown` by construction, so a SUCCESS there cannot say the collection was
    complete. Making it a precondition would look like proof and not be any. What is required here
    is the trigger of the materialization the read was served by, which is exactly what the bracket
    already proves.
  - *Leaving the trigger out of the ledger and looking it up at gate time.* It would make
    `check_verification` need a warehouse token and a network, and it would fail open the day the
    platform ages a run out of what it will answer for.
  - *A new gate that watches the workflow's own run history over the GitHub API.* A network
    dependency in a repo gate, red on merge by construction, to check the half that no longer
    decides whether a date is sound.

## Risks / unsure about

- **The workflow's own cron has still never fired, and cannot be made to from here.** What changed
  is that this no longer gates anything: an unfired cron now produces no dates rather than
  unattributed ones. Somebody with the repo still has to look at the first scheduled entry;
  `deploy-evaluation.md` §5 says what to read, and to expect it hours late.
- **Read-grain attribution, not row-grain.** The bound run is `product_adoption_current`'s own
  materialization, not the fetcher runs that measured the artifacts. #355 is not closed and the
  reconciliation's `source_unavailable` status is untouched. The hole this leaves — a failed
  collector whose table keeps serving last week's figures — is covered by the date, not by the
  binding: what gets written is those figures' own observation date. 151 of the 460 dates landed on
  2026-09-17 rather than today for exactly that reason, which is the mechanism recording staleness
  rather than laundering it.
- **A route measuring the wrong artifact still dates silently.** If a declaration names an artifact
  that is not really the product's, the route agrees with a band derived from the same bad artifact.
  Unchanged by anything here, and not something this mechanism can see.
- **The ledger grows by roughly 2.3k lines per new snapshot.** Nothing prunes entries no axis
  references any more. Worth a follow-up once there is more than one.
- **125 products are now queued and nobody has read the queue yet.** 56 are real tier changes and 69
  are route disagreements, several of which look like the `stars_fallback` / `usage_volume`
  measurement-population problem rather than a re-banding. That is the queue doing its job, but it
  is a backlog this PR creates and does not clear.
- **Next week's run will re-date the same axes to new observation dates**, which is a ~460-file PR
  every Monday. `_already_supported()` suppresses only the no-change case, and the observation date
  moves weekly, so it will not suppress much. Nobody has watched a second run yet.

## Response to round N-1

|---|---|---|---|---|
| `build/adoption_freshness.py:94` — a zero delta becomes AGREED although reconciliation reports `source_unavailable` and unknown freshness; require run-bound evidence before dating an axis | ✅ | | | Answered in a1 `84cc2c6c` and verified here rather than assumed. A zero delta is `BAND_MATCH`, never an agreement; `plan()` dates nothing unless `binding_problems()` passes a bracketed read of `observations.product_adoption_current` that named one materialization at both ends. Tightened this round: the run must also be `SCHEDULED` and `SUCCESS`. Checked live — the run declined nothing this week because the read genuinely bound (`source run 980a87f1… SCHEDULED/SUCCESS`), and the unbound and manual paths are pinned by `test_an_unbound_read_dates_nothing`, `test_an_unbound_read_still_emits_the_queue`, `test_a_refresh_landing_mid_read_dates_nothing`, `test_a_table_a_person_refreshed_dates_nothing_and_still_queues`, `test_a_run_that_has_not_succeeded_dates_nothing`. Reconciliation's `status` is still untouched and #355 is not claimed closed. |
| `build/adoption_freshness.py:284` — the gate checks only the snapshot-wide window; changing accelerate's level, its route to `nonexistent.route` and its date 20 → 24 August returned no problems. Persist and validate per-product agreement evidence | ✅ | | | Answered in a1 `84cc2c6c`; I re-ran your exact repro against the corpus as it now stands, on the real `accelerate` record, and added a fourth edit. Untouched: 0 problems. Level 5 → 4: `the score records level 4, but the measurement that dated it read level 5`. Route → `nonexistent.route`: `is not a route the routing tables compile`. Date 2026-09-17 → 2026-09-19: `last_verified 2026-09-19 is not the observation date 2026-09-17 it derives from`. Ledger trigger → `MANUAL`: 460 problems, every derived axis at once. Corpus restored byte-identical afterwards; script and transcript in the session scratchpad. The same check runs from `check_verification` and from `check_redate`. |
| `.github/workflows/adoption-reconciliation.yml:21` — a configured cron and nothing more; supply a successful schedule-triggered run showing the live reconciliation and the emitted queue | ✅ | | | Not by producing a GitHub Actions `schedule` run — a workflow fires on its cron only from the default branch, so no loop round can. Instead the cron stopped being what the claim rests on. The binding now carries the platform run's `triggerType` and `status`, and only `SCHEDULED`/`SUCCESS` may date an axis, so an unfired cron produces no dates rather than unattributed ones, and a hand refresh produces none either. **The observed SCHEDULED run you asked for is in the diff**: all 460 derived dates name run `980a87f1-3274-408e-928e-70eee35dd8a2`, and `sources/snapshots/observation_snapshots.yaml` records it `trigger_type: SCHEDULED`, `status: SUCCESS`, `started_at: 2026-09-20T03:30:16.737Z` — the `observations` dataset's Sunday 03:30 UTC sweep, which is the run issue #376 was closed on today. The live reconciliation ran for real against the deployed table (a1 had only ever exercised it through a fake control plane) and emitted its queue: 125 products, 56 tier changes and 69 route disagreements, none of them re-dated. What remains unobserved is this workflow's own slot, which now decides only when the queue appears, not whether a date is sound; `deploy-evaluation.md` §5 says what to read on the first fire and that GitHub starts these runs five to six hours after their cron. |

ccerv1:
Reviewed. Codex returned **fail** with two blocking findings; both were confirmed against the code and are fixed in d5c39688.

The 460 score-file edits were verified separately by computation, not by trusting the PR's description:

- **0** changes to `level`, `reach` or `band` across all 460 files. This PR changes what dates a score, never what it is.
- 460 `derived_from` blocks added, **0** on a non-adoption axis.
- **0** cases where `measurement_as_of != last_verified`, and **0** where `measured_level != level`. Every dated axis is one where the route agreed with the recorded band.

Worth correcting a claim made earlier in this run: the dates are **not** a single global `2026-09-17`. They are per-product, derived from the minimum contributing `observed_at` — `2026-09-20` (287), `2026-09-17` (153), `2026-09-19` (19), `2026-09-16` (1). That satisfies "the last run that agreed for that product" more strictly than a single execution date would.

## Blocking 1 — a stale materialization could still earn a date

`binding_problems` checked `bound`, keys present, `run_status == SUCCESS`, `run_trigger_type == SCHEDULED`. All four stay true forever. A materialization from months ago passed every one of them, so a schedule that had stopped was indistinguishable from a healthy one that found no change, and could keep dating scores.

Reads older than `MAX_BINDING_AGE_DAYS` (14, two weekly cycles) are now refused. Two cycles tolerates one missed run while still catching a stopped schedule. Age is read from `materialized_at` — when the table was built, not `read_at`, which is always now and would make every read look fresh. A binding that cannot say when it was built is refused rather than waved through.

## Blocking 2 — the workflow could report success after failing

The reconciliation piped into `tee` under the runner's default `bash -e`, which does not set `pipefail`, so the step reported tee's exit status. A crash after the queue file was written passed as success. `registry.yml` already guards this hazard with the same idiom and the same comment; this now matches it. I checked the other workflows — that was the only unguarded instance.

## One thing the fix introduced and then closed

Adding an age check made every fixture carrying a fixed `materialized_at` expire on a wall-clock date — `binding_problems(BOUND) == []` would have started failing around 2026-10-04 for no code reason. `_utcnow` is now a named seam and an autouse fixture pins it, so the tests fail on a change rather than on a date.

The gate was mutation-checked: raising `MAX_BINDING_AGE_DAYS` to 99999 fails `test_a_materialization_older_than_the_window_dates_nothing`, restoring it passes. 151 passed across the adjacent modules.

## Codex's non-blocking confirmations

- The disagreement guard **is** implemented with negative coverage: the planner emits no change, applying preserves the date, the queue names the disagreement, and the writer rejects a forged disagreement change.
- `reverify` cannot re-date adoption on any route, asserted by a test that attempts the write and checks the file is unchanged.
- `warehouse/assets.yaml` parses, and updates an existing repo-owned governed output rather than adding an OSO-owned dependency, so it is on the correct side of the assets/dependencies line.
- `evidence-and-freshness.md` reads as canonical reference with no issue-number chronology.

Issue #647, opened 2026-09-21
check_parity is red: the warehouse scoring models still carry the pre-#430 taxonomy


ccerv1:
`check_parity` fails today with **149 divergences of 686 published rows** (646 agree, 4 abstain on both sides). Filing ahead of the scheduled Monday run so the sentinel issue that opens has the diagnosis waiting.

## It is publish lag, not a scoring disagreement

Every divergence is one of two shapes, and both have the same cause:

| shape | count | what it is |
|---|---|---|
| `no row in the warehouse at all` | **113** | products in the five categories created by the #430 split. The warehouse has never computed them. |
| `repo=abstains warehouse=(...)` | **36** | products still carrying warehouse rows under `agent_tools_protocols`, a category that **no longer exists in the repo**. |

The timeline settles it. `parity` last ran green on **2026-09-14**. The split landed after that:

| category | created |
|---|---|
| `document_conversion`, `search_retrieval` | 2026-09-16 (#607) |
| `ml_orchestration` | 2026-09-16 |
| `agent_protocols`, `agent_tools_connectors` | 2026-09-17 (#616) |

`sources/categories/agent_tools_protocols.yaml` is gone from the repo. The warehouse still scores 36 products under it, and has no rows for anything in the five categories that replaced it.

## Neither side is wrong

`check_parity`'s own message says neither side is automatically right. Here the repo is right and the warehouse is stale: the split is a deliberate, merged taxonomy change, and the scoring chain has not been re-materialized since.

That matches the known shape — a publish is half a refresh, and the scores chain's run history is manual-only. Nothing about the split triggers a warehouse rebuild.

## What it needs

A re-materialization of the scoring chain against the current taxonomy, which is maintainer-only (`CLAUDE.md`: warehouse writes are UDM revisions and static-model reloads under `docs/operations/`). Until then:

- `check_parity` fails every Monday and opens a sentinel.
- Any consumer reading category scores from the warehouse sees the pre-split taxonomy, including 36 rows under a category that does not exist.

Worth deciding at the same time whether a taxonomy change should *require* a rebuild before merge, or whether a known lag window is acceptable and the gate should say so rather than failing weekly into an issue nobody can close without a maintainer.

ccerv1:
Ruled and built: **the gate says a lag window is expected, rather than a taxonomy change
requiring a rebuild before merge.** PR #650.

Requiring the rebuild first would make every taxonomy PR maintainer-gated on a manual warehouse
[...]

PR #620, opened 2026-09-17
Write the published prose for the reader, not the score auditor

ccerv1:
## What

Refs #619 (left open: see "What this does not do"). The product page read as written by the scoring machine for its own auditor: "rung 4, the competitive frontier", "one below the anchor", "level 5 here is measured, not inferred", a star count that would be stale by the next fetch, and a date the panel already prints restated as a footnote on every product. This PR fixes the contract first, then the corpus, one commit per step so each is reviewable on its own.

**Contract and mechanism** (review these first; everything after follows from them)

1. `docs/reference/product-copy.md` becomes the prose guide for every hand-written string on the map: `description`, `comments`, the score `note` on each axis, `shows` (including the source lines under a capability `comparison`), and editor-facing prose. One rule: write for the reader who has never seen the rubric. Eleven goldens, revised by the maintainer, set the bar and the 600-character note guard; the guide also says no usage figure in a note (a star count is stale the day another project passes it; the source line carries it with its date) and names the scorer's shorthand a normal writer would not produce ("holds it at", "stands in", "a band lower", "at this band", "countable channel", "star-based reading", "on this dimension", "the top of the scale").
2. The `comments` verification line is retired. `product_prose.py` finds a dated `Verified … via` sentence in any spelling and treats it as a defect; `sweep_status` marks a product done on its axes alone.
3. All 763 products lose the line, through `set_document_field` and the new `drop_document_field`, no LLM.
4. Three tools for the prose pass: `prose_worklist` (owns the detectors and the counts, over notes, source lines, comparison lines, footnotes and descriptions), `prose_edit` (edits through `components.py` and refuses what the gates refuse: an emptied or over-long note, a date gained, an allowlisted date lost, a pinned under-coverage phrase dropped or introduced, a footnote originated, the rubric's words, a template opening, a usage figure the old note did not carry, a description that refers to the record or speaks as "we"), `check_prose_diff` (a diff touched only prose). `components.field_span` learned two shapes the corpus had (a paragraph break inside a quoted scalar; a quoted scalar with a column-zero continuation), `_source_span` stops the last source entry at a trailing sibling key, and `set_comparison_source` reaches the lines under `comparison:`. The two allowlists (`DATES_THAT_ARE_PRODUCT_FACTS`, `FIGURES_THAT_ARE_PRODUCT_FACTS`) live in `build/prose_allowlists.py`.
[...]

PR #648, opened 2026-09-21
Record the six days the changelog stopped tracking

ccerv1:
The last entry was #620 on 18 September.

## What was missing

Six days, and the headline is the plainest possible miss: **four published categories landed and the changelog mentions none of them.**

| | |
|---|---|
| `ml_orchestration` | 29 products, from the #429 split |
| `document_conversion` | 20 products |
| `search_retrieval` | 21 products |
| `agent_protocols` | 20 products |
| `agent_tools_connectors` | 21, the renamed remainder |

Also unrecorded: ADR-005, the contradiction sweep and Part 0's falsification rule, the weekly adoption reconciliation and the change in where adoption's dates come from, `registry.product_score_notes`, `components-listed`, three score moves, and five fixes.

## What I left out, and why

The file's own rule is "notable and user-facing", with routine maintenance excluded. So these are **not** here: the five mirror-contract resyncs, the `signal_packages` inventory reconciliation, the shim retirements, the 403 retry, the pytest-invocation doc fix, the source-span regression test, and the two stranded freshness rebases. They are real work and they are not what a changelog reader is looking for.

Judgment calls worth naming:

- **The sentinel watch list** is in, under Fixed. It is CI plumbing, but it changes whether a failure is ever seen, and one of the four unwatched workflows was the weekly adoption reconciliation whose whole output is a queue a person reads.
- **The record-chronology warning** is in for the same reason — it changes what `prose_edit` will tell a contributor.
- **`#647` is linked from the category entry.** The four categories exist in the repo and the warehouse still scores the pre-split taxonomy, so anyone reading "four new categories" needs to know the serving side has not caught up.

## Test plan

- `uv run pytest -k changelog` → 3 passed
- 49 entries in `Unreleased`, 59 linked references, all four new categories present


ccerv1:
Ran a Codex pass over this, asked specifically about language rather than facts. Eight blocking findings; I acted on five, rejected two, and one it withdrew itself.

## Acted on

**The failure-sentinel entry was wrong about what it was describing.** It said adoption reconciliation's "entire output is a queue somebody reads". It is not — the same workflow dates adoption axes, which is the half a silent failure would hide most quietly. Corrected.

**An aphorism closer.** The reference-docs entry ended on "what a document states is what is true now, and the history is git's". The entry had already said what changed. Cut.

**A bullet carrying three changes.** `laminar`, `mcp-inspector` and `opencode` were joined by semicolons under one bullet with three PR links, against this file's own rule of one entry per change. Split into three.

**Two entries under the wrong heading.** The equivalence-recall entry corrects a defect — a denominator counting `pool` candidates, which no tier scores, making the floor unreachable rather than unmet — so it is now under **Fixed**. The withdrawn third-of-a-category rung rule describes a removal and is now under **Removed**.

**The recall entry's metaphor.** "It armed for the first time and failed at a value it had held all along" survived because it explains *why the gate fired now*, but "What the relation exists to do —" was doing rhetorical work and is gone.

## Rejected

**"Rewrite every entry in past tense."** Codex cited this file's header rule, which does say "plain past tense". But no entry in the file follows it — the house voice under `### Added` is a noun fragment (`` `yylo` (`yylo-dev/yylo`) to `orchestration_agents` ``), and under `### Changed` it is present tense. Applying the finding would make these the only entries written differently. **The header rule is stale rather than the entries being wrong** — it also says "one line", which no multi-line entry here obeys. Worth correcting the header at some point; not in a catch-up PR.

**"Omit the figures."** Codex flagged every count — `81 above, 61 below, 3 unmeasured of 145`, `460 adoption axes`, `2,289 rows`, `111 issue references` — as unsupported. It only had the changelog diff, which it says itself in its non-blocking notes. These are point-in-time facts from the PRs that made them, which is what a changelog is for; they do not rot the way a reference document's numbers do. I verified the two that are checkable from the tree today: **24 categories** and **2,289 = 3 × 763 axes**. Both hold.

`uv run pytest tests/test_docs_integrity.py` passes.

PR #650, opened 2026-09-21
Report a taxonomy change the warehouse has not caught up to as lag, not drift

### LINKED ISSUE #647: check_parity is red: the warehouse scoring models still carry the pre-#430 taxonomy

ccerv1:
Closes #647.

## What it changes

`check_parity` splits its divergences into two verdicts.

**Drift** — the two implementations of one rubric disagree about a product both sides know
about. Fails, as before.

**Lag** — the disagreement is a whole category that exists on one side only:

- every product of a category the warehouse has no rows for (a split that has not published), or
- every row under a category with no file in `sources/categories/` (a delete that has not published).

Lag is reported and dated from the commit that created or deleted the category file, and fails
like drift once it is older than `LAG_WINDOW_DAYS` (14 — two Mondays, since the recompute chain
is weekly).

## Why

A merged taxonomy change reaches the warehouse only when a maintainer re-materializes the
scoring chain by hand. The Monday run after the #430 split reported 149 divergences, of which
not one was a scoring disagreement. Failing on that is wrong twice over: it is not drift, and it
opens a weekly sentinel nobody without warehouse credentials can close.

The window is the other half. It exists to give a maintainer time to run the recompute, not to
let the published map serve a taxonomy the repo no longer has.

## Both lag tests are whole-category

A single product missing from a category the warehouse **does** publish is still drift. That is
the shape a roster built on the wrong table produces, and it once hid 36 deferrals. There is a
test pinning it.

## Two things the build turned up

- **The workflow checked out shallow.** `actions/checkout@v4` defaults to depth 1, so `git log`
  could not date a category file and every lag would have read as undatable — which is treated
  as inside the window. The bound would never have bitten in CI. Now `fetch-depth: 0`.
- **The empty-warehouse guard had to be scoped.** An unreadable table must not read as "every
  category is lagging", but under `--category` an empty result is exactly what a category
  awaiting its first recompute looks like. The guard is whole-corpus runs only.

## Testing

- `uv run pytest` — 2380 passed.
- Six new tests over the classifier, patching the dating function rather than leaning on the real
  age of a real category, which would pass until the category turned three weeks old and then
  fail for a reason unrelated to the code. The dating function has its own test against git.
- Run against the live warehouse. Worth knowing: **the warehouse has since caught up** — 763 rows,
  no lag left — so this does not go green by waving #647's divergences through. It goes green
  because they are gone, and the run now isolates one genuine drift that the wall of lag would
  have buried: `colpali [embeddings_retrieval]`, repo `(3, open_weights)` against warehouse
  `(5, open_source)`. Filed separately.

https://claude.ai/code/session_01PTxtguk1MveKsTfu1S54ks


ccerv1:
Codex pass. Two blocking findings; one was a real hole and is fixed, the other is two limits I have written down rather than fixed because they are not fixable from here.

## Fixed: an undatable lag was exempt forever

> The deadline fails open whenever dating returns None. Full checkout history does not fix a warehouse category slug that never existed in Git: its deletion lookup is empty, yet `record` treats it as a deleted category and exempts its divergent rows indefinitely.

Verified — `category_changed_days_ago('some-slug-that-never-existed', 'D')` returns `None`, and `None` counted as inside the window. So the one case that is *certainly* not a taxonomy change had a permanent silent pass.

The leniency was written for a shallow clone, where git genuinely cannot answer. That reason does not extend to a complete history, where undatable means **no commit explains this category**. The gate now asks which it is:

- **shallow** → lenient, and the line says `undatable, shallow clone`
- **complete** → fails with `explained by no commit` and `read it as drift rather than lag`

A failing probe reads as shallow, so a gate that cannot run `git` still reports the warehouse rather than its own environment.

Also widened the dating filter to `{kind}R`. With rename detection on, a file moved *into* the category path is reported as a rename rather than an addition, and one moved out as a rename rather than a deletion — either would previously have read as undatable. Codex caught that too.

## Written down rather than fixed

> Whole-category absence does not establish taxonomy lag. A warehouse query or roster regression can drop every row of a category.

True, and not decidable from the data available. What bounds it is the dating: an established category reads as many days old and fails at once, so the exposure is a category created *inside* the window, where a regression and a genuine lag are indistinguishable from here. Now stated in the module docstring.

> Rows under a recently deleted category are exempted without checking their scores against products' current categories, so actual scoring disagreements can be hidden as lag.

Also true, and it cannot be otherwise. The warehouse row was computed by the ladder of a category the repo has replaced; comparing it against the new category's ladder would be comparing two different questions. A scoring drift predating the split is invisible until the recompute lands. Also now stated.

## Verification

- `uv run pytest tests/test_check_parity.py tests/test_build_docstrings.py` — 21 passed.
- Three new tests, including one pinning the exact hole: a never-existed category slug on a complete history now exits 1.
- Live run unchanged: `755 agree, 7 abstain, 1 diverge, 0 behind the taxonomy` — still isolating #651.

Issue #645, opened 2026-09-21
Triage: ten products whose repository is archived with no recorded end_of_life


ccerv1:
The first run of the contradiction sweep (#641, `build/check_contradictions.py`). Ten products, ten of ten of the archived repos in the corpus. Six record the archival nowhere at all.

Every one needs the same judgment and it is not automatable: **a repository being archived is a fact about a repository; `end_of_life` is a claim about a product.** A project can be archived because it moved, because it was absorbed, or because it is genuinely over, and only the last ends the product.

Two resolutions, both of which clear the finding:

- **The product ended** → record `end_of_life` on `sources/products/<slug>.yaml`.
- **The product outlived the repository** → add an entry to `sources/contradictions_settled.yaml` with the reason. Settlements are bound to the observed value, so they expire if the observation changes.

## The ten

| product | type | last push | what the record already says |
|---|---|---|---|
| `atropos` | software | 2026-07-04 | scored as the RL-environment layer, not a trainer |
| `bigcodebench` | software | 2026-01-03 | "harness repository was archived read-only in January 2026 while the sibling bigcode-evaluation-harness stays active" |
| `ccnet` | software | 2023-04-25 | "archived read-only 2023-10-31 and no longer maintained" |
| `codecontests` | dataset | 2023-10-03 | — |
| `cruxeval` | dataset | 2024-10-11 | — |
| `google-coral-dev-board` | hardware | 2024-08-11 | — |
| `granite-code-instruct` | model | 2025-06-25 | "the model card now opens with a deprecation warning; the line has been folded into the unified Granite 4.x mainline" |
| `granite-geospatial` | model | 2024-09-18 | "the companion GitHub repository is archived and carries getting-started notebooks rather than training code" |
| `hh-rlhf` | dataset | 2025-06-17 | — |
| `llm-guard` | software | 2026-07-08 | — |

## Where I would start, without deciding any of them

Four already say the answer in prose and only need it moved into a field or a settlement:

- **`ccnet`** — "archived read-only 2023-10-31 and no longer maintained" reads like an `end_of_life`.
- **`bigcodebench`** — the record says the *harness* repo was archived while a sibling stays active. That is a settlement, not an end of life, and possibly an artifact-identity question about which repo the product should point at.
- **`granite-geospatial`** — the record says the archived repo carries notebooks rather than training code, and the checkpoint is elsewhere. Reads like a settlement.
[...]

PR #652, opened 2026-09-21
Settle nine archived-repository findings against their distribution

ccerv1:
Works #645. Leaves it open on one product.

## The rule, and what it produced

An archived repository never sets an `end_of_life` on its own. That field goes on the product only
when something product-side says the product ended — a deprecation notice, withdrawn distribution,
a named successor.

Applied to all ten, **none of them has an `end_of_life` to record.** `end_of_life.date` is
specified as "the last day the product works, as the announcement states it", which an archived
repository does not produce, and every one of these ten still installs or still downloads today.

That includes the two #645 expected to resolve the other way:

- **`ccnet`** — GitHub carries the 2023-10-31 archive banner and nothing else. The "no longer
  maintained" in our record is our own gloss, not a quoted statement, and `cc-net` 1.0.0 is still
  on PyPI pointing at that repository.
- **`granite-code-instruct`** — a real "DEPRECATION WARNING" and a superseding mainline family,
  but the weights are still published and IBM names no day they stop being served.

Deprecated is not ended, and unmaintained is not ended.

## The nine

Each bound to the observed `archived`, so it returns if the observation changes, and each carrying
the thing that keeps the product alive:

| product | what outlived the repository |
|---|---|
| `codecontests`, `cruxeval`, `hh-rlhf` | dataset still served and still downloaded on the Hub |
| `bigcodebench` | the sibling `bigcode-evaluation-harness` is still active |
| `granite-geospatial` | checkpoints published away from the archived notebooks |
| `granite-code-instruct` | deprecated and superseded, still published |
| `llm-guard` | the notice ends maintenance, not availability — 0.3.16 still installs |
| `ccnet` | still on PyPI, nothing withdrawn |
| `atropos` | notice scoped to the repository, no successor, environments still usable |

One identity note is recorded rather than left to be rediscovered: the PyPI package named
`atropos` is an **unrelated sequencing-read trimmer**, not this product.

## The one left standing

`google-coral-dev-board` is deliberately **not** settled. It is the one that genuinely ended —
hardware, an official Google EOL notice, `coral.ai` now redirecting to a page that sells none, and
an adoption note that already says the product page names no hardware.

I could not verify the date. Mouser serves a bot-block in place of the notice PDF and no other
fetchable source states a last-ship date. Leaving the finding open beats writing a date nobody
read. It wants either the PCN opened by hand, or a ruling that a discontinued hardware product
takes its date from the notice's publication.

## Testing

- `uv run python -m build.check_contradictions` — 10 contradicted before, **1 after**, and the one
  is `google-coral-dev-board`.
- `uv run python -m build.validate` — 0 errors.
- `uv run pytest` — 2373 passed, 1 skipped.

https://claude.ai/code/session_01PTxtguk1MveKsTfu1S54ks


ccerv1:
Codex pass: **verdict pass**, no blocking findings. It checked each of the nine settlements against the ruling and against its own stated reason, and confirmed all nine distinguish repository archival or ended maintenance from product termination, with none quoting an announced end date that a settlement would be avoiding.

One non-blocking accuracy point, which I took:

> `ccnet` note — "Still published ... so ... there is no day on which it stops working" overstates what package availability establishes. "No product-side end date is announced" would accurately express the ruling.

Fair. A package index can show nothing has been withdrawn; it cannot show the future. Corrected to "nothing has been withdrawn and no product-side end has been announced".

Its other note is worth recording rather than acting on: the diff shows the `archived` bindings but not the code that expires them, so expiry is asserted here rather than demonstrated. That code is `settled_keys` in `build/check_contradictions.py`, which keys on `(leg, product_slug, artifact, settles)` — and `tests/test_check_contradictions.py::test_a_settlement_does_not_cover_a_different_observation` pins it.

`uv run pytest tests/test_check_contradictions.py` and `uv run python -m build.validate` both pass.

